PR #32, opened 2016-01-29
Change the <operation.HttpMethod> to <operation.HttpMethodUpper>

duartepatriani:
I'm using your project to generate my AngularJS api client, and I found a problem, the METHOD of the http call was empty, I notice that AngularJs.txt T4 was using "operation.HttpMethod" instead "operation.HttpMethodUpper" which is been using Jquery T4 template who works fine.


rsuter:
Thank you for you fix... I haven't tested the Angular 1.x template yet... Is everything working for you? Please submit more PRs if more fixes are needed. I do not know the Angular http service that well...

duartepatriani:
You're welcome, your project is very useful, I'll start to use the generated client today, if I find something I'll fix and let you know.

PR #34, opened 2016-02-03
Add client seting

irbissaibri:
I'm using NSwag for generate client, and can't authorize in app. I added client settings for transfer api credentials, and addeddateTimeConverter for dyanamic changes settings datetime formats.


rsuter:
Thanks for your PR. However the generated code is extremely tied to your needs and produces lots of code. Additionally I think the client should not be configurable because it is couple to an active REST service with a fix configuration... Therefore the client should be directly generated the way it must be used, and should not be configurable. All your special settings can be implemented with partial classes and base classes as shown here: 

https://github.com/NSwag/NSwag/wiki/SwaggerToCSharpClientGenerator

PR #38, opened 2016-02-10
fixed conditional usging text instead number

duartepatriani:
The property "status" in the response object is a number, I fixed the conditional

[Angular http documentation](https://docs.angularjs.org/api/ng/service/$http)


rsuter:
Thank you for fixing this.

rsuter:
Now available as NuGet: https://www.nuget.org/packages/NSwag.Core/1.35.5884.20833
Download and install latest MSI installer to use in NSwagStudio

PR #43, opened 2016-03-03
Fixed crash when no Controllers are specified as arguments.

sbonello:
webapi2swagger /Assembly:C:\Projects\EDM\src\EDMService\bin\EDMService.dll /Controllers: /Controller:


rsuter:
Thanks for contributing, v1.37 should be live soon...

PR #46, opened 2016-03-10
can now resolve dll dependencies outside of webapi assembly folder

loictr:
Added a ReferencePaths parameter to the WebApiAssemblyToSwagger command to provide additional folders where to look for referenced assemblies


rsuter:
This looks good, there is one broken unit test and ill fix that... Can you explain how you can load a .NET Core Web API assembly?

loictr:
Hi,

Actually, I do not load it. It loads the dnx451 version that comes along
for default visual studio 2015 xproj projects. So this can be a workaround.
The issue with dnx451 was the location of the referenced dlls. It is not
the best solution but it does the job for me. And it may do it for other
users.

rsuter:
What relative paths are you using for generating the Swagger spec for a dnx project?

loictr:
Hi,

I used it in multiple ways. My final usage is as a TFS 2015 task. My build
includes several custom powershell tasks to launch the commands :
- dnu restore
- dnu publish --no-source
- nswag webapi2swagger
- nswag swagger2csclient
- ...

So, the webapi2client arguments are based on the folders tree created by
the publish command. I pass as "assembly" the latest built assembly for my
webapi, and in referencePaths the package folder created by the publish
command.
The working directory is the root of the pulled source code.
My dnu publish "out" value is ./<MyPublishFolder>.
My "relativePaths" value is ./<MyPublishFolder>/approot/packages.


Best regards,

Loïc

Issue #70, opened 2016-04-26
Visual Studio 2015 Integration / Post Build ?


CastleSoft:
What is the recommended way to integrate the Command Line for NSwag into a Webapi project ?

Currently I build the WebAPI and either switch to the GUI or command line.

But it would be nice to compile and automatically have the NSwag TS files generated.

Is this something that could be done using Nuget ? Or is it best via Post Build commands ?

LIke Arebis does ? You can then reference a .../Client/MyWebapi.js   containing the wrapper ?

http://www.codeproject.com/Articles/872033/ASP-NET-Web-API-Javascript-Clients

https://www.nuget.org/packages/Arebis.AspNet.WebApi.JsClients.cs/

rsuter:
Currently there is no NSwag build target / NuGet package with build targets available. I think the best option is to add a post compile target/task and execute the command line tool...

rsuter:
Use NSwag.MSBuild

rsuter:
See https://github.com/NSwag/NSwag/wiki/MSBuild

PR #129, opened 2016-05-20
Issue #70 - Added NuSpec file for NSwag Console.

jkonecki:
The new package can be easily referenced in a project and a simple PostBuild task can be added as required.


rsuter:
Thanks for PR, I have to check the generated nuget.. I think it has dependencies to all other packages even if everything is included in the package... If it is like that we have to fix this first..

jkonecki:
Sure. I haven't defined any dependencies so it should be a stand-alone package. Let me know if otherwise.

rsuter:
... the nuspec has not been not picked up. You have to rename it to `NSwag.nuspec` (same as project file) so that the build agent finds the nuspec...

rsuter:
.. and please add an empty `<dependencies></dependencies>` tag - maybe this forces the packager to omit all dependencies..

rsuter:
Usually the NuGet packager automatically determines all referenced projects in the csproj file and adds them to the dependency list...

rsuter:
How do you reference the nuget dlls in the postbuild and where do you install this nuget? I think you have to hardcode the full path (ie. full nuget name + version) into the csproj/msbuild task

jkonecki:
Yes, that's correct. Similar is needed for example for NUnit.ConsoleRunner.
What should be the Id for the console package? After I change the name to NSwag.nuspec the package name will be called NSwag.nuget. I think NSwag.Console.nuget is a better name for the package.

rsuter:
the id will be copied from the AssemblyInfo.cs which is NSwag.Console...

jkonecki:
Ah, yes. You're using $id$. Making the change now

rsuter:
I think it would be much better, if we create a `NSwag.MSBuild` package with a .targets file which is automatically added to the project and which provides nice build tasks and always uses the installed package...

rsuter:
Now its the wrong id :-): https://ci.appveyor.com/project/rsuter/nswag-25x6o/build/195/artifacts
we have to hard code it in the nuspec

rsuter:
And the dependencies are added: 

image

jkonecki:
How are you building the packages? I couldn't find anything in code.
Are you passing -IncludedReferencedProjects?
I think there's an option to prefer nuspec file over csproj

rsuter:
Its built by AppVeyor with IncludeReferencedProjects (required for other nuspecs)... I have to check how to do this - and if possible do the .targets solution because it is much simpler to use..

jkonecki:
Targets file is a nice idea - my only issue is that I would like to avoid having to install all NSwag packages in my project with the references added as I don't want to deploy NSwag.*.dlls with the application.

rsuter:
It will be a development dependency and no dlls are referenced, just the targets file, but this file will no be published..

jkonecki:
At the moment I'm integrating NSwag into my project and having a package with exe inside is sufficient for me. I'm happy to help you with the targets file in the near future.

rsuter:
Ok, it will be one of my next tasks..

rsuter:
See here: https://github.com/NSwag/NSwag/wiki/MSBuild

rsuter:
Is this the way you expect this to be?

rsuter:
Use NSwag.MSBuild package instead

jkonecki:
The `NSwag.MSBuild` package works perfectly fine. Thank you very much!

I however have a problem with the following command:

`nswag,exe webapi2swagger /assembly:MyWebApi.dll /output:swagger.json`
Please note that I don't provide `/controller` argument. 

Instead of processing all the controllers the app shows the message: _The Web API controller full class name or empty to load all controllers from the assembly._ followed by _Controller:_ and prompt. 
According to the wiki omitting the `/controller` argument should be enough to process all controllers. What am I doing wrong?

jkonecki:
I've noticed that if I press <Enter> in the console there will be an exception in 

```
        public AppDomainIsolation(string assemblyDirectory, string assemblyConfiguration)
        {
            if (string.IsNullOrEmpty(assemblyDirectory))
                throw new ArgumentNullException(nameof(assemblyDirectory));
```

as the `assembyDirectory` is an empty string. It is returned from `Path.GetDirectoryName(Settings.AssemblyPath)`

Did you mean `Path.GetFullPath(Settings.AssemblyPath)`?

rsuter:
You are probably right, if assembly path is relative it may be a problem... I also have to check the cmd line param defaults..

rsuter:
Ok both problems should be fixed in: https://github.com/NSwag/NSwag/commit/73159d287f4704e88296eac33f3b5a4e1d543187

Can you download the cmd line tools from the build artifacts and replace your files to test it:
https://ci.appveyor.com/project/rsuter/nswag-25x6o/build/214

If everything is fine, I'll release a new nuget...

jkonecki:
Works fine, thank you!
I'm getting `The operation has more than one body parameter` exception but this is related to my schema I believe. You can publish the packages.

rsuter:
v2.33 should be live in 15 minutes... please inform me if NSwag.MSBuild upgrade works flawlessly (ie new targets file is referenced in project file, ...)

jkonecki:
Upgrade was fine, project file was updated correctly. Thanks again!

rsuter:
Ok, fine.. Please create new issues if you encounter more problems... I mainly use the UI..

jkonecki:
I will. I think you can close the related issue #70.

Issue #78, opened 2016-05-02
Cmd Line webapi2swagger: Create missing output directory


CastleSoft:
Given the following Post-Build event command line:

```
nswag webapi2swagger /assembly:$(TargetPath) /controller:WebApplication1.Controllers.ValuesController /output:$(ProjectDir)\TypeScript\$(SolutionName).json
nswag swagger2tsclient /input:$(ProjectDir)\TypeScript\$(SolutionName).json /output:$(ProjectDir)\TypeScript\$(SolutionName).ts
```

If the TypeScript folder doesn't exist, nswag fails with an error.

It would be nice if the code created the directory if it didn't exist.

rsuter:
I'm sure there is an msbuild task to create a missing directory... try this until this feature is implemented.

jkonecki:
I believe this issue is fixed and can be closed.

rsuter:
Fixed in v2.33

rsuter:
See https://github.com/NSwag/NSwag/pull/130

PR #130, opened 2016-05-20
Issue #78 - Creating output directory if it doesn't exist.

jkonecki:


rsuter:
Does this `directory.Create()` work recursively (ie create all required directories)?

jkonecki:
Yes. It will create all parent directories as well if they don't exist.

PR #190, opened 2016-07-04
Create empty response when using ResponseType attribute with void

fhoring:
- used same behavior as for default result types
 - on code generation permits to create resultType as Task instead of Task<void>

### DIFF
--- a/src/NSwag.CodeGeneration/SwaggerGenerators/WebApi/WebApiToSwaggerGenerator.cs
+++ b/src/NSwag.CodeGeneration/SwaggerGenerators/WebApi/WebApiToSwaggerGenerator.cs
@@ -646,12 +646,14 @@ namespace NSwag.CodeGeneration.SwaggerGenerators.WebApi
                             description = dynResultTypeAttribute.Description;
                     }
 
-                    operation.Responses[httpStatusCode] = new SwaggerResponse
-                    {
-                        Description = description ?? string.Empty,
-                        IsNullableRaw = mayBeNull,
-                        Schema = CreateAndAddSchema(service, returnType, mayBeNull, null, schemaResolver)
-                    };
+                    operation.Responses[httpStatusCode] = IsVoidResponse(returnType) ?
+                        new SwaggerResponse() :
+                        new SwaggerResponse
+                        {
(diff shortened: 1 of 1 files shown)


src/NSwag.CodeGeneration/SwaggerGenerators/WebApi/WebApiToSwaggerGenerator.cs
```diff
@@ -646,12 +646,14 @@ private void LoadReturnType(SwaggerService service, SwaggerOperation operation,
...
                     }
 
-                    operation.Responses[httpStatusCode] = new SwaggerResponse
-                    {
-                        Description = description ?? string.Empty,
-                        IsNullableRaw = mayBeNull,
-                        Schema = CreateAndAddSchema(service, returnType, mayBeNull, null, schemaResolver)
-                    };
+                    operation.Responses[httpStatusCode] = IsVoidResponse(returnType) ?
+                        new SwaggerResponse() :
```
rsuter: Here you should also add `Description = description ?? string.Empty, `

rsuter:
What problem does this solve?

fhoring:
I have the impression that c# client Generation using the Attribute Response type isn't working. (Using [ResponseType(typeof(void))] in endpoints) isn't working. The property ResultType is generated as Task<void> instead of Task. The fix could also be done in client generation code but it looked more appropriate here.

rsuter:
How does the generated swagger look like for a void action?

rsuter:
I think its more a generator problem because it correctly detects task<void>... should be simple to fix in the CSharpTypeResolver..

rsuter:
No, in the generator itself in the method GetResultType: 

https://github.com/NSwag/NSwag/blob/master/src/NSwag.CodeGeneration/CodeGenerators/CSharp/SwaggerToCSharpGeneratorBase.cs

rsuter:
Ok, your solution is correct... you also need to set the Description property because it cannot be null: 
https://github.com/NSwag/NSwag/commit/961c713f393e1607d04c8cd3dca3862d8b22e538

rsuter:
Thanks for the PR

fhoring:
Thanks

PR #249, opened 2016-08-22
Make model public

ShaKann:
Hi. From 5.x version I am able to use ITemplateFactory to provide my own templates. That's really good approach (I made something similar in my previous fork) but there is just one problem - models are still internal so I can't easily access theirs properties. Also I don't have intelisense on my T4 template. There will be much easier if I can use them directly in my code. I have create PR with appropriate changes. Please take a look and tell me: if this is good approach and if not how I can easily reuse your model?

### DIFF
--- a/src/NSwag.CodeGeneration/CodeGenerators/CSharp/Models/ClientTemplateModel.cs
+++ b/src/NSwag.CodeGeneration/CodeGenerators/CSharp/Models/ClientTemplateModel.cs
@@ -12,11 +12,21 @@ using NSwag.CodeGeneration.CodeGenerators.Models;
 
 namespace NSwag.CodeGeneration.CodeGenerators.CSharp.Models
 {
-    internal class ClientTemplateModel
+    /// <summary>
+    /// 
+    /// </summary>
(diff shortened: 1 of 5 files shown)


ShaKann:
Just one thing - I used stag comments as I don't know exactly what description will be correct.

rsuter:
Yes, making the models public would be the next stop. Also some models are still dynamic (i.e. of type object) and must be converted to a strongly-typed class...

ShaKann:
So, first you want to strongly-type model and then make it public? Ok... So when do you think this will be released and do you think I can help you with that?

rsuter:
I will merge your PR, but we need to improve the XML comments first: 

- All tags on a single line
- Remove the <value> tags
- Add class <summary> tag contents (ie. "The CSharp client template model")

rsuter:
Then we'll refactor the model classes and add typed model classes for the missing models (see Visual Studio Tasks)...

rsuter:
For more discussions: https://github.com/NSwag/NSwag/issues/51

PR #257, opened 2016-08-25
FIx getAll for AngularJS

ShaKann:
Fix of getAll(). "?" was added even without parameters. Now it is added only when needed.


rsuter:
Thank you. Maybe you can add this change to the other templates as well (JQuery, AngularJS, C#)?

PR #295, opened 2016-09-16
Master

michaelaird:
Prefering let/const to var.


rsuter:
Thanks for the PR. Please check the current version in the "master" branch. Can we close https://github.com/NSwag/NSwag/issues/240?

PR #297, opened 2016-09-20
Addition of UseHttpRequestMessageCreationMethod option

aarro:
I needed the ability to modify headers, namely the Authorization header, on each request before it was sent. In order to do that and still stay in the spirit of supporting the UseHttpClientCreationMethod option, where it felt really hacky to attempt to use the HttpClient's DefaultRequestHeaders, I created a new option for UseHttpRequestMessageCreationMethod. This allows the base class to handle the creation of an HttpRequestMessage object that's used by all calls.

In order to support using the HttpRequestMessage for all calls, I also changed all the verb-based methods being used off of HttpClient to be SendAsync.


rsuter:
You should make this method async so that it can be used in more scenarios: 

    await CreateHttpRequestMessageAsync(cancellationToken).ConfigureAwait(false);

aarro:
Good idea...done.

rsuter:
Ok merged, please check if everything is fine with the latest code in the "master" branch...

rsuter:
Added setting to UI: https://github.com/NSwag/NSwag/commit/305db1d4afcd4179fec2b0ea70127af631cfe4b5

rsuter:
Thanks for the PR

rsuter:
With this change, we can also remove 
    internal static class HttpExtensions
correct?

aarro:
Yes I think HttpExtensions can go away...sorry I didn't notice that before.

rsuter:
Ok, removed: https://github.com/NSwag/NSwag/commit/0ec3f6b01d37f2b1777c414383a8166c783cf5b1

Thanks for your contribution.

PR #314, opened 2016-09-28
Changed RequestOptionsArgs to RequestInit for fetch (TS)

grovesNL:
RequestOptionsArgs appears to be Angular specific, the fetch specification uses the type RequestInit


rsuter:
Thanks for the PR and improving NSwag

PR #346, opened 2016-10-09
Removed unnecessary Owin dependency

cpx:
NSwag.AssemblyLoader has a dependency on Microsoft.Owin and Owin, which causes the NuGet package to depend on these as well.

This PR simply removes those two. Everything compiles so it seems they are not needed.


rsuter:
Good find. They must have been added by mistake...

cpx:
Btw, could you publish another NuGet package? I'm *this close* to getting this up and running in our CI, but it would be nice to get rid of this package dependency. :)

rsuter:
I will release it tomorrow... ok?

cpx:
Cool. Sorry to keep bugging you, I'm just really excited to start using this for production code. :)

rsuter:
Ill fix the FileParameter issue first and will release then..

PR #347, opened 2016-10-09
Don't generate FileParameter class in C# client if not necessary.

cpx:
The TypeScript generator and template has a check for whether it's necessary to generate a type for file parameters, but the C# generator is missing it. This PR adds the same functionality to the C# generator.


rsuter:
The C# generator always needs the FileParameter type (only type with known name) in the rendered JsonExceptionConverter class to find the model assembly. Its a little bit hacky. We could also render a static info class and use this instead...

cpx:
Are you referring to this line? https://github.com/NSwag/NSwag/blob/deaa9779856fd5fd1d51555ebb860c5f562d1668/src/NSwag.CodeGeneration/CodeGenerators/CSharp/Templates/JsonExceptionConverterTemplate.tt#L10

Couldn't it resolve the model assembly using itself? Something like:
```csharp
_modelAssembly = typeof(JsonExceptionConverter).Assembly;
```

Also, how does the reader code use this? Looking at this line:
```csharp
exceptionType = pair.Value.GetType(pair.Key + "." + discriminator);
```
It looks like it would evaluate ```pair.Key + "." + discriminator``` to "FileParameter.{discriminator}"?

rsuter:
It must be "Namespace": Fixed in https://github.com/NSwag/NSwag/commit/98b0dbee04f5c0f6b8e7f4e0d6c2e9ee9c65861b

And we cannot use `typeof(JsonExceptionConverter).Assembly` because the models (i.e. the exception models, models library) may be in a different assembly than the `JsonExceptionConverter` class (client library)...

cpx:
But how does that work together with this code? https://github.com/NSwag/NSwag/blob/deaa9779856fd5fd1d51555ebb860c5f562d1668/src/NSwag.CodeGeneration/CodeGenerators/CSharp/Templates/JsonExceptionConverterTemplate.tt#L81

Wouldn't that load the exception type anyway from the same assembly and namespace as FileParameter and JsonExceptionConverter?

rsuter:
With the CSharp generator, you can output the models and clients into different files (which may reside in different assemblies). The FileParameter goes into the models file output, the JsonExceptionConverter goes to the clients file output... The JsonExceptionConverter is only rendered when an custom exception model is defined, it would be better to use this exception model class instead of FileParameter and your problem would also be solved...

rsuter:
Also needed: https://github.com/NSwag/NSwag/commit/56ba61df6437f5815f3edadfe9a73ed6f479961b

Liero:
This code generates FileParameter twice, althougt it isn't use anywhere:

     static SwaggerToCSharpClientGeneratorSettings CreateBasicSettings()
     {
         var settings = new SwaggerToCSharpClientGeneratorSettings
         {
             //ClientBaseClass = "UispBackendBase",
             UseBaseUrl = true,
             ClassName = "UispBackend",
             GenerateClientClasses = false,
             GenerateExceptionClasses = false,
             GenerateClientInterfaces = false,
             GenerateDtoTypes = false,
             ExceptionClass = "TechnicalFaultException",
         };
         settings.CSharpGeneratorSettings.Namespace = "Accenture.UI.Backend";
         return settings;
     }

          var swaggerDocument = await SwaggerDocument.FromFileAsync(swaggerFile);

          var contractsSettings = CreateBasicSettings();
          contractsSettings.CSharpGeneratorSettings.Namespace += ".Contracts";
          contractsSettings.GenerateDtoTypes = true;
          contractsSettings.GenerateExceptionClasses = true;
          Generate(swaggerDocument, contractsSettings, outputFolder + @"\Contracts\Entities.g.cs");

          var interfacesSettings = CreateBasicSettings();
          interfacesSettings.CSharpGeneratorSettings.Namespace += ".Contracts";
          interfacesSettings.GenerateClientInterfaces = true;
          Generate(swaggerDocument, interfacesSettings, outputFolder + @"\Contracts\Interfaces.g.cs");

          var communicationSettings = CreateBasicSettings();
          communicationSettings.CSharpGeneratorSettings.Namespace += ".Communication";
          communicationSettings.GenerateClientClasses = true;
          communicationSettings.AdditionalNamespaceUsages = new[] { contractsSettings.CSharpGeneratorSettings.Namespace };
          Generate(swaggerDocument, communicationSettings, outputFolder + @"\Communication\ClientProxy.g.cs");

PR #362, opened 2016-10-13
bug in AngularJS client generation

alpercelik:
in AngularJS client code generation "?" is not added to url when there are any query parameters and no path parameters.

for web api controller below
```
    [RoutePrefix("api/v1")]
    [Authorize]
    public class ValuesController : ApiController
    {
        [HttpGet]
        [Route("foo")]
        [ResponseType("200", typeof(IEnumerable<int>))]
        public IHttpActionResult Get(string date, int tagId, int dataTypeId)
        {
            return Ok(new[] { 1, 2, 3 });
        }
    }
```

generated AngularJS client code is as below, missing "?" at the end

```
    get(date: string, tagId: number, dataTypeId: number): ng.IPromise<number[]> {
        let url_ = this.baseUrl + "/api/v1/foo"; 

        if (date === undefined)
            throw new Error("The parameter 'date' must be defined.");
        else
            url_ += "date=" + encodeURIComponent("" + date) + "&"; 
        if (tagId === undefined || tagId === null)
            throw new Error("The parameter 'tagId' must be defined and cannot be null.");
        else
            url_ += "tagId=" + encodeURIComponent("" + tagId) + "&"; 
        if (dataTypeId === undefined || dataTypeId === null)
            throw new Error("The parameter 'dataTypeId' must be defined and cannot be null.");
        else
            url_ += "dataTypeId=" + encodeURIComponent("" + dataTypeId) + "&"; 

        const content_ = "";

        return this.http({
            url: url_,
            method: "GET",
            data: content_,
            transformResponse: [], 
            headers: {
                "Content-Type": "application/json; charset=UTF-8"
            }
        }).then((response) => {
            return this.processGet(response);
        });
    }
```

### DIFF
--- a/src/NSwag.CodeGeneration/CodeGenerators/TypeScript/Templates/AngularJSClientTemplate.cs
+++ b/src/NSwag.CodeGeneration/CodeGenerators/TypeScript/Templates/AngularJSClientTemplate.cs
@@ -461,7 +461,7 @@ if(!parameter.IsLast){
             #line hidden
             
             #line 52 "C:\Data\Projects\NSwag\src\NSwag.CodeGeneration\CodeGenerators\TypeScript\Templates\AngularJSClientTemplate.tt"
-if(operation.PathParameters.Any()){
+if(operation.QueryParameters.Any()){
             
             #line default
             #line hidden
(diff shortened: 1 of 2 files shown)


rsuter:
Why did you close the pr?

rsuter:
Fixed: https://github.com/NSwag/NSwag/commit/e93da94b2486110dac74589d70a8b1b1fa862186

alpercelik:
sorry my bad, committed with wrong email. I opened another pull request.

rsuter:
I think I'll release next week, but you can get the latest MSI here: https://ci.appveyor.com/project/rsuter/nswag-25x6o/build/artifacts

rsuter:
Is this ok for you?

alpercelik:
Fine for me. I already informed colleagues that are using nswagstudio, about new installer from the link you provided. Thanks. So for now fine for us, though there may be some others that are affected with this.

rsuter:
v6.12

PR #384, opened 2016-10-25
fixed AngularJS client; promise not correctly resolved for unsuccessful response code

alpercelik:
In AngularJS client, no error callback is present. so processXXX method is not called.

When error callback is missing in client then, response object itself is passed to error callback of the caller not the parsed value of response.data.

### DIFF
--- a/src/NSwag.CodeGeneration/CodeGenerators/TypeScript/Templates/AngularJSClientTemplate.cs
+++ b/src/NSwag.CodeGeneration/CodeGenerators/TypeScript/Templates/AngularJSClientTemplate.cs
@@ -923,9 +923,45 @@ if(Model.IsExtended){
             
             #line default
             #line hidden
-            this.Write("        });\r\n    }\r\n\r\n    private process");
+            this.Write("        },(response) => {\r\n");
+            
+            #line 122 "C:\Data\NSwag\src\NSwag.CodeGeneration\CodeGenerators\TypeScript\Templates\AngularJSClientTemplate.tt"
+if(Model.IsExtended){
+            
+            #line default
+            #line hidden
+            this.Write("            return this.transformResult(url_, response, (response) => this.proces" +
+                    "s");
+            
+            #line 123 "C:\Data\NSwag\src\NSwag.CodeGeneration\CodeGenerators\TypeScript\Templates\AngularJSClientTemplate.tt"
+            this.Write(this.ToStringHelper.ToStringWithCulture(operation.OperationNameUpper));
+            
+            #line default
+            #line hidden
+            this.Write("(response));\r\n");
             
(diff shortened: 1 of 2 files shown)


rsuter:
Is the catched object always a response object? Maybe we have to check this too or rethrow else...

alpercelik:
Yes, response object that is resolved or rejected are same response object and passed to both success and error callback, as per angularjs code.

In angular if the response code is success (between, 200 and 299) 

```
// in angular

function isSuccess(status) {
  return 200 <= status && status < 300;
}
```
then promise is resolved otherwise rejected with same response object

```
/**
 * Resolves the raw $http promise.
 */
function resolvePromise(response, status, headers, statusText) {
  //status: HTTP response status code, 0, -1 (aborted by timeout / promise)
  status = status >= -1 ? status : 0;

  (isSuccess(status) ? deferred.resolve : deferred.reject)({
    data: response,
    status: status,
    headers: headersGetter(headers),
    config: config,
    statusText: statusText
  });
}

```

So the the processXXX method is not called and does not process any response with unsuccessful response (outside range (200-299) but has code to parse response.data and throw error. 

As I see in jquerycallbacks client it calls same method in case of fail
```
jQuery.ajax({
    url: url_,
    beforeSend: this.beforeSend,
    type: "get",
    data: content_,
    dataType: "text",
    headers: {
        "Content-Type": "application/json; charset=UTF-8"
    }
}).done((data, textStatus, xhr) => {
    this.processGetWithCallbacks(url_, xhr, onSuccess, onFail);
}).fail((xhr) => {
    this.processGetWithCallbacks(url_, xhr, onSuccess, onFail);
});
```

So, I think same should be for AngularJS client as well (and also others?)

Would it be good idea to to have to different methods for success and error callback handlers? There may be multiple success and also multiple fail response results. Methods responsible for processing success and error status code cases could be separate.

rsuter:
Everything is correct what you said. My only question is: What object is in request if no response could be retrieved? For example the server is not available...

alpercelik:
Sorry, misunderstood your question.

That is then different. Then promise is rejected  with status code -1.

```
  var requestError = function() {
    // The response is always empty
    // See https://xhr.spec.whatwg.org/#request-error-steps and https://fetch.spec.whatwg.org/#concept-network-error
    completeRequest(callback, -1, null, null, '');
  };

```

alpercelik:
in this case following response is passed to error callback function
```
{data: null, status: -1, config: Object, statusText: ""}
```

rsuter:
I added a check: https://github.com/NSwag/NSwag/commit/4c1e4a9460e9f73083e736dd97bdb8530c7e8324

rsuter:
Is this ok for you? 

Please try the latest CI version and report if everything is fine: 
https://ci.appveyor.com/project/rsuter/nswag-25x6o/build/artifacts

... then I'll create a new release.

alpercelik:
This is good, thanks.

PR #386, opened 2016-10-29
Fix argument name to match property name

billknye:
Looks like this was a missed change from a copy/paste.

Added in: https://github.com/NSwag/NSwag/commit/c7ecee95422b17f9a50681df93339fbe01c5665e


billknye:
This causes a conflict with the real UseHttpClientCreationMethod and makes rather silly code generated such as:

`result_ = JsonConvert.DeserializeObject<ObservableCollection<Person>>(Encoding.UTF8.GetString(responseData_, 0, responseData_.Length), new JsonConverter[] { new true() });`

rsuter:
Thank you for fixing this.

rsuter:
I will create a release today with this fix...

rsuter:
Fixed in v6.17

Issue #62, opened 2016-04-20
TypeScript: Improve date time handling in clients


rsuter:
1. Support `"\/Date(1198908717056)\/"` by replacing all matches with `new Date(...)`: http://weblogs.asp.net/bleroy/dates-and-json
2. Support ISO date time format by converting string values to  Date  (only possible with TS classes): 
Must be fixed in NJsonSchema: https://github.com/NJsonSchema/NJsonSchema/issues/55
3. Add optional reviver function to `JSON.parse(text[, reviver])`: 
https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse

```
var text = response.text();
var data = JSON.parse(text, this.reviver);
```
Date/time revivier function: http://stackoverflow.com/questions/1792865/how-to-use-json-parse-reviver-parameter-to-parse-date-string

The default reviver function should implement 1. from this list...

rsuter:
1. Support added: https://github.com/NSwag/NSwag/commit/c71f2ae9615c987be45c732f73c948a714028a73
2. Still open
3. Reviver added to Angular 2 tpl: https://github.com/NSwag/NSwag/commit/c71f2ae9615c987be45c732f73c948a714028a73

berhir:
I have tested the new version and I get some TypeScript errors.


`readonly` properties are planned for TypeScript 2.0 but we still use 1.8.x. Is it really necessary to break backward compatibility?

And the following error:
 ```
var text = response.text().replace(/\/Date((-?\d*))\//, (a: string, b: string) => { return new Date(+b); });

Argument of type '(a: string, b: string) => Date' is not assignable to parameter of type '(substring: string, ...args: any[]) => string'.
  Type 'Date' is not assignable to type 'string'.	
```

rsuter:
Oh no, I'll fix this ASAP. 
Readonly is already implemented for some months... maybe I should add an option to disable this... Where does your readonly properties come from? Do you have readonly attributes on the C# classes?

rsuter:
Ok, fixed, I removed the replace part (because it is not really needed anymore)... only problem are the readonly props: Ill add an option to disable the generation of the readonly keyword

berhir:
We have readonly attributes on our C# classes and Swashbuckle reflects this in the generated swagger.json.
I have tried it again with NSwagStudio version 2.5.5940.40228 (06.04.2016 22:20:56) and there are no readonly attributes in the generated TypeScript interfaces.

[...]

PR #399, opened 2016-11-02
Add date reviver to fetch (#62)

grovesNL:
Adds date reviver to fetch - uses the same methodology that was added to Angular

### DIFF
--- a/src/NSwag.CodeGeneration/CodeGenerators/TypeScript/Templates/FetchClientTemplate.cs
+++ b/src/NSwag.CodeGeneration/CodeGenerators/TypeScript/Templates/FetchClientTemplate.cs
@@ -273,14 +273,16 @@ if(Model.GenerateClientInterfaces){
             this.Write(@"{
     private baseUrl: string; 
     private http: { fetch(url: RequestInfo, init?: RequestInit): Promise<Response> };
+	private jsonParseReviver: (key: string, value: any) => any = undefined; 
 
-    constructor(baseUrl?: string, http?: { fetch(url: RequestInfo, init?: RequestInit): Promise<Response> }) {
+    constructor(baseUrl?: string, http?: { fetch(url: RequestInfo, init?: RequestInit): Promise<Response> }, jsonParseReviver?: (key: string, value: any) => any) {
         this.baseUrl = baseUrl ? baseUrl : """";
         this.http = http ? http : window;
+		this.jsonParseReviver = jsonParseReviver;
     }
 ");
             
-            #line 31 "C:\Data\NSwag\src\NSwag.CodeGeneration\CodeGenerators\TypeScript\Templates\FetchClientTemplate.tt"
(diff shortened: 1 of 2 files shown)


rsuter:
The JSON reviver has been removed from the Angular 2 template because it is bad design when the user of the client class has to implement the correct JSON reviver (or may alter the behavior of the client class which is tailored to a specific service). We have to generate code that already works. 

My suggestion: Implement the reviver as is but remove it from the constructor and make the field protected. This way you can write an extension class and set the reviver in the extension class constructor. This way we ensure that the class behavior is encapsulated and everything is working right after the generation. 

What do you think?

Extension classes: https://github.com/NSwag/NSwag/wiki/SwaggerToTypeScriptClientGenerator#extended-classes-and-extension-code

rsuter:
See https://github.com/NSwag/NSwag/pull/400

grovesNL:
@rsuter: Agreed, that's a much better idea. I actually considered this when I added it but wanted to follow the Angular 2 template. Let's use the same pattern as #400 for fetch

rsuter:
I've updated the wiki: https://github.com/NSwag/NSwag/wiki/SwaggerToTypeScriptClientGenerator#add-a-custom-jsonparse-reviver-function

rsuter:
Just remove the ctor param and assignment and I'll merge both PRs...

rsuter:
... and change the jsonParseReviver field to protected.

grovesNL:
Thanks! I guess the only downside is that `jsonParseReviver` is a property instead of a method (like `transformOptions`) so it has to be set inside the constructor rather than a regular method override. I'm not sure if there is a way to declare it as a method instead.

rsuter:
v6.21

rsuter:
Using a method would be possible, but i think its fine as is...

admir86:
Is that right, now I have to extend each service class if I need to set the `jsonParseReviver` for each generated class. Is there not a easier solution?
for me the solution with the optional constructor parameter was perfect, the user was able to change the `jsonParseReviver` but it was optional. 
but maybe I have not understand the problem with this solution

RSuter:
In general I'd say that the user of the client should not be able to change the internal serialization logic as it should already be generated in a correct way otherwise it's a "leaky abstraction".

admir86:
ok I understand. 
but it's also bad if I have to create for each new class an extend. the problem I have my nswag.json is on a share and I use it for all my projects to be sure the settings are always the same. If I have to extend the settings (`ExtensionClasses`), with each new api controller I have to create the nswag.json for each project.

Is there not a solution to resolve this problem with a base class? maybe call a function of the base class to set the `jsonParseReviver` ?

RSuter:
Cant you set this in a common BaseClass?

admir86:
yes, I can, It's just for the date deserialization. should be always the same.

PR #402, opened 2016-11-02
Add Aurelia to list of frameworks supported in README.md

AshleyGrant:


rsuter:
:-) thanks for adding Aurelia to the list, in the template list it was already added... (removed)

AshleyGrant:
You're right. Not sure how I missed that. Cheers.

PR #421, opened 2016-11-16
Add NullHandling parameter to AssemblyTypeToSwaggerCommandBase

minichma:
The schema generated by the types2swagger command currently cannot be used with tools like swagger-codegen because the null-value handling is based on "oneOf" rather than on "requires" patterns. This could be configured via the NJsonSchema's NullHandling configuration setting.

This pull request exposes the NullHandling setting via the parameters of the AssymblyTypeToSwagger command, so it can be configured as needed.

### DIFF
--- a/src/NSwag.Commands/Commands/AssemblyTypeToSwaggerCommandBase.cs
+++ b/src/NSwag.Commands/Commands/AssemblyTypeToSwaggerCommandBase.cs
@@ -81,6 +81,13 @@ namespace NSwag.Commands
             set { Settings.GenerateKnownTypes = value; }
         }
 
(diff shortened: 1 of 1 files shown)


rsuter:
This shouldnt be a config but always swagger when the output is a swagger spec...

minichma:
This would certainly solve my issue. It seemed to me, that it was intentional to have types2swagger generate JsonSchema. But yes, as it effectively generates swagger, the configurability isn't really required. At least the default config should be swagger.

Want me to update my pull request? Guess, it might be easier, just to fix it?

rsuter:
I fixed it: https://github.com/NSwag/NSwag/commit/ea6a71387eb391552ef8cda6e9d66e82698e1167

rsuter:
Thank you for finding the bug...

minichma:
Thanks for fixing :-)

PR #423, opened 2016-11-17
Explicitly allow undefined as type of jsonParseReviver

grovesNL:
Added to Fetch/Angular templates

### DIFF
--- a/src/NSwag.CodeGeneration/CodeGenerators/TypeScript/Templates/Angular2ClientTemplate.cs
+++ b/src/NSwag.CodeGeneration/CodeGenerators/TypeScript/Templates/Angular2ClientTemplate.cs
@@ -259,7 +259,7 @@ if(Model.GenerateClientInterfaces){
             this.Write(@"{
     private http: Http = null; 
     private baseUrl: string = undefined; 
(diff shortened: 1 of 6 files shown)


grovesNL:
This is useful when using [`strictNullChecks`](https://www.typescriptlang.org/docs/release-notes/typescript-2.0.html)

rsuter:
I think this breaks for all users of TS 1.8?

grovesNL:
@rsuter You're right, I didn't realize `undefined` was only introduced as a type in 2.0. I'm not sure what to do without breaking 1.8 compatibility, unless I can opt in to 2.0+ functionality or diverge the templates. Currently the compiler will fail on that line for any project using `--strictNullChecks`, so anybody using it would have to post-process or create a separate template to modify that line (which I don't mind doing, but there may be others that hit the same issue).

rsuter:
Added TypeScriptVersion setting: https://github.com/NJsonSchema/NJsonSchema/commit/7bfa8b3d5a609db70403310bc510434e7eac1828

grovesNL:
@rsuter Great, thanks

rsuter:
Feature added (see commit)

rsuter:
v7.3

PR #490, opened 2016-12-23
Fixed issue where parameter from operation did not get merged down in…

PeterHenell:
…to the ActualParameter property.

### DIFF
--- a/src/NSwag.Core/SwaggerOperation.cs
+++ b/src/NSwag.Core/SwaggerOperation.cs
@@ -80,7 +80,9 @@ namespace NSwag
             get
             {
                 var allParameters = Parent?.Parameters == null ? Parameters :
-                    Parameters.Concat(Parent.Parameters.Where(p => Parameters.All(op => op.Name != p.Name && op.Kind != p.Kind)));
+                    Parameters.Concat(Parent.Parameters)
+                    .GroupBy(p => p.Name + "|" + p.Kind)
+                    .Select(p => p.First());
 
                 return new ReadOnlyCollection<SwaggerParameter>(allParameters
                     .Select(p => p.ActualSchema is SwaggerParameter ? (SwaggerParameter)p.ActualSchema : p)


rsuter:
I dont think this is correct. With this change, you'll end up with just the parameters with the first name and kind and the others are thrown away... Just the opposite which is expected from this property...

PeterHenell:
Not exactly sure what you mean. 
This will take the unique set of parameters from both Parameters and Parent.Parameters, based on the Name and Kind.

The previous code would exclude parameters from parent if it had the same kind as the child.

```
 static void Main(string[] args)
        {
            var parentParameters = new[]
            {               
                new { Name = "SecureToken", Kind = "Header", MoreProperties="This is my secure token", Rand=Guid.NewGuid() },
                new { Name = "ShouldBePartOfAllParameters", Kind = "Query", MoreProperties="There is no reason this should not be part of the final result", Rand=Guid.NewGuid() }
            };

            var Parameters = new[]
            {
                new { Name = "ID", Kind = "Header", MoreProperties="There can be other fields", Rand=Guid.NewGuid() }
            };

            var allParameters =
                Parameters.Concat(parentParameters)
                    .GroupBy(p => p.Name + "|" + p.Kind)
                    .Select(p => p.First());

            Console.WriteLine("Result");
            foreach (var p in allParameters)
            {
                Console.WriteLine(p);
            }
            
            var prevCodeResult = Parameters.Concat(parentParameters.Where(p => Parameters.All(op => op.Name != p.Name && op.Kind != p.Kind)));
            Console.WriteLine("Previous Code Result");
            foreach (var p in prevCodeResult)
            {
                Console.WriteLine(p);
            }
        }
```

```
Result
{ Name = ID, Kind = Header, MoreProperties = There can be other fields, Rand = 3b64797b-d229-42e0-ac76-9ed4a9326c95 }
{ Name = SecureToken, Kind = Header, MoreProperties = This is my secure token, Rand = 34765c70-b270-4a4f-8323-8a0457bb8da7 }
{ Name = ShouldBePartOfAllParameters, Kind = Query, MoreProperties = There is no reason this should not be part of the final result, Rand = eabf219e-0106-44ff-b2a4-9ecf3adcc789 }
Previous Code Result
{ Name = ID, Kind = Header, MoreProperties = There can be other fields, Rand = 3b64797b-d229-42e0-ac76-9ed4a9326c95 }
{ Name = ShouldBePartOfAllParameters, Kind = Query, MoreProperties = There is no reason this should not be part of the final result, Rand = eabf219e-0106-44ff-b2a4-9ecf3adcc789 }
```

PR #493, opened 2016-12-27
Enable binding a complex object array to webapi 2

aelbatal:
in relation with the following issue:
https://github.com/NSwag/NSwag/issues/492

### DIFF
--- a/src/NSwag.CodeGeneration/CodeGenerators/Models/ParameterModel.cs
+++ b/src/NSwag.CodeGeneration/CodeGenerators/Models/ParameterModel.cs
@@ -99,6 +99,11 @@ namespace NSwag.CodeGeneration.CodeGenerators.Models
             Schema.Item?.ActualSchema.Format == JsonFormatStrings.Date) &&
             _clientGeneratorBase.GetType(Schema.Item.ActualSchema, IsNullable, "Response") != "string";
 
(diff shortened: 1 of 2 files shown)


rsuter:
I you change this in this way, then for example string[] arrays are no longer supported, because it can only "serialize" complex objects... Your "special" foreach must only be applied for arrays of type object...

ahmedelbatal:
I agree. how about now? I modified it to solve both simple and complex array types.

rsuter:
The logic should be implemented in T4 and not in TypeScript because the array type is already available at generation time... Add a IsObjectArray (schema.Item.Type == JsonObjectType.Object) parameter to ParameterModel and use that...

rsuter:
Thank you for the PR... how can we test this? do you generate the Swagger spec with NSwag?

aelbatal:
Thank you for applying my edit. Unfortunately NSwag was hanging with some ApiControllers when I used it to generate the Swagger spec. I didn't spend much time to investigate this. So I'm now using the Swashbuckle Nuget Package to generate the Swagger spec, then passing it to `NSwag.SwaggerDocument.FromJsonAsync`.

rsuter:
And this works with Swashbuckle? Can you provide a sample (web api action and swagger operation) so that we can support this scenario in nswag too?

PR #526, opened 2017-01-14
export keyword for module and namespace blocks in FileTemplate.tt for TypeScript.

abogushevsky:
Hi guys!
First of all i want to thank you for your work on NSwag. I’ve looked at several WebApi to TypeScript generators and NSwag is the best. It's easy to use, and very flexible. Moreover it works with ASP.NET Core controllers - exactly what we need in our new project.
I would like to suggest you a little fix in FileTemplate.tt to  generate modules and namespace with export keyword. It is impossible now to import and use generated entities if module and/or namespace were specified in generator settings.

### DIFF
--- a/src/NSwag.CodeGeneration/CodeGenerators/TypeScript/Templates/FileTemplate.tt
+++ b/src/NSwag.CodeGeneration/CodeGenerators/TypeScript/Templates/FileTemplate.tt
@@ -24,10 +24,10 @@ import { HttpClient } from 'aurelia-fetch-client';
 <#}#>
 
 <#if(Model.HasModuleName){#>
-module <#=Model.ModuleName#> {
+export module <#=Model.ModuleName#> {
 <#}
   if(Model.HasNamespace){#>
-namespace <#=Model.Namespace#> {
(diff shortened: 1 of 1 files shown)


rsuter:
Doesnt this break some scenarios? Otherwise it is fine for me...

abogushevsky:
I'll read more about exporting modules and namespaces in TS and write a comment later. Do you now a way to use exported class from non exported namespace in another class from another namespace?

rsuter:
I dont know. I never used namespace or module in my generated code... For me, the file is the "module"

abogushevsky:
We a re not going to use modules too but we want to use namespaces to explicitly separate generated code from any other. That's why I tryed to use namespaces and found out that I can't import anything from generated file. I think that many other teams developing enterprise projects could decide to use namespaces and experience the same problem.

abogushevsky:
I've figured out that there is no need to export namespaces and/or modules. There was my misunderstanding of theese features in TypeScript. I apologize for wasting your time and closing this pull request.

rsuter:
Ok, thanks for looking into this. Enjoy using NSwag.

Issue #546, opened 2017-01-29
Custom ParameterBindingAttribute: More than one body parameter


domenichelfenstein:
I have a very simple T4 in which I just want to write the swagger output in a file using the NSwag.AssemblyLoader (see attached file).
In my WebApi I created a custom ParameterBindingAttribute (the reason is not of interest here).
NSwag seems to have a problem with this scenario because of the way it is implemented:
It searches for Query-Bindings and assumes that every other binding has to be a body parameter binding. And because there can only be one body parameter it refuses work correct when bindings of other types were found.
My proposal is to specifically search for body/query parameters and ignore all parameters that are not of the Type FromBodyAttribute or FromQueryAttribute.
[nswag.txt](https://github.com/NSwag/NSwag/files/737449/nswag.txt)

rsuter:
With the referenced commit, you can now add SwaggerIgnoreAttribute to the parameter so that is ignored, you can also inherit from SwaggerIgnoreAttribute in your custom attribute so that it is always ignored (just create a class SwaggerIgnoreAttribute in your project and inherit from it).

rsuter:
Is this fine for you?

domenichelfenstein:
Thanks! This works for me as well. As long as I can run NSwag without throwing Exceptions I'm fine with it.

domenichelfenstein:
Will there be a preview build available in nuget that enables me to use this improvement?

rsuter:
The CI feed is here: https://www.myget.org/gallery/nswag-ci

domenichelfenstein:
perfect, thanks!

rsuter:
Can you close the issue when it's working for you?

PR #547, opened 2017-01-29
Update OperationParameterProcessor.cs

domenichelfenstein:
ignores parameters that have custom attributes

the reason for this can be seen in issue #546


rsuter:
Running this code may break some scenarios because we cannot know if all parameters with custom attributes must be ignored...

domenichelfenstein:
Thanks for reviewing!
You are probably right about my assumption.

On the other hand as of now the code
- misinterprets all custom attributed parameters as body parameters and therefore
- throws an InvalidOperationException (saying that class XYZ has more than one body parameter) when a parameter with a custom attribute is used alongside another parameter with a FromBody attribute

So I can only think of one scenario where a working setup wouldn't work after my code change:
When someone created a custom attribute that derives from the FromBody attribute. I don't know if somebody did that.

PR #549, opened 2017-01-30
Added inline delegate-configuration settings

weitzhandler:
Style Owin and .NET core.
I hope it'll compile, unsure these operators are supported by the .NET and C# version of the repo.


weitzhandler:
Doesn't belong here. Should make my change in the AspNetCore extensions file.

PR #550, opened 2017-01-30
Added inline delegate-configuration for settings

weitzhandler:


rsuter:
SwaggerJsonSchemaGenerator should not be exposed as public API and SwaggerUiOwinSettings inherits from JsonSchemaGeneratorSettings and thus you shouldn't create a second setting object of this type...

weitzhandler:
@rsuter 
I followed the [original method](https://github.com/NSwag/NSwag/blob/master/src/NSwag.AspNetCore/SwaggerExtensions.cs#L70)'s pattern.

rsuter:
Is my commit ok for you?

weitzhandler:
Sure,
I'm pretty new to NSwag anyway, so if you feel that access to the schema settings is redundant than it's just great.
BTW, can you please help me with [my issue](../issues/553)?

PR #552, opened 2017-01-31
Replace try/finally with using.

paulomorgado:
The `ClientTemplate.tt` template uses `try`/`finally` when it could/should use `using`.

Minor identation fixes.


rsuter:
This try/finally is used on purpose because in some scenarios (i.e. the client is injected or the return value is FileResponse) the client / response must not be disposed...

PR #578, opened 2017-02-08
Improved controller response generation

Egorikas:
These are improvments which we discussed If I understood you right. Thank you

https://github.com/NSwag/NSwag/issues/521

### DIFF
--- a/src/NSwag.CodeGeneration.CSharp/Templates/ControllerTemplate.tt
+++ b/src/NSwag.CodeGeneration.CSharp/Templates/ControllerTemplate.tt
@@ -35,7 +35,7 @@ public partial class <#=Model.Class#>Controller : <#if(Model.HasBaseClass){#><#=
 <#  }}if(operation.HasResultDescription){#>    /// <returns><#=ConversionUtilities.ConvertCSharpDocBreaks(operation.ResultDescription, 1)#></returns>
 <#  }if(operation.IsDeprecated){#>    [System.Obsolete]
 <#  }#>    [System.Web.Http.Http<#=operation.HttpMethodUpper#>, System.Web.Http.Route("<#=operation.Path#>")]
-    public <#if(Model.WrapResponses){#>System.Threading.Tasks.Task<HttpResponseMessage><#}else{#><#=operation.ResultType#><#}#> <#=operation.OperationNameUpper#>(<#foreach(var parameter in operation.Parameters){#><#=parameter.Type#> <#=parameter.VariableName#><#if(Model.GenerateOptionalParameters && parameter.IsOptional){#> = null<#}#><#if(!parameter.IsLast){#>, <#}}#>)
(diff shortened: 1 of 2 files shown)


src/NSwag.CodeGeneration.CSharp/Templates/ControllerTemplate.tt
```diff
@@ -27,16 +27,30 @@ public partial class <#=Model.Class#>Controller : <#if(Model.HasBaseClass){#><#=
...
 <#      if(parameter.HasDescription){#>    /// <param name="<#=parameter.VariableName#>"><#=ConversionUtilities.ConvertCSharpDocBreaks(parameter.Description, 1)#></param>
 <#  }}if(operation.HasResultDescription){#>    /// <returns><#=ConversionUtilities.ConvertCSharpDocBreaks(operation.ResultDescription, 1)#></returns>
 <#  }if(operation.IsDeprecated){#>    [System.Obsolete]
 <#  }#>    [System.Web.Http.Http<#=operation.HttpMethodUpper#>, System.Web.Http.Route("<#=operation.Path#>")]
-    public <#=operation.ResultType#> <#=operation.OperationNameUpper#>(<#foreach(var parameter in operation.Parameters){#><#=parameter.Type#> <#=parameter.VariableName#><#if(Model.GenerateOptionalParameters && parameter.IsOptional){#> = null<#}#><#if(!parameter.IsLast){#>, <#}}#>)
-    {
+    public <#if(Model.WrapResponses){#>System.Threading.Tasks.Task<HttpResponseMessage><#}else{#><#=operation.ResultType#><#}#> <#=operation.OperationNameUpper#>(<#foreach(var parameter in operation.Parameters){#><#=parameter.Type#> <#=parameter.VariableName#><#if(Model.GenerateOptionalParameters && parameter.IsOptional){#> = null<#}#><#if(!parameter.IsLast){#>, <#}}#>)
```
rsuter: Missing async keyword

src/NSwag.CodeGeneration.CSharp/Templates/ControllerTemplate.tt
```diff
@@ -27,16 +27,30 @@ public partial class <#=Model.Class#>Controller : <#if(Model.HasBaseClass){#><#=
...
+    {	
+<#          if (Model.WrapResponses){#>
+        var result = await _implementation.<#=operation.OperationNameUpper#>Async(<#foreach(var parameter in operation.Parameters){#><#=parameter.VariableName#><#if(!parameter.IsLast){#>, <#}}#>);
+
+	    var status = (HttpStatusCode)Enum.Parse(typeof(HttpStatusCode), result.StatusCode);
```
rsuter: Should also use Swagger spec status code when not set in response obj

Egorikas:
I'll fix this tommorow

src/NSwag.CodeGeneration.CSharp/Templates/ControllerTemplate.tt
Egorikas: I don't clearly understand, what I need to do. As I understand when we use a custom response class, we always have "Status" property. In another hand, when we don't use our cusom response, the implementation will be defined as return _implementation.SomeMethod. Can you explain this problem or say where I can see code solving similar situation

rsuter: I think its fine as is...

PR #588, opened 2017-02-12
Fetch typescript fixes

bakesteve:
Fixes to prevent client errors in Typescript with implicit any disabled.
Also prevents errors for assigning the result200 variable to null


bakesteve:
@rsuter one note, I didnt commit the .cs files, only the .tt, as I had the source in a different location so they were committing my directory structure inside the .cs file
I assumed that was ok as tehy would be compiled as part of the build process, but looking at master that isnt the case. not sure if that is ok for nuget etc, but if not, then you might need to go in and save these
apologies if that was detailed in a readme/contrib file somewhere

PR #606, opened 2017-02-20
Restore setter for `TypeScriptGeneratorSettings`

### LINKED ISSUE #605: TypeScriptGeneratorSettings setter has been removed
The setter for `TypeScriptGeneratorSettings` seems as though it was mistakenly removed in https://github.com/NSwag/NSwag/commit/ff9a2eea967b713bb22552c55a5db301c4053cab

grovesNL:
Fixes #605


grovesNL:
Behavior appears to be intentional, added comment to #605

rsuter:
I removed the setter to make sure that the NullHandling is always Swagger (exception you change it intentionally): 

https://github.com/NSwag/NSwag/blob/master/src/NSwag.CodeGeneration.TypeScript/SwaggerToTypeScriptClientGeneratorSettings.cs#L26

If the setter is public, you can create a new setting instance and forget to set the NullHandling to Swagger which causes a lot of problems...

PR #607, opened 2017-02-20
Optimize string manipulation

paulomorgado:
Pull parameter validation to the top of the method and use a StringBuilder to generate URL.

The generated .NET clients use lots of string instantiations. This pull request replaces those string instantiations with a StringBuilder.

### DIFF
--- a/src/NSwag.CodeGeneration.CSharp/Templates/ClientTemplate.tt
+++ b/src/NSwag.CodeGeneration.CSharp/Templates/ClientTemplate.tt
@@ -86,8 +86,8 @@ public partial interface I<#=Model.Class#>
     }
 
 <#}#>
-    partial void PrepareRequest(System.Net.Http.HttpClient request, ref string url);
-
+    partial void PrepareRequest(System.Net.Http.HttpClient request, string url);
+    partial void PrepareRequest(System.Net.Http.HttpClient request, System.Text.StringBuilder urlBuilder);
     partial void ProcessResponse(System.Net.Http.HttpClient request, System.Net.Http.HttpResponseMessage response);
 
 <#foreach(var operation in Model.Operations){#>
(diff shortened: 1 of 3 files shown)


rsuter:
Please see my review

paulomorgado:
Where?

src/NSwag.CodeGeneration.CSharp/Templates/ClientTemplate.tt
```diff
@@ -86,7 +86,7 @@ public partial interface I<#=Model.Class#>
     }
 
 <#}#>
-    partial void PrepareRequest(System.Net.Http.HttpClient request, ref string url);
+    partial void PrepareRequest(System.Net.Http.HttpClient request, System.Text.StringBuilder urlBuilder);
```
paulomorgado: The only breaking change I saw was on `PrepareRequest` but, it the user needs to make changes to the url, a `StringBuilder` is better than a `string`.

There's nothing wrong with having:

    var url_ = rulBuilder_.ToString();
    ...
    request_.RequestUri = new System.Uri(url_, System.UriKind.RelativeOrAbsolute);

The compiler/runtime will, most like it, optimize it away. But, what's the point of that?

I choose to keep the breaking change as I've seen others (like `WrapSuccessResponses` to  `WrapResponses`)

src/NSwagStudio/NSwagStudio.csproj
```diff
@@ -15,6 +15,21 @@
     <WarningLevel>4</WarningLevel>
```
paulomorgado: Sorry. It wasn't supposed to have changed.

Dropped the change and rebased from current master.

src/NSwag.CodeGeneration.CSharp/Templates/ClientTemplate.tt
rsuter: I think we shouldn't break this because the feature was available for more than a year... WrapSuccessResponses has been renamed two weeks after its introduction and an opt-in feature which makes this breaking change far less problematic.

paulomorgado: `PrepareRequest` implementations that are modifying the URL are one of the benefiters from the change.

I've made a change that brings the new and keeps the old in a compatible way.

src/NSwag.CodeGeneration.CSharp/Templates/ClientTemplate.tt
```diff
@@ -126,44 +126,43 @@ public partial interface I<#=Model.Class#>
...
 <#  }else if(parameter.IsArray){#>
-            foreach (var item_ in <#=parameter.VariableName#>) { url_ += string.Format("<#=parameter.Name#>={0}&", System.Uri.EscapeDataString(item_.ToString())); }
+foreach (var item_ in <#=parameter.VariableName#>) { urlBuilder_.Append("<#=parameter.Name#>=").Append(System.Uri.EscapeDataString(item_.ToString())).Append("&"); }
 <#  }else{#>
-            url_ += string.Format("<#=parameter.Name#>={0}&", System.Uri.EscapeDataString(<#=parameter.VariableName#><#if(parameter.Type.EndsWith("?")){#>.Value<#}#>.ToString()));
+urlBuilder_.Append("<#=parameter.Name#>=").Append(System.Uri.EscapeDataString(<#=parameter.VariableName#><#if(parameter.Type.EndsWith("?")){#>.Value<#}#>.ToString())).Append("&");
 <#  }#>
 <#}#>
+<#if(hasQueryParameters){#>
+        urlBuilder_.Length--;
```
rsuter: I think this is wrong, because it is only generated when there is no ? or &... 
Maybe this is better: 

    urlBuilder.TrimEnd('&').TrimEnd('?')

paulomorgado: Hum! Do you have any concrete case where it fails?

paulomorgado: Do you have any concrete case where it fails?

Where does that `TrimEnd` come from?

rsuter: I hoped that StringBuilder has this method :-)

paulomorgado: Well, it doesn't. :-(

But, what's wrong with just removing the last character when you know it's not needed there?

rsuter: No, i checkt your changes and they are absolutely fine... My loptop got no more juice but ill merge soon

paulomorgado: I'm trying to figure out a way to unit test the execution of the generated code. I'll let you know if I came up with something.

rsuter: For c# there are already some unit tests against the generated clients... But ideally we should have many more and also for the typescript clients

rsuter:
Ok, thank you for the PR

PR #632, opened 2017-02-26
Update TS tools and settings

### LINKED ISSUE #631: Updating TypeScriptTools to latest...
screen shot 2017-02-26 at 4 23 38 pm

Not sure if it's possible for us to update to the latest. I had issues finding the version you're currently using so I've been disabling working with the TypeScript project

grovesNL:
- Updated TS tools (should fix #631)
- Move tsconfig.json back into csproj for NSwag.Integration.TypeScriptWeb
- Update emitted JS/source maps in NSwag.Demo.Web (should fix https://github.com/NSwag/NSwag/issues/625#issuecomment-282590798)


rsuter:
The ci build is broken. It just compiles the sln but the TypeScript has errors. Is there a way to fix that so that it compiles everything?

grovesNL:
@rsuter Can we exclude the TypeScript projects from the builds for now? The TypeScript compiler issues appear as actual errors and not warnings. Eventually after all the problems are corrected we can build them with CI.

The other option is to move it back to tsconfig and disable TypeScript compilation in the project. Although that's basically the same as disabling the project anyway.

grovesNL:
I've added another configuration to the solution to compile NSwag.Integration.TypeScriptWeb with strict compiler options. We can use that temporarily until all of the build errors are resolved.

The remaining errors in CI should be caused by AppVeyor not running `npm install` in that project. Could we run `npm install` in the NSwag.Integration.TypeScriptWeb directory prior to the build step? Maybe something like `npm --prefix C:\projects\nswag-25x6o\src\NSwag.Integration.TypeScriptWeb install C:\projects\nswag-25x6o\src\NSwag.Integration.TypeScriptWeb`

rsuter:
Ok this seems to work: https://ci.appveyor.com/project/rsuter/nswag-25x6o/build/2045

Issue #634, opened 2017-02-27
Add support for allowing all parameter types, complex or primitive, to be defined by Type Mappers.


Cephei:
This is a byproduct of issue #621.

Example:
```
public Task<HttpResponseMessage> PostRevision(
            string documentId, 
            [MaybeMultipart]DocumentRevisionDto revisionDto,
            [DocumentCreationSource][NonBody]DocumentCreationSource source = null,
            string sourceId = null, 
            string transactionId = null)
```

Here,
- `[MaybeMultipart]` creates a custom HttpParameterBinder that reads from the body. Thus should be considered a SwaggerParameterKind.Body.
- `[DocumentCreationSource]` creates a custom HttpParameterBinder that reads from a primitive query string parameter, `source` that has constraints source type enumerations (e.g. `source=[word|powerpoint|excel|googledrive|onedrive]`)

`DocumentCreationSource` is a complex object that can be created from a string; hence the custom binder. According to the OpenAPI spec, you can only use a `schema` to describe `body` parameters. However, you can still describe all other parameters using many JsonSchema fields.

`SwaggerGenerator` currently calls `GenerateAndAppendSchemaFromTypeAsync` when creating body parameters. This appends the associated schema. I'm looking into a solution for parameters that _may_ be complex types _described_ as primitive types. An interesting find in NJsonSchema is that `PrimitiveTypeMapper` does *not* add the schema to the `SchemaResolver`, whereas `ObjectTypeMapper` does.

Cephei:
Continuing the last thought expressed:

OpenAPI actually does support non-referenced schema body types. They provide an example:

```
{
  "name": "user",
  "in": "body",
  "description": "user to add to the system",
  "required": true,
  "schema": {
    "type": "array",
    "items": {
      "type": "string"
    }
  }
}
```

`return new JsonSchema4 { SchemaReference = SchemaResolver.GetSchema(type, false) };` will not handle this case correctly.

Cephei:
The fix for this is as follows: OWIN/WebAPI only.

For custom `ParameterBinderAttribute` implementations, you must decorate either the parameter or the `ParameterBinderAttribute` with a new attribute that is defined:

```
[AttributeUsage(AttributeTargets.Class | AttributeTargets.Parameter)]
public class WillReadBodyAttribute : Attribute
{
    public WillReadBodyAttribute(bool willReadBody = false)
    {
        WillReadBody = willReadBody;
    }

    public bool WillReadBody { get; }
}
```

[...]

Issue #621, opened 2017-02-24
Detecting proper SwaggerParameterKind from multiple custom parameter binders.


Cephei:
I'm running into complications due to `OperationParameterProcessor.EnsureSingleBodyParameter`. Unfortunately, aside from `[FromBody]`, `[FromUri]` etc, it's difficult to determine whether a custom parameter binder will read from the body unless you have an instance of it and check the `HttpParameterBinding.WillReadBody`.

Example Problem:
```
public Task<HttpResponseMessage> PostRevision(
            string documentId, 
            [MaybeMultipart]DocumentRevisionDto revisionDto,
            [DocumentCreationSource]DocumentCreationSource source = null,
            string sourceId = null, 
            string transactionId = null)
```

Here, `[MaybeMultipart]` create a custom parameter binder that reads from the body. `[DocumentCreationSource]`, however, does *not* read from the body, but is a custom parameter binder.

One option could be to create a dummy attribute, something like `[NonBodyAttribute]`:
```
public Task<HttpResponseMessage> PostRevision(
            string documentId, 
            [MaybeMultipart]DocumentRevisionDto revisionDto,
            [DocumentCreationSource][NonBody]DocumentCreationSource source = null,
            string sourceId = null, 
            string transactionId = null)
```

Then in `OperationParameterProcessor.ProcessAsync` add a new check:
```
dynamic nonBodyAttribute = parameterAttributes.SingleOrDefault(a => a.GetType().Name == "NonBodyAttribute");
```
... and update the conditional.

Other than that, or removing the `EnsureSingleBodyParameter` validation , I'm at a loss of ideas. This will break for all custom parameter binders.

rsuter:
I think you actually have an instance of the attribute and you can check HttpParameterBinding.WillReadBody (but you have to do it via reflection).

rsuter:
Sorry, posted in wrong issue :)

Cephei:
Unfortunately that won't work. `HttpParameterBinding.WillReadBody` is on the actual binding. The attribute has no such property; it just acts as a factory method to create the binding.

Cephei:
I also can't remove `EnsureSingleBodyParameter` like earlier suggested. In our API we have multiple custom binders/attributes. Some read from the body, others not. It's unfortunate Microsoft didn't move WillReadBody to the attribute. The only thing I can think of now is to create a new attribute like suggested above.

PR #636, opened 2017-02-27
Adds support for custom HttpParameterBinders and Type Mapped primitives.

Cephei:
#621 #634


Cephei:
Not sure why the integration tests are failing here. @rsuter can you please assist when you have time. Thanks!

grovesNL:
@Cephei I think your build should be ok. They're just compilation errors from TypeScript that I'm working on fixing in https://github.com/NSwag/NSwag/pull/632

rsuter:
Looks good, thanks. But is there really no way to instantiate this factory and read the property? I'd like to avoid adding a new NSwag attribute...

Cephei:
I so wish that was possible. I hated this being the solution. Here's an example to help illustrate why it's not:

```
    [AttributeUsage(AttributeTargets.Class | AttributeTargets.Parameter)]
    public sealed class DocumentCreationSourceAttribute : ParameterBindingAttribute
    {
        public override HttpParameterBinding GetBinding(HttpParameterDescriptor parameter)
        {
            if (parameter == null)
            {
                throw new Exception("parameter");
            }

            if (parameter.Configuration.DependencyResolver == null)
            {
                return null;
            }

            return new DocumentCreationSourceParameterBinding(
                parameter, 
                parameter.Configuration.DependencyResolver.GetService(typeof(IDocumentCreationSourceProvider)) as IDocumentCreationSourceProvider);
        }
    }
```

The actual parameter binder:
```
    public class DocumentCreationSourceParameterBinding : HttpParameterBinding
    {
        private static readonly Task<Object> _CompletedTaskReturningNull = Task.FromResult<Object>(null);

        private readonly IDocumentCreationSourceProvider _Provider;

        public DocumentCreationSourceParameterBinding(
            HttpParameterDescriptor parameter,
            IDocumentCreationSourceProvider provider) : base(parameter)
        {
            _Provider = provider;
        }

        public override Task ExecuteBindingAsync(
            ModelMetadataProvider metadataProvider,
            HttpActionContext actionContext,
            CancellationToken cancellationToken)
        {
            var source = actionContext.Request
                .GetQueryNameValuePairs()
                .MaybeSingle(kvp => kvp.Key.Equals("source", StringComparison.OrdinalIgnoreCase) && !String.IsNullOrWhiteSpace(kvp.Value))
                .Bind(kvp => _Provider.Find(kvp.Value).ToMaybe())
                .FromMaybe(null);

            actionContext.ActionArguments[Descriptor.ParameterName] = source;

            return _CompletedTaskReturningNull;
        }
    }
```

Cephei:
Also, I made it so they don't have to use the NSwag.Attributes.WillReadBodyAttribute. They can use their own WillReadBodyAttribute as long as it follows convention.

rsuter:
Why do we need a WillReadBody property on the attribute? I think the attribute alone should be alone to mark a parameter/attribute that it reads the body (like [Required]). Or at least the default should be "true"

rsuter:
2. Why do you disable this for Asp.NET Core? Doesn't it make sense there?

Cephei:
*WillReadBody Property*
I considered this as well, and the original intentions was to support .net core ... before learning more about Core. So Owin/WebAPI differs from Core. Complex types in WebAPI are assumed to read from the body. Whereas complex types in Core are bound from the URI (reflectively looking at properties).

If I don't expose this property, WebAPI would automatically think my `DocumentCreationSourceAttribute` will read from the body. Without this property, I would need to have a [NonBody] attribute. I didn't like that.

*Asp.net Core*
I'm very new to .net core but tried to do as much research on this as I could. In Core, there is no more ParameterBindingAttribute or HttpParameterBinder. Core uses ModelView binding which you're going to have an even larger challenge supporting. Aside from the attributes you already support (e.g. FromUri, FromBody, FromRoute, FromHeader) you can implement your own IModelBinder / IModelBinderProvider and add them to your `MvcOptions`. (see https://www.stevejgordon.co.uk/html-encode-string-aspnet-core-model-binding)

These model binders are centralized much how Owin/WebAPI actually supports centralized parameter bindings (which will be a problem for me soon). Swashbuckle leverages your `HttpConfiguration` to create OpenAPI documentation. The `HttpConfiguration` instance providers you all your route information and access to centralized model binders. I originally loved that NSwag didn't require `HttpConfiguration` and just ApiController Type's or assemblies, however, that comes at a cost. Something to consider going forward ... parameter/model binding does not require attributes. They can be part of the core model binder collection that can't be discovered using reflection.

Cephei:
Example of centralized model binding in WebAPI / OWIN:

```
// NOTE HERE
// configuration is an instance of HttpConfiguration

// --------------------------------------------- ModelBinders --------------------------------------
            var bindingRules = configuration.ParameterBindingRules;
            bindingRules.Add(param =>
            {
                if (!param.ParameterType.IsGenericType) return null;
                if (!typeof(PatchRequestBase).IsAssignableFrom(param.ParameterType)) return null;

                return param.BindWithAttribute(new PatchRequestParameterBindingAttribute());
            });

            bindingRules.Add(
                param =>
                {
                    if (!(typeof(QueryOptions).IsAssignableFrom(param.ParameterType)))
                    {
                        return null;
                    } 

                    return param.BindWithAttribute(new QueryOptionsParameterBindingAttribute());
                });
```

rsuter:
**WillReadBody Property**

Ok, then its fine. Can we at least have the default "true"?

**ASP.NET Core**

Maybe we should also use ApiExplorer for ASP.NET Core and ASP.NET... I think there shouldn't be a downside to that except more dependencies... Or is this a bad idea except that it will be time consuming? Is Swashbuckle working better out of the box or with extenisons?

Cephei:
We can default it to true. I was torn on that but since HttpParameterBinding.WillReadBody defaults to false, I went with that.

Cephei:
Better support for .NET Core / WebAPI would require a bit of a redesign around NSwag, and more dependencies like you stated, but I do think it would be worth it. One of the reasons I loved NSwag was that I have multiple HttpConfiguration's within an Owin pipeline:

/v{apiVersion}/ -> main AppBuilder for the entire API
/scim/v{apiVersion}/ -> SCIM routes from Owin.Scim (project)

Ideally, I'd love to be able to say something like:
```
public class OpenApiController : ApiController
{
    private readonly IEnumerable<HttpConfiguration> _configurations;

    public OpenApiController(IEnumerable<HttpConfiguration> configurations)
    {
        _configurations = configurations.ToList();
    }

    [Route("swagger/v{apiVersion}")]
    public async Task<HttpResponseMessage> GetSwagger(uint apiVersion)
    {
        var settings = new WebApiToSwaggerGeneratorSettings
        {
....
        };

        var generator = new WebApiToSwaggerGenerator2(settings);
        var swagger = await generator.GenerateForControllersAsync(_configurations);

        return Request.CreateResponse(HttpStatusCode.OK, swagger);
    }
}
```

That being said, I am looking at transitioning my company's product to Core this year. I'm not nearly as familiar with the Core pipeline in comparison to OWIN.

Cephei:
This is something I can look into though once I get further along with NSwag. Swashbuckle worked nicely out of the box and once I got going with it, however, I soon realized it's completely disconnected from JsonSchema. The Swashbuckle author also seems not interested in supporting polymorphism (which sucks in OpenAPI v2, but fully supported in the upcoming v3). The lack of inheritance and polymorphism support made me leave Swashbuckle entirely.

rsuter:
I think we should remove the default entirely - a good compromise I think...

Cephei:
Would it be reasonable to get this merged into CI today?

Issue #622, opened 2017-02-24
Known Response Types throw Exception with wrapped responses


graemechristie:
The current implementation of Client generator will throw exceptions when the response code is not 2XX, even when the wrapResponses option is set to true. It is perfectly valid and not exceptional for a swagger api to return a non 2xx response code.

The canonical example is returning a 404 code, when a resource is not found. This is the behaviour of most of our Api's when a user requests a specific resource that is not found.

e.g. the swagger document will contain:

        "responses": {
          "200": {
            "description": "Success",
            "schema": { "$ref": "#/definitions/Transaction" }
          },
          "404": {
            "description": "Not Found",
            "schema": { "$ref": "#/definitions/Transaction" }
          }

Having to wrap every request in a try {} catch block just to determine if the item is not found is extremely messy, and what I would consider code smell. Exceptions should occur when an error condition occurs on the current code path, not when an expected response (this response is defined in the known responses of the API, so it is expected) occurs.

I realise changing this behaviour may be considered a breaking change (although I cannot imagine a real world use case for the wrapResponses option, if only 200 status codes are handled) . If I was to submit a PR to address this, would you expect it to be behind a configuration variable of some kind ?

rsuter:
The idea of this setting is to give access to some http level data of the response (ie http headers). The problem is that a method can only have one return type. In your case both responses have the same type so it would be possible to merge them... But with what rule? For me, your sample is fine, i would expect an exception on 404...

Question: What would be the method return type if 200 returns an instance of foo and 404 bar?

How to detect successful vs exception responses?

graemechristie:
In the case where a 404 was a known response type, I would not expect an exception to be thrown. The response was an expected condition where a resource has not been found, the service advertised that the response should be expected. It's a normal part of the application flow. 

Currently we would have to wrap most of our service calls in try/catch blocks. Using exceptions to manage normal control flow is pretty bad practice (not to mention very verbose and inelegant), and I wouldn't ask my team to go down that path.

I'd just like to note that I would only expect that multiple responses codes/types would be supported for Wrapped Responses (or behind another feature flag if you want to leave wrapped responses as they are). For simple interfaces where the return from the client if the DTO object your current logic would be used

> Question: What would be the method return type if 200 returns an instance of foo and 404 bar?

[...]

PR #667, opened 2017-03-10
TypeScript response code handling bug

sandord:
Generating typescript client code for the controller below, resulted in exception throw statements for status codes that are not errors.

    public class FooController : ApiController
        {
            [Route("foos/")]
            [SwaggerResponse(200, null)]
            [SwaggerResponse(201, null)]
            [SwaggerResponse(400, null)]
            [SwaggerResponse(500, null)]
            public Foo[] GetFoos([FromUri] Bar[] bars)
            {
                return new Foo[0];
            }
        }

Before:

    protected processGetFoos(response: Response): void {
        const responseText = response.text();

        if (status === 200) {
            return null;
        } else if (status === 201) {
            this.throwException("A server error occurred.", status, responseText);
        } else if (status === 400) {
            this.throwException("A server error occurred.", status, responseText);
        } else if (status === 500) {
            this.throwException("A server error occurred.", status, responseText);
            this.throwException("An unexpected server error occurred.", status, responseText);
        }
        return null;
    }

Clearly, a HTTP 201 is not supposed to be handled as an error.

After:

    protected processGetFoos(response: Response): void {
        const responseText = response.text();

        if (status === 200) {
            return null;
        } else if (status === 201) {
            return null;
        } else if (status === 400) {
            this.throwException("A server error occurred.", status, responseText);
        } else if (status === 500) {
            this.throwException("A server error occurred.", status, responseText);
            this.throwException("An unexpected server error occurred.", status, responseText);
        }
        return null;
    }

### DIFF
--- a/src/NSwag.CodeGeneration.CSharp/Models/CSharpOperationModel.cs
+++ b/src/NSwag.CodeGeneration.CSharp/Models/CSharpOperationModel.cs
@@ -152,7 +152,7 @@ namespace NSwag.CodeGeneration.CSharp.Models
         /// <returns></returns>
         protected override CSharpResponseModel CreateResponseModel(string statusCode, SwaggerResponse response, JsonSchema4 exceptionSchema, IClientGenerator generator, ClientGeneratorBaseSettings settings)
         {
-            return new CSharpResponseModel(statusCode, response, response == GetSuccessResponse(), exceptionSchema, generator, settings.CodeGeneratorSettings);
+            return new CSharpResponseModel(statusCode, response, HttpUtilities.IsSuccessStatusCode(statusCode), exceptionSchema, generator, settings.CodeGeneratorSettings);
         }
     }
 }
--- a/src/NSwag.CodeGeneration.TypeScript.Tests/NSwag.CodeGeneration.TypeScript.Tests.csproj
+++ b/src/NSwag.CodeGeneration.TypeScript.Tests/NSwag.CodeGeneration.TypeScript.Tests.csproj
@@ -60,10 +60,15 @@
   </ItemGroup>
   <ItemGroup>
     <Compile Include="ArrayParameterTests.cs" />
+    <Compile Include="StatusCodeHandlingTests.cs" />
     <Compile Include="OperationParameterTests.cs" />
     <Compile Include="Properties\AssemblyInfo.cs" />
   </ItemGroup>
   <ItemGroup>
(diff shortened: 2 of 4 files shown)


sandord:
The same bug is fixed in `CSharpOperationModel` as well, though I didn't include a test for that.

rsuter:
This is not a bug but by design - what if 200 and 201 do not return the same response type? Then you would have not compilable code... Returning from 200 and 201 is only allowed if both have the same Response type...

sandord:
I get it.

I actually wasn't returing both 200 and 201 in my case but due to the fact that I used the `[ResponseType(typeof(Model))]` attribute as well (that's the one from the `System.Web.Http.Description` namespace), a 200 was added in addition to my 201.

So it turns out that this:

        public class FooController : ApiController
        {
            [Route("foos/")]
            [System.Web.Http.Description.ResponseType(typeof(Foo[]))]
            [SwaggerResponse(201, null)]
            public IHttpActionResult GetFoos([FromUri] Bar[] bars)
            {
                return Ok(new Foo[0]);
            }
        }

leads to:

    protected processGetFoos(response: Response): Foo[] {
        const responseText = response.text();

        if (status === 200) {
            let result200: Foo[] = null;
            let resultData200 = responseText === "" ? null : JSON.parse(responseText, this.jsonParseReviver);
            if (resultData200 && resultData200.constructor === Array) {
                result200 = [];
                for (let item of resultData200)
                    result200.push(Foo.fromJS(item));
            }
            return result200;
        } else if (status === 201) {
            this.throwException("A server error occurred.", status, responseText);
            this.throwException("An unexpected server error occurred.", status, responseText);
        }
        return null;
    }

I guess I'll be removing all `[ResponseType(typeof(...))]` declarations in my code base where ever I have a `[SwaggerResponse(...)]` as well.

As a final thought, wouldn't it be great if I had been warned by the client generator that multiple non-error responses were declared? The client code generated in that case just doesn't make much sense anyhow. I'd say: either allow it and generate sensible code or prohibit it.

grovesNL:
There was some discussion in #622 related to handling multiple non-exceptional response types

grovesNL:
@sandord How did you end up handling this?

sandord:
Since I wasn't really returning multiple success codes after all (a 200 snuck in due to the `ResponseModelAttribute`, next to my 201), I didn't have to.

Issue #639, opened 2017-02-28
RoutePrefix with optional parameter causes invalid Paths


Cephei:
Example:
```
[Route, RoutePrefix("users/{userId?}")]
public class UsersController : ApiController
{
    private readonly IUserManager _UserManager;

    public UsersController(IUserManager userManager)
    {
        _UserManager = userManager;
    }

    public Task<HttpResponseMessage> PostUser(UserDetailsDto userDto)
    {
        return _UserManager.CreateUser(userDto)
            .AwaitToHttpResponseMessageAsync(Request, HttpStatusCode.Created);
    }

    [Route("devices")]
    [AcceptVerbs("POST")]
    public Task<HttpResponseMessage> PostUserDeviceId(string userId, UserDeviceDto userDevice)
    {
        return _UserManager.AddUserDeviceId(userDevice.DeviceId, userDevice.Platform)
            .AwaitToHttpResponseMessageAsync(Request);
    }
}
```

swaggerDocument.Paths contains three paths.

rsuter:
There are still problems with optional path parameters...

rsuter:
So this was fixed with the route PR?

Cephei:
All my routing is working properly now but I'm slowly enabling swagger generation for more controllers endpoints as I document them. Only am working with two controllers now as I refine my documentation approach (not using xml or attribute annotations).

rsuter:
Ok thanks for your work here... I hope i can release next week. This will be a major version because we have some minor breaking changes... Do you see other important changes we should include in this release?

PR #683, opened 2017-03-17
Addressing several routing issues.

Cephei:
#639 

Several issues with routing.
Asp.Net Core vs Owin/WebAPI

Resources:
https://docs.microsoft.com/en-us/aspnet/core/fundamentals/routing
https://docs.microsoft.com/en-us/aspnet/core/mvc/controllers/routing
https://docs.microsoft.com/en-us/aspnet/web-api/overview/web-api-routing-and-actions/attribute-routing-in-web-api-2
https://docs.microsoft.com/en-us/aspnet/web-api/overview/web-api-routing-and-actions/routing-in-aspnet-web-api

The majority of changes are within `GetHttpPaths` & `ExpandOptionalHttpPathParameters`.
The logic used to be:
```
             .....
            else if (routePrefixAttribute != null && routeAttributeOnClass != null)
                httpPaths.Add(routePrefixAttribute.Prefix + "/" + routeAttributeOnClass.Template);
            // TODO: Check if this is correct
            else if (routePrefixAttribute != null && (
                (method.GetParameters().Length == 0 && method.Name == "Get") ||
                method.Name == "Post" ||
                method.Name == "Put" ||
                method.Name == "Delete"))
            {
                httpPaths.Add(routePrefixAttribute.Prefix);
            }
            else if (routeAttributeOnClass != null)
                httpPaths.Add(routeAttributeOnClass.Template);
            else
                httpPaths.Add(Settings.DefaultUrlTemplate ?? string.Empty);
```

The `Settings.DefaultUrlTemplate` should not be considered when a `RoutePrefix` is present as was in `PersonsController`. I was unable to find any routing rules that governed the following:
```
else if (routePrefixAttribute != null && (
                (method.GetParameters().Length == 0 && method.Name == "Get") ||
                method.Name == "Post" ||
                method.Name == "Put" ||
                method.Name == "Delete"))
```

I am allowed to have a `RoutePrefix` with a `GET` method which has parameters. That's completely valid. The last else condition should allow for the MVC style of routing with `{controller}/{action}/{id?}`.

Also the default route convention in both Core and Web API has the `{id}` as optional so I changed the `DefaultUrlTemplate` to reflect that.

Lastly, the fix for handling `RoutePrefix`:
I inverted the logic:
```
.Select(p =>
    "/" + p
    .Replace("[", "{")
    .Replace("]", "}")
    .Replace("{controller}", controllerName)
    .Replace("{action}", actionName)
    .Trim('/'))
.SelectMany(p => ExpandOptionalHttpPathParameters(p, method))
```
to first expand the optional paths. 
```
.SelectMany(p => ExpandOptionalHttpPathParameters(p, method))
.Select(p =>
    "/" + p
    .Replace("[", "{")
    .Replace("]", "}")
    .Replace("{controller}", controllerName)
    .Replace("{action}", actionName)
    .Trim('/'))
```
Before, if `ExpandOptionalHttpPathParameters` returned values, it would overwrite the the prior `Select` `"/" + p.Replace(...)` value.

Then within `ExpandOptionalHttpPathParameters` it no longer does two sequential recursive operations per-optional-route-segment (ie. `?}`):
```
if (segment.EndsWith("?}"))
                {
                    // Only expand if optional parameter is available in action method
                    if (method.GetParameters().Any(p => segment.StartsWith("{" + p.Name + ":") || segment.StartsWith("{" + p.Name + "?")))
                    {
                        foreach (var p in ExpandOptionalHttpPathParameters(string.Join("/", segments.Take(i).Concat(new[] { segment.Replace("?", "") }).Concat(segments.Skip(i + 1))), method))
                            yield return p;
                    }
                    else
                    {
                        foreach (var p in ExpandOptionalHttpPathParameters(string.Join("/", segments.Take(i).Concat(segments.Skip(i + 1))), method))
                            yield return p;
                    }

                    yield break;
                }
```

Before, you would get back multiple paths for a single endpoint with a `RoutePrefix`. Now you get the correct amount.


rsuter:
Looks like a big improvement. Thank you. I will test this with my projects...

Cephei:
Thanks bud. I'm testing it further as well with our SaaS product's API. Will keep you posted if anything arises. Unfortunately our product is still on Owin/WebAPI so I won't be able to examine in depth with Core just yet.

rsuter:
Do you see breaking changes in .NET Core? It is important to different between them in some places... I'll have a look...

Cephei:
Not that I can think of. What I'd like to do is have a complete suite of unit tests that encapsulate all WebAPI routing possibilities and a second suite for all of Asp.net Core. When I get some time I'll get started on this.

rsuter:
That would be awesome...

Just FYI: The .NET Core projects are compiled against the NSwag NuGet packages (they have to be published first). You cannot directly change the code and rerun the unit tests... You have to switch to project references first. I have to improve the tooling here and migrate all project to .NET Standard 1.0 (its already supported but not yet converted).

Cephei:
Oo, I definitely was not aware of that. What's the reason for that?

rsuter:
With the previous tooling (project.json) you could not reference a .NET PCL (old version of .NET Standard 1.0) from a .NET Core project. Now its possible, but I need to fix the build scripts so that the packages are correctly built.

Issue #638, opened 2017-02-28
Discuss changes for TypeScript strict checks


grovesNL:
Just wanted a place to document and discuss any changes we consider in the TypeScript clients as we add stricter checks in case the context behind the original types were missed.

Firstly [consider the `return null;` within a method expecting a return type of `void`](https://github.com/NSwag/NSwag/blob/master/src/NSwag.Integration.TypeScriptWeb/scripts/serviceClientsAngularJS.ts#L51). Do we intentionally return null here (in which case we should update the return type) or could this statement simply be `return;` instead? If this could be clarified I'd be happy to make the change, but I want to avoid breaking changes if `null` is actually expected.

I found where `return null;` was added to try to gain some context. It was added in 
https://github.com/NSwag/NSwag/commit/0edf24125a1775c4f8ad1f8f57d24308b9c36fe8#diff-1e10b65d13165f5da1c352bfcbceba19 but I can't tell whether this is the desired behavior.

grovesNL:
@rsuter Based on the above, could you take a look at https://github.com/NSwag/NSwag/compare/master...grovesNL:typescript-type-improvements when you have some time and let me know your thoughts? I can open a PR but I consider it experimental for now.

rsuter:
The return type must match the generated method signature (ie the signature wins).

grovesNL:
@rsuter But it some cases the signature that was used was impossible. i.e. [`processGetAll` claims to have a return type of `Parent[]`](https://github.com/NSwag/NSwag/blob/master/src/NSwag.Integration.TypeScriptWeb/scripts/serviceClientsAngularJS.ts#L490) but that's only true for one case. In several other cases it will return `null`, so the signature should be `Parent[] | null`.

Likewise because this return type is used in [`getAll`](https://github.com/NSwag/NSwag/blob/master/src/NSwag.Integration.TypeScriptWeb/scripts/serviceClientsAngularJS.ts#L464), `ng.IPromise<Person[]>` isn't true either. In fact there is another branch where there is no explicit return so that one is `undefined`. The new signature must then become `ng.IPromise<Person[] | null | undefined>`, unless we change it to `return null;` explicitly which will narrow it to `ng.IPromise<Person[] | null>`. But there is still a null case regardless.

rsuter:
[...]

PR #698, opened 2017-03-30
Typescript type improvements

grovesNL:
See #638

I haven't modified tsconfig yet as there are still a few areas (like `Blob` as discussed in Gitter) that should be improved first. You can enable `strictNullChecks`, `noUnusedLocals` etc. in order to see where these areas still exist.

We can leave the PR open if you want to continue making changes here, or we can merge it now and create separate PRs for incremental improvements. Let me know if there is anything wrong with the current set of commits.


grovesNL:
Actually I enabled `strictNullChecks` again. With `strictNullChecks` disabled, there are implicit any warnings for the return type of the various `process...` methods.

So it turns out we may have to use explicit return types to support the case when `noImplicitAny` is used but not `strictNullChecks` (which is too bad - I was hoping to avoid explicit return types if possible). I can restore the return types in this PR or open a separate one after this is merged.

rsuter:
I think ill merge this now and will look into fetch/blobs and you can create another pr, ok? Do you think its ok to only support blobs (ie browser compatibility)?

PR #713, opened 2017-04-10
Fix ignoring parameters

LordJZ:
In all other places the SwaggerIgnoreAttribute is matched exactly by name; but for parameters a type that inherits from SwaggerIgnoreAttribute is expected. This makes hiding parameters tedious as it requires creating a new attribute type and produces inconsistency in the way that annotations are handled. This commit makes annotation handling consistent and allows [SwaggerIgnore] on parameters.


rsuter:
Good find, thanks. But we should change it to (everywhere): 

.AssignableTo("SwaggerIgnoreAttribute", TypeNameStyle.Name)

rsuter:
... but yes, your change is still better than the current version :-)

PR #779, opened 2017-05-31
Bugfixes for when Model.WrapResponses is enabled

eihag:
4 fixes:
1) Fully qualified System.Net.HttpStatusCode
2) Fully qualified System.Enum
3) Using System.Net.Http to access extension method in Request.CreateResponse (line 48)
4) Use "Add" function for response headers to resolve compilation issue.

### DIFF
--- a/src/NSwag.CodeGeneration.CSharp/Templates/ControllerTemplate.tt
+++ b/src/NSwag.CodeGeneration.CSharp/Templates/ControllerTemplate.tt
@@ -1,6 +1,10 @@
 ﻿<#@ template visibility="internal" #>
 <#@ import namespace="NJsonSchema" #>
 
+<# if (Model.WrapResponses){#>
+using System.Net.Http;
+<# }#>
+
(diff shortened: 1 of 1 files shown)


RSuter:
Thanks for the PR. Please to not import any namespaces like

    using System.Net.Http;

For the extension method, use the fully qualified type and call the static method...

eihag:
Yes, that was my first option. However, it seems a fully qualified type creates an ambiguous reference. Example:
System.Net.Http.HttpResponseMessage response = System.Net.Http.HttpRequestMessageExtensions.CreateResponse(Request, status, result.Result);

produces the following error:
Error	CS0433	The type 'HttpRequestMessageExtensions' exists in both 'System.Net.Http.Formatting, ...' and 'System.Web.Http, ...'	

The "using" line committed is a workaround. Currently, the generated file does not compile.

RSuter:
This is very strange - there must be a way to use the static extension method and avoid the using, ...

PR #783, opened 2017-06-05
Follow class hierarchy when locating RouteAttributes

hmflash:
* Fix InvalidOperationException (Sequence contains more than one
   element) when RouteAttribute is used at different inheritance
   levels.
 * Changed attribute finding code to walk the inheritance hierarchy
   and only look for attributes at each level, stopping once
   an attribute is found.
 * Added unit tests to demonstrate the behavior.

### DIFF
--- a/src/NSwag.SwaggerGeneration.WebApi.Tests/NSwag.SwaggerGeneration.WebApi.Tests.csproj
+++ b/src/NSwag.SwaggerGeneration.WebApi.Tests/NSwag.SwaggerGeneration.WebApi.Tests.csproj
@@ -115,6 +115,7 @@
     <Compile Include="Attributes\RouteTests.cs" />
     <Compile Include="Attributes\TagsTests.cs" />
     <Compile Include="ControllerClassesTests.cs" />
+    <Compile Include="RouteAttributeTests.cs" />
     <Compile Include="EnumListTests.cs" />
     <Compile Include="FileParameterTests.cs" />
     <Compile Include="FileResponseTests.cs" />
--- a/src/NSwag.SwaggerGeneration.WebApi.Tests/RouteAttributeTests.cs
+++ b/src/NSwag.SwaggerGeneration.WebApi.Tests/RouteAttributeTests.cs
@@ -1,0 +1,50 @@
+﻿using System.Threading.Tasks;
+using System.Web.Http;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace NSwag.SwaggerGeneration.WebApi.Tests
+{
+    [TestClass]
+    public class RouteAttributeTests
+    {
+       [Route("/api/[controller]")]
+       public abstract class BaseTestController : ApiController
+       {
+       }
+
+        public class FooController : BaseTestController
+        {
+            public void Post()
+            {
+            }
+        }
+
+        [Route("/api/other")]
+        public class BarController : BaseTestController
+        {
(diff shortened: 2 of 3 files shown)


RSuter:
Thank you for the pr. Will check and merge soon. Looks good.

RSuter:
Is the a.Template != null check not needed? Or did you miss it?

hmflash:
As far as I can tell, its not needed.  The RouteAttribute will not allow you to construct one with a null template.

https://github.com/ASP-NET-MVC/aspnetwebstack/blob/master/src/System.Web.Http/RouteAttribute.cs
https://github.com/ASP-NET-MVC/aspnetwebstack/blob/master/src/System.Web.Mvc/RouteAttribute.cs

RSuter:
BTW: In one of my projects I had the case where Template is null... https://github.com/NSwag/NSwag/commit/d7d0e24079dfa90162ffc237bc40bb1b27ded2fe

hmflash:
Good to know.  I did some more extensive unit testing and am observing a strange problem.

When trying to generate swagger for the following controller:
```
[AttributeUsage(AttributeTargets.Class)]
class CustomRouteAttribute : Attribute, IHttpRouteInfoProvider
{
	public CustomRouteAttribute(string template)
	{
		Template = template;
	}

	public string Template { get; }
	public int Order { get; } = 0;
	public string Name { get; set; }
}

[CustomRoute("api/[controller]")]
public class MyApiController : ApiController
{
	public void Post()
	{
	}
}
```

The `(dynamic)a).Template != null` portion of the LINQ is throwing this exception:
```
Microsoft.CSharp.RuntimeBinder.RuntimeBinderException: 'System.Attribute' does not contain a definition for 'Template'
    at CallSite.Target(Closure , CallSite , Object )
   at System.Dynamic.UpdateDelegates.UpdateAndExecute1[T0,TRet](CallSite site, T0 arg0)
   at NSwag.SwaggerGeneration.WebApi.WebApiToSwaggerGenerator.<>c.<GetRouteAttributes>b__20_3(Attribute a)
   at System.Linq.Enumerable.<>c__DisplayClass6_0`1.<CombinePredicates>b__0(TSource x)
   at System.Linq.Enumerable.WhereEnumerableIterator`1.MoveNext()
   at System.Linq.Enumerable.SingleOrDefault[TSource](IEnumerable`1 source)
   at NSwag.SwaggerGeneration.WebApi.WebApiToSwaggerGenerator.GetRouteAttribute(Type type)
   at NSwag.SwaggerGeneration.WebApi.WebApiToSwaggerGenerator.GetHttpPaths(Type controllerType, MethodInfo method)
```

It is very strange as the `ObjectExtensions.HasProperty(a, "Template")` is true (as it should be) and the debugger does show the attribute as having a `Template` property with non-null value.  Going to do some more digging.

RSuter:
Strange, but I saw a similar issue - try using `ObjectExtensions.TryGetPropertyValue("Template", null) != null` instead of HasProperty and the dynamic cast...

hmflash:
Ah ha!  So if a user's custom route attribute class is not public, calling `((dynamic)a).Template` results in a `RuntimeBinderException`.  Using `TryGetPropertyValue` does fix the problem and things work as expected. I can fix up the code and open a new PR for you to review.

RSuter:
This is strange and bad - all usages of dynamic may fail like this... But for now i think its fine. A pr would be awesome...

PR #784, opened 2017-06-07
Fix exception when using custom route attributes

hmflash:
* Fix potential `RuntimeBinderException` when using custom internal or private `IHttpRouteInfoProvider` attributes
 * `((dynamic)a).Template` throws when the user defined attribute is internal or private, so replace with normal reflection and use `PropertyInfo.GetValue()`
 * Add test to verify internal user defined route attributes work properly
 * Add test to verify route attrs w/null templates are skipped

I looked at the other dynamic usage in the `NSwag.SwaggerGeneration.WebApi` assembly and I think it is ok.  The other usage is doing a.GetType().Name prior to the dynamic cast, and all the target types are public.

Only GetRouteAttribute and GetRoutePrefixAttribute check to see if the attribute implements an interface before performing a dynamic cast.

### DIFF
--- a/src/NSwag.SwaggerGeneration.WebApi.Tests/RouteAttributeTests.cs
+++ b/src/NSwag.SwaggerGeneration.WebApi.Tests/RouteAttributeTests.cs
@@ -1,5 +1,7 @@
-﻿using System.Threading.Tasks;
+﻿using System;
+using System.Threading.Tasks;
 using System.Web.Http;
+using System.Web.Http.Routing;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 
 namespace NSwag.SwaggerGeneration.WebApi.Tests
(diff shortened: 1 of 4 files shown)


hmflash:
I couldn't make up my mind on the most elegant way to fix this.  I opted for facade classes vs refactoring `IEnumerable<Attribute> GetRouteAttributes` into `IEnumerable<string> GetRouteTemplates`.  If you thing the latter is better I'm happy to change to that.

RSuter:
I think `.Where(a => ObjectExtensions.HasProperty(a, "Template")` is still needed because in one framework (asp.net or asp.net core) the property Template is completely missing (I'm not sure, but its just to be safe)

hmflash:
Good to know.  I'll add that back in.

RSuter:
Thank you for the fix and the very clean code...

RSuter:
Ok, everything looks good. 
v11 will be released soon... Do you have any open issues?

hmflash:
Looks good to me!

PR #786, opened 2017-06-08
Added HttpOperationAttribute on generated methods

exyi:
Added `HttpOperationAttribute` to NSwag.Annotations that contains HTTP method and path that the method is going to invoke. It is only generated when `GenerateDescriptorAttributes` on settings is `true`.

I need it for metaprogramming - for https://github.com/riganti/dotvvm to integrate with web API. But is also may be useful as "documentation" of the generated client.

### DIFF
--- a/src/NSwag.Annotations/HttpOperationAttribute.cs
+++ b/src/NSwag.Annotations/HttpOperationAttribute.cs
@@ -1,0 +1,16 @@
+using System;
+
+namespace NSwag.Annotations
+{
+    public class HttpOperationAttribute: Attribute
+    {
+        public string Method { get; }
+        public string Path { get; }
+
+        public HttpOperationAttribute(string method = null, string path = null)
+        {
+            Method = method;
+            Path = path;
+        }
+    }
+}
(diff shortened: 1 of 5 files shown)


RSuter:
In general, I'm ok with this enhancement... But we shouldn't put the attribute in NSwag.Attributes, because they are all used with the Swagger generator and not for the generated code (rule for the generated code: No depencency to NSwag or NJsonSchema). 

What do you think of a setting `HttpOperationAttributeType` (default: null, e.g. "MyNamespace.HttpOperationAttribute") which generates the attribute like

    [<#=model.HttpOperationAttribute#>(OperationId = <#=model.OperationId#>, Method = <#=model.Method#>, Path = <#=model.Path#>)]

And you have to implement the attribute somewhere in your project...

What do you think?

exyi:
Ok, this is a good idea, just give me some time to do it, I'm just leaving to holiday ;)

RSuter:
Ok thank you for the PR. Enjoy your holiday.

exyi:
Hi, I think that it should work as you suggested finally. Just sorry for the line ending changes, but I was not able to change it back on my machine.

RSuter:
This can now easily be implemented with a custom template: https://github.com/RSuter/NSwag/wiki/Templates

RSuter:
Is that ok for you?

PR #818, opened 2017-06-25
Values for optional parameters

tomasherceg:
The optional parameters can have default values in Swagger metadata, however NSwag doesn't use them and uses `null` for any optional parameter:

```
public MyMethod(int? parameter = null) 
{
    ...
}
```

This PR adds a new configuration option `GenerateOptionalParameterDefaultValues` (default is `false` so there are no breaking changes).
When this option is switched on, the default value for string and numeric parameters will be generated, e.g.: 

```
public MyMethod(int? parameter = 1) 
{
    ...
}
```

It supports C# and TypeScript and I have also added a few tests.


RSuter:
I think we should use

https://github.com/RSuter/NJsonSchema/blob/15c3d4a128dc7fdfa9c5f73d172982b83d6bc7e8/src/NJsonSchema.CodeGeneration/DefaultValueGenerator.cs

(and CSharpDefaultValueGenerator / TypeScriptDefaultValueGenerator) for the DefaultValueCode. I think we also need to enhance the generators with handling for float/double/etc like you did in the PR

tomasherceg:
Ah cool, I didn't know about this class. 
I'll create a PR to the NJsonSchema library to enhance the support for all numeric types and adjust this PR to use the method.

RSuter:
This would be awesome... NJS is also maintained by me, so ill merge this PR as fast as possible...

tomasherceg:
OK, I'll try to do it later today or tomorrow.

tomasherceg:
I have changed the implementation to use the default value mechanism from NJsonSchema.

RSuter:
Have you seen my question in the NJS issue?

> Just one question: What is the type of schema.Default? Isnt it a JToken instead of a primitive value sometimes?

tomasherceg:
Oh, I haven't noticed the question. 

According to the tests in NSwag, the `schema.Default` contains the real value, at least for the use case I am testing - optional parameters in methods of string, numeric and boolean types. We ignore the others anyway.

Or are you aware of some situation where `schema.Default` can be JToken?

RSuter:
The Problem: 

            //// Arrange
            var schema = new JsonSchema4();
            schema.Default = 100m;

            //// Act
            var json = schema.ToJson();
            var schema2 = await JsonSchema4.FromJsonAsync(json);

            //// Assert
            Assert.AreEqual(schema.Default, schema2.Default);

Error: 

    Microsoft.VisualStudio.TestTools.UnitTesting.AssertFailedException: 
    'Assert.AreEqual failed. Expected:<100 (System.Decimal)>. Actual:<100 (System.Double)>. '

tomasherceg:
Oh, I see. We should use the type of the parameter instead of the actual type of the default value.

RSuter:
See https://github.com/RSuter/NJsonSchema/blob/master/src/NJsonSchema/Generation/JsonObjectTypeDescription.cs

RSuter:
I think the problem is that you do not have a schema for the default value - and the default value is just a JSON token... We could add a custom/extension property with the schema and generate this.

RSuter:
x-defaultSchema?

PR #819, opened 2017-06-25
Fix clients encoding parameter values into URIs using current culture

sandord:
NSwag generates C# clients that encode parameters into request URIs using the current culture. This becomes a problem when running on a machine that has a culture set that - for example - encodes decimal separators as a comma.

Providing a parameter to the client with a value of `0.5` then becomes encoded as `0,5` (`0%2C5`) in the URI and that value cannot be parsed as a `float`, `double` or `decimal` by the WebAPI controller correctly, since WebAPI uses the `CultureInfo.InvariantCulture` to parse arguments.

This pull request fixes the client template by explicitly converting parameter values using the invariant culture and provides an accompanying integration test.


RSuter:
Thank you for the PR.

RSuter:
v11.2.0

PR #826, opened 2017-06-29
Add support for changing default date/time parameter formatting

krisztiankocsis:
**Note that I didn't push the generated template CS file because it replaces all paths in #line statements with my project location!**


RSuter:
Thank you for the PR.

> Note that I didn't push the generated template CS file because it replaces all paths in #line statements with my project location!

This is good, this way I can check all changes introduced by the PR...

RSuter:
Instead of checking invalidChars, we should use a regex to check if its a valid property literal, eg

Regex.Match("[a-zA-Z][a-zA-Z0-9_]*")

RSuter:
I think this one is fine: 

https://stackoverflow.com/a/6671856/876814

`^[a-zA-Z_$][0-9a-zA-Z_$]*$`

RSuter:
Sorry, i posted the comments in the wrong PR :)

PR #827, opened 2017-06-29
Unified and customizable JSON serialization settings

krisztiankocsis:


krisztiankocsis:
C# generated client template is not committed!

src/NSwag.CodeGeneration.CSharp/Templates/ClientTemplate.tt
```diff
@@ -113,10 +113,20 @@ public partial interface I<#=Model.Class#>
     }
 
 <#}#>
+    partial void UpdateJsonSettings(ref Newtonsoft.Json.JsonSerializerSettings settings);
```
RSuter: Don't use ref as it is already a reference and with ref you could completely replace the settings...

RSuter: + Rename to "UpdateJsonSerializerSettings"

src/NSwag.CodeGeneration.CSharp/Templates/ClientTemplate.tt
```diff
@@ -113,10 +113,20 @@ public partial interface I<#=Model.Class#>
...
     partial void PrepareRequest(<#=Model.HttpClientType#> client, System.Net.Http.HttpRequestMessage request, string url);
     partial void PrepareRequest(<#=Model.HttpClientType#> client, System.Net.Http.HttpRequestMessage request, System.Text.StringBuilder urlBuilder);
     partial void ProcessResponse(<#=Model.HttpClientType#> client, System.Net.Http.HttpResponseMessage response);
 
+    Newtonsoft.Json.JsonSerializerSettings NewJsonSettings()
```
RSuter: Rename to CreateJsonSerializerSettings (a method must contain a verb, but new is not a verb)

src/NSwag.CodeGeneration.CSharp/Templates/ClientTemplate.tt
krisztiankocsis: Done

src/NSwag.CodeGeneration.CSharp/Templates/ClientTemplate.tt
krisztiankocsis: Done

RSuter:
Looks good...

src/NSwag.CodeGeneration.CSharp/Models/CSharpClientTemplateModel.cs
```diff
@@ -129,7 +129,14 @@ public string JsonSerializerParameterCode
                 if (RequiresJsonExceptionConverter)
                     jsonConverters = jsonConverters.Concat(new[] { "JsonExceptionConverter" });
 
-                return CSharpJsonSerializerGenerator.GenerateJsonSerializerParameterCode(handleReferences, jsonConverters.ToList());
+                var parameterCode = CSharpJsonSerializerGenerator.GenerateJsonSerializerParameterCode(handleReferences, jsonConverters.ToList());
+
+                if (parameterCode.Length == 0)
+                    parameterCode = "new Newtonsoft.Json.JsonSerializerSettings()";
+                else
+                    parameterCode = parameterCode.Substring(2);
```
RSuter: Why do you transform the code? This way the handleReferences setting is not respected... We should directly return `parameterCode.Substring(2); ` (and I should fix GenerateJsonSerializerParameterCode [remove ", "]).

RSuter:
We also need a GenerateUpdateJsonSerializerSettingsMethod setting (default: true) to implement it in the base class. Ill do that...

src/NSwag.CodeGeneration.CSharp/Models/CSharpClientTemplateModel.cs
krisztiankocsis: 1. There is a way in GenerateJsonSerializerParameterCode() when it returns an empty string, handleReference is not respected only in this case, which I think is also OK.
2. Due to probable empty string, we cannot simply return Substring(2).
3. Changing GenerateJsonSerializerParameterCode() implementation is a good idea (see below), but don't forget your current users who expect the current behavior.

I recommend the following:
* Update GenerateJsonSerializerParameterCode() by overload it:
  * Add two new parameters (just for clarification): bool initializer (when true, omit the ", " prefix)
  * Another collection for converters that cannot be instantiated but can be accessed via static properties (like the ones in NodaTime)
* The current one should call this method with false and empty collection for the new parameters

The later argument can be used in my third PR to support singleton, property accessed converters (e.g. NodaTime).

What do you think?

krisztiankocsis:
Sounds good

src/NSwag.CodeGeneration.CSharp/Templates/ClientTemplate.tt
```diff
@@ -105,7 +105,9 @@ public partial interface I<#=Model.Class#>
 		_settings = new System.Lazy<Newtonsoft.Json.JsonSerializerSettings>(() => 
         {
             var settings = <#=Model.JsonSerializerParameterCode#>;
+<#if(Model.GenerateUpdateJsonSerializerSettingsMethod){#>
```
RSuter: We need this always, either partial (true) or in base class (false)

krisztiankocsis: Ok, I got it, reverting it

PR #874, opened 2017-08-09
Use TransferRequestHandler instead of runAllManagedModulesForAllRequests

### LINKED ISSUE #871: Needlessly complex method is suggested for handling the static files
https://github.com/RSuter/NSwag/wiki/OwinGlobalAsax suggests adding three handlers to deal with the static files, but this method is needlessly complex. I assume you got your info from https://stackoverflow.com/a/41155395 ?

I've posted a new answer to that question, repeating it here for convenience:

By default, the StaticFile handler is the last handler to be executed. This means that simply intercepting all NSwag requests is enough:

```xml
<add name="NSwag" path="swagger" verb="*" type="System.Web.Handlers.TransferRequestHandler" preCondition="integratedMode,runtimeVersionv4.0" />
```

This also [used to be](https://github.com/domaindrivendev/Swashbuckle/issues/57) the recommended method by Swashbuckle, until it switched to extensionless URLs. You may notice that ASP.NET is able to handle extensionless URLs with the same type of handler in `<add name="ExtensionlessUrlHandler-Integrated-4.0"`.

In addition, this should also be documented on https://github.com/RSuter/NSwag/wiki/Middlewares because this problem is not limited to the Global.asax method.

stijnherreman:
This fixes #871.

The wiki pages have also been updated:
* https://github.com/RSuter/NSwag/wiki/OwinGlobalAsax
* https://github.com/RSuter/NSwag/wiki/Middlewares

### DIFF
--- a/src/NSwag.Sample.NetGlobalAsax/Web.config
+++ b/src/NSwag.Sample.NetGlobalAsax/Web.config
@@ -24,9 +24,10 @@
       <remove name="OPTIONSVerbHandler" />
       <remove name="TRACEVerbHandler" />
       <add name="ExtensionlessUrlHandler-Integrated-4.0" path="*." verb="*" type="System.Web.Handlers.TransferRequestHandler" preCondition="integratedMode,runtimeVersionv4.0" />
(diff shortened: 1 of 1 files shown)


RSuter:
Thank you very for this.

PR #994, opened 2017-10-12
WIP Initial work for NSwag.SwaggerGenerator.AspNetCore

pranavkm:
See https://github.com/RSuter/NSwag/issues/999


pranavkm:
cc @RSuter \ @rynowak

RSuter:
Hi pranav, thanks for the PR. 
Can you please create a PR to merge into "api-explorer" (https://github.com/RSuter/NSwag/tree/

pranavkm:
Rebased on api-explorer branch. Switched to SDK-based csproj + xunit.

src/NSwag.SwaggerGeneration.AspNetCore/AspNetCoreToSwaggerGeneratorSettings.cs
```diff
@@ -0,0 +1,63 @@
...
+        public DocumentProcessorCollection DocumentProcessors { get; } = new DocumentProcessorCollection
+        {
+            new DocumentTagsProcessor()
+        };
+
+        /// <summary>Gets or sets the document template representing the initial Swagger specification (JSON data).</summary>
+        public string DocumentTemplate { get; set; }
+
+        /// <summary>Gets or sets a value indicating whether the controllers are hosted by ASP.NET Core.</summary>
+        public bool IsAspNetCore { get; set; }
```
RSuter: Not needed, always true

src/NSwag.SwaggerGeneration.AspNetCore.Tests/AspNetCoreToSwaggerGenerationTests.cs
```diff
@@ -0,0 +1,101 @@
...
+        }
+
+        private class TestModel { }
+
+        private class ProblemDetails { }
+
+        private class ApiResponseType2_1 : ApiResponseType
+        {
+            public bool IsDefaultResponse { get; set; }
+        }
```
rynowak: Pretty clever

src/NSwag.SwaggerGeneration.AspNetCore.Tests/AspNetCoreToSwaggerGenerationTests.cs
```diff
@@ -0,0 +1,101 @@
+﻿using System.Linq;
```
rynowak: Should this have a license header?

src/NSwag.SwaggerGeneration.AspNetCore/AspNetCoreToSwaggerGenerator.cs
```diff
@@ -0,0 +1,257 @@
...
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Mvc.ApiExplorer;
+using Microsoft.AspNetCore.Mvc.Controllers;
+using NJsonSchema;
+using NJsonSchema.Infrastructure;
+using NSwag.SwaggerGeneration.Processors.Contexts;
+
+namespace NSwag.SwaggerGeneration.AspNetCore
+{
+    /// <summary>Generates a <see cref="SwaggerDocument"/> object for the given Web API class type. </summary>
```
rynowak: Should this be updated?

src/NSwag.SwaggerGeneration.AspNetCore/AspNetCoreToSwaggerGenerator.cs
```diff
@@ -0,0 +1,257 @@
...
+        /// <summary>Initializes a new instance of the <see cref="AspNetCoreToSwaggerGenerator" /> class.</summary>
+        /// <param name="settings">The settings.</param>
+        /// <param name="schemaGenerator">The schema generator.</param>
+        public AspNetCoreToSwaggerGenerator(AspNetCoreToSwaggerGeneratorSettings settings, SwaggerJsonSchemaGenerator schemaGenerator)
+        {
+            Settings = settings;
+            _schemaGenerator = schemaGenerator;
+        }
+
+        /// <summary>Gets or sets the generator settings.</summary>
```
rynowak: just gets

src/NSwag.SwaggerGeneration.AspNetCore/AspNetCoreToSwaggerGenerator.cs
```diff
@@ -0,0 +1,257 @@
...
+        public AspNetCoreToSwaggerGeneratorSettings Settings { get; }
+
+        /// <summary>Generates a Swagger specification for the given <see cref="ApiDescriptionGroupCollection"/>.</summary>
+        /// <param name="apiDescriptionGroups"><see cref="ApiDescriptionGroup"/>.</param>
+        /// <returns>The <see cref="SwaggerDocument" />.</returns>
+        /// <exception cref="InvalidOperationException">The operation has more than one body parameter.</exception>
+        public Task<SwaggerDocument> GenerateAsync(ApiDescriptionGroupCollection apiDescriptionGroups)
+        {
+            var apiDescriptions = apiDescriptionGroups.Items.SelectMany(g => g.Items);
+            return GenerateAsync(apiDescriptions);
```
rynowak: I think maybe this is an open question, or a future improvement but we might not want to destroy the group information. 

The intended usage of groups is for tags or versioning. See the grouping in http://petstore.swagger.io/ for example.

I'm a little concerned if we build an API that **can't** expose the groupings.

src/NSwag.SwaggerGeneration.AspNetCore/AspNetCoreToSwaggerGenerator.cs
```diff
@@ -0,0 +1,257 @@
...
+        /// Generates a Swagger specification for the given <see cref="ApiDescription"/> sequence.
+        /// </summary>
+        /// <param name="apiDescriptions"><see cref="ApiDescription"/></param>
+        /// <returns>The <see cref="SwaggerDocument" />.</returns>
+        public async Task<SwaggerDocument> GenerateAsync(IEnumerable<ApiDescription> apiDescriptions)
+        {
+            var document = await CreateDocumentAsync(Settings).ConfigureAwait(false);
+            var schemaResolver = new SwaggerSchemaResolver(document, Settings);
+
+            var apiGroups = apiDescriptions.Where(apiDescription => apiDescription.ActionDescriptor is ControllerActionDescriptor)
```
rynowak: use `OfType`

src/NSwag.SwaggerGeneration.AspNetCore/AspNetCoreToSwaggerGenerator.cs
```diff
@@ -0,0 +1,257 @@
...
+                {
+                    var path = operation.Path.Replace("//", "/");
+
+                    if (!document.Paths.ContainsKey(path))
+                        document.Paths[path] = new SwaggerOperations();
+
+                    if (document.Paths[path].ContainsKey(operation.Method))
+                    {
+                        throw new InvalidOperationException("The method '" + operation.Method + "' on path '" + path + "' is registered multiple times " +
+                            "(check the DefaultUrlTemplate setting [default for Web API: 'api/{controller}/{id}'; for MVC projects: '{controller}/{action}/{id?}']).");
```
rynowak: Is this error message relevant? We don't support ApiExplorer for conventional routes. You can still have duplicates, but the way to fix it is wrong here. Consider a message similar to what we do here https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNetCore.Mvc.Core/Internal/ActionSelector.cs#L140

src/NSwag.SwaggerGeneration.AspNetCore/AspNetCoreToSwaggerGenerator.cs
```diff
@@ -0,0 +1,257 @@
...
+            }
+        }
+
+        private async Task<SwaggerDocument> CreateDocumentAsync(AspNetCoreToSwaggerGeneratorSettings settings)
+        {
+            var document = !string.IsNullOrEmpty(settings.DocumentTemplate) ? await SwaggerDocument.FromJsonAsync(settings.DocumentTemplate).ConfigureAwait(false) : new SwaggerDocument();
+
+            document.Generator = "NSwag v" + SwaggerDocument.ToolchainVersion + " (NJsonSchema v" + JsonSchema4.ToolchainVersion + ")";
+            document.Consumes = new List<string> { "application/json" };
+            document.Produces = new List<string> { "application/json" };
```
rynowak: I think this is fine for a first cut, but we need to take a second look at this.

src/NSwag.SwaggerGeneration.AspNetCore.Tests/AspNetCoreToSwaggerGenerationTests.cs
pranavkm: The other tests didn't - https://github.com/RSuter/NSwag/blob/master/src/NSwag.SwaggerGeneration.WebApi.Tests/ControllerClassesTests.cs. @RSuter ?

src/NSwag.SwaggerGeneration.AspNetCore/AspNetCoreToSwaggerGenerator.cs
pranavkm: Sure. I wasn't sure what that translated in to here. I'll add it as tags for the time being if it's present.

src/NSwag.SwaggerGeneration.AspNetCore/AspNetCoreToSwaggerGenerator.cs
pranavkm: It's a subproperty and we want to get the top level thing. Btw, this was an open question. Should we attempt to generate documents for ApiDescriptions with the available information if we aren't dealing with `ControllerActionDescriptor` instances?

src/NSwag.SwaggerGeneration.AspNetCore/AspNetCoreToSwaggerGenerator.cs
rynowak: I'm not trying to suggest that we just **do** something without an agreement that it's the right thing. I want to make sure that the API we build doesn't prevent it.

src/NSwag.SwaggerGeneration.AspNetCore/AspNetCoreToSwaggerGenerator.cs
rynowak: Ah gotcha.  Ignore me then.

> Should we attempt to generate documents for ApiDescriptions with the available information if we aren't dealing with ControllerActionDescriptor instances?

NSwag requires a method and class - therefore `ControllerActionDescriptor` is required.

src/NSwag.SwaggerGeneration.AspNetCore/AspNetCoreToSwaggerGenerator.cs
```diff
@@ -0,0 +1,257 @@
...
+            }
+
+            var number = 1;
+            while (document.Operations.Any(o => o.Operation.OperationId == operationId + (number > 1 ? "_" + number : string.Empty)))
+                number++;
+
+            return operationId + (number > 1 ? number.ToString() : string.Empty);
+        }
+
+        private string GetActionName(MethodInfo method)
```
rynowak: Should this use the actiondescriptor?

src/NSwag.SwaggerGeneration.AspNetCore/AspNetCoreToSwaggerGenerator.cs
```diff
@@ -0,0 +1,257 @@
...
+        {
+            string operationId;
+
+            dynamic swaggerOperationAttribute = method.GetCustomAttributes().FirstOrDefault(a => a.GetType().Name == "SwaggerOperationAttribute");
+            if (swaggerOperationAttribute != null && !string.IsNullOrEmpty(swaggerOperationAttribute.OperationId))
+                operationId = swaggerOperationAttribute.OperationId;
+            else
+            {
+                if (controllerName.EndsWith("Controller"))
+                    controllerName = controllerName.Substring(0, controllerName.Length - 10);
```
rynowak: Should this come from the action descriptor?

src/NSwag.SwaggerGeneration.AspNetCore/AspNetCoreToSwaggerGenerator.cs
```diff
@@ -0,0 +1,257 @@
...
+        private string GetActionName(MethodInfo method)
+        {
+            var methodName = method.Name;
+            if (methodName.EndsWith("Async"))
+                methodName = methodName.Substring(0, methodName.Length - 5);
+
+            return methodName;
+        }
+
+        private SwaggerOperationMethod GetSwaggerOperationMethod(string httpMethod)
```
rynowak: Should this be `Enum.TryParse`?

src/NSwag.SwaggerGeneration.AspNetCore/AspNetCoreToSwaggerGeneratorSettings.cs
```diff
@@ -0,0 +1,63 @@
...
+
+        /// <summary>Gets or sets a value indicating whether to add path parameters which are missing in the action method.</summary>
+        public bool AddMissingPathParameters { get; set; }
```
rynowak: @pranavkm - can you check if ApiExplorer already does this on your behalf?

src/NSwag.SwaggerGeneration.AspNetCore/OperationParameterProcessor.cs
```diff
@@ -0,0 +1,334 @@
...
+
+            var methodParameters = context.MethodInfo.GetParameters();
+
+            foreach (var apiParameter in parameters.Where(p => p.Source != null))
+            {
+                var parameter = methodParameters.FirstOrDefault(m => m.Name == apiParameter.Name);
+                if (parameter == null || parameter.GetCustomAttributes().Any(a => a.GetType().Name == "SwaggerIgnoreAttribute"))
+                {
+                    continue;
+                }
```
rynowak: Should we use https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNetCore.Mvc.Abstractions/ApiExplorer/ApiParameterDescription.cs#L43

src/NSwag.SwaggerGeneration.AspNetCore/OperationParameterProcessor.cs
```diff
@@ -0,0 +1,334 @@
...
+                                // Try to find a [WillReadBody] attribute on either the action parameter or the bindingAttribute's class
+                                var willReadBodyAttribute = attributes.TryGetIfAssignableTo("WillReadBodyAttribute", TypeNameStyle.Name);
+
+                                if (willReadBodyAttribute == null)
+                                    await AddBodyParameterAsync(context, parameterName, parameter).ConfigureAwait(false);
+                                else
+                                {
+                                    // Try to get a boolean property value from the attribute which explicity tells us whether to read from the body
+                                    // If no such property exists, then default to false since WebAPI's HttpParameterBinding.WillReadBody defaults to false
+                                    var willReadBody = willReadBodyAttribute.TryGetPropertyValue("WillReadBody", true);
```
rynowak: This all looks like Legacy

rynowak: All of the `WillReadBody`

rynowak: Oh nvm this is part of NSwag. Ignore me

src/NSwag.SwaggerGeneration.AspNetCore/OperationParameterProcessor.cs
```diff
@@ -0,0 +1,334 @@
...
+                                }
+                            }
+                            else
+                                await AddPrimitiveParameterAsync(parameterName, context.OperationDescription.Operation, parameter, context.SwaggerGenerator).ConfigureAwait(false);
+                        }
+                    }
+                }
+            }
+
+            if (_settings.AddMissingPathParameters)
```
rynowak: We may not need this, I think ApiDescription does it. Need to check though

src/NSwag.SwaggerGeneration.AspNetCore/OperationParameterProcessor.cs
```diff
@@ -0,0 +1,334 @@
...
+        {
+            if (operationDescription.Operation.ActualParameters.Count(p => p.Kind == SwaggerParameterKind.Body) > 1)
+                throw new InvalidOperationException("The operation '" + operationDescription.Operation.OperationId + "' has more than one body parameter.");
+        }
+
+        private void UpdateConsumedTypes(SwaggerOperationDescription operationDescription)
+        {
+            if (operationDescription.Operation.ActualParameters.Any(p => p.Type == JsonObjectType.File))
+                operationDescription.Operation.Consumes = new List<string> { "multipart/form-data" };
+        }
```
rynowak: @pranavkm - do we do anything like this in MVC? We should think about adding an implicit `[Consumes(...)]` when you have a form file

rynowak: @RSuter  btw, our philosophy of such things is that we try to use ApiExplorer to describe the behavior of the framework in sensible situations.

So understanding that `IFormFile` and `IFormFileCollection` -> file uploads is something we should do. It's also fine if NSwag does it as well, but we want to patch holes like this in our story.

src/NSwag.SwaggerGeneration.AspNetCore/OperationParameterProcessor.cs
RSuter: I think this is legacy: https://github.com/RSuter/NSwag/issues/634

src/NSwag.SwaggerGeneration.AspNetCore/OperationParameterProcessor.cs
```diff
@@ -0,0 +1,334 @@
...
+                            operationParameter.IsNullableRaw = false;
+                            operationParameter.IsRequired = true; // Path is always required => property not needed
+                        }
+                        else if (fromHeaderAttribute != null)
+                            operationParameter.Kind = SwaggerParameterKind.Header;
+                        else
+                            operationParameter.Kind = SwaggerParameterKind.Query;
+
+                        operation.Parameters.Add(operationParameter);
+                    }
```
rynowak: @pranavkm  - this really seems like stuff that ApiDescription/ActionDescriptor should include already

src/NSwag.SwaggerGeneration.AspNetCore/OperationResponseProcessor.cs
```diff
@@ -0,0 +1,222 @@
...
+            var successXmlDescription = await parameter.GetDescriptionAsync(parameter.GetCustomAttributes()).ConfigureAwait(false) ?? string.Empty;
+
+            var responseTypeAttributes = context.MethodInfo.GetCustomAttributes()
+                .Where(a => a.GetType().Name == "ResponseTypeAttribute" ||
+                            a.GetType().Name == "SwaggerResponseAttribute")
+                .Concat(context.MethodInfo.DeclaringType.GetTypeInfo().GetCustomAttributes()
+                    .Where(a => a.GetType().Name == "SwaggerResponseAttribute"))
+                .ToList();
+
+            if (context.ApiDescription.SupportedResponseTypes.Any())
```
rynowak: I believe ApiDescription will also add a default `200` response type no matter what. This all looks like it needs more testing and thought

src/NSwag.SwaggerGeneration.AspNetCore/OperationResponseProcessor.cs
```diff
@@ -0,0 +1,222 @@
...
+                    if (attributeType?.GetRuntimeProperty("IsNullable") != null)
+                        isNullable = responseTypeAttribute.IsNullable;
+
+                    responses.Add(new OperationResponseModel(httpStatusCode, returnType, isNullable, description));
+                }
+
+                foreach (var statusCodeGroup in responses.GroupBy(r => r.HttpStatusCode))
+                {
+                    var httpStatusCode = statusCodeGroup.Key;
+                    var returnType = statusCodeGroup.Select(r => r.ResponseType).FindCommonBaseType();
```
rynowak: `ApiDescription` doesn't really support this - multiple response types with the same status code. We should consider throwing in MVC if someone tries to, but we won't generate a set of actions like this today.

src/NSwag.SwaggerGeneration.AspNetCore/OperationResponseProcessor.cs
```diff
@@ -0,0 +1,222 @@
...
+            else
+                await LoadDefaultSuccessResponseAsync(context.OperationDescription.Operation, context.MethodInfo,
+                    successXmlDescription, context.SchemaGenerator, context.SchemaResolver).ConfigureAwait(false);
+
+            return true;
+        }
+
+        private async Task<ICollection<JsonExpectedSchema>> GenerateExpectedSchemasAsync(OperationProcessorContext context, IGrouping<string, OperationResponseModel> group)
+        {
+            if (group.Count() > 1)
```
rynowak: I don't think this check is necessary, I think a grouping always has at least one member.

src/NSwag.SwaggerGeneration.AspNetCore/OperationResponseProcessor.cs
RSuter: Currently one rule is important: If at least one SwaggerResponeAttribute or SwaggerDefaultResponseAttribute is defined than only these are used for the Swagger/OpenApi spec. More https://github.com/RSuter/NSwag/wiki/WebApiToSwaggerGenerator#additionally-supported-attributes

rynowak: Yeah, I suppose my comment was in reference to the `.Any()` - I'm not sure it's protecting anything.

I thing having the swagger/nswag attributes (when present) override the framework's description is the right thing to do in this case 👍

src/NSwag.SwaggerGeneration.AspNetCore.Tests/AspNetCoreToSwaggerGenerationTests.cs
RSuter: I was too lazy to add it in the tests :)

src/NSwag.SwaggerGeneration.AspNetCore/AspNetCoreToSwaggerGenerator.cs
RSuter: The operation id should be in the form “controller_operation” to ensure correct code generation (multiple client classes per controller and same operation/method names)

src/NSwag.SwaggerGeneration.AspNetCore/OperationParameterProcessor.cs
pranavkm: Yeah, that seems fine. We'll add inference as part of https://github.com/aspnet/Mvc/issues/6847 for 2.1.0 and leave this piece of code around for legacy support

RSuter:
Please also add this fix to your generator: 

https://github.com/RSuter/NSwag/commit/1776b6a12bfd43fb900d69e7d95e24701cd19a84
Unit test: https://github.com/RSuter/NSwag/commit/476f3bd55002b29f1b15009476c8ce45f9d35a17

src/NSwag.SwaggerGeneration.AspNetCore.Tests/AspNetCoreToSwaggerGenerationTests.cs
pranavkm: Cool. I'll go and add it to new tests as I go through them.

src/NSwag.SwaggerGeneration.AspNetCore/OperationParameterProcessor.cs
pranavkm: This was added more recently (as part of 2.0.0). Seems easier to just look it up rather than reflection + fallback.

src/NSwag.SwaggerGeneration.AspNetCore/AspNetCoreToSwaggerGeneratorSettings.cs
pranavkm: Yup - it's here - https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNetCore.Mvc.ApiExplorer/DefaultApiDescriptionProvider.cs#L212-L222.

src/NSwag.SwaggerGeneration.AspNetCore/OperationParameterProcessor.cs
pranavkm: Ok, I'll remove looking up this attribute.

pranavkm:
> > src/NSwag.SwaggerGeneration.AspNetCore/NSwag.snk

> Is this right? I would imagine there's already a key file in the repo

There's a copy of the key file under every project. @RSuter I picked up the key file from the WebApiGenerator project. Was this the right thing to do?

RSuter:
I think I had problems when referencing a single snk file... this is why I copied it... But is it really required? Maybe for "legacy" .NET apps running with ASP.NET Core?

pranavkm:
@RSuter the rules for referencing strong named assemblies are the same between .NET and .NET Core. It's still required for consumers that are strong named.

RSuter:
@pranavkm then I think we should still sign it... You can also move the .snk file to NSwag/src/ so that we can reference it from all projects.

pranavkm:
Updated. @rynowak  could you look at the test setup stuff to make sure I'm doing the right things?

src/NSwag.SwaggerGeneration.AspNetCore/OperationResponseProcessor.cs
```diff
@@ -0,0 +1,116 @@
...
+            var operation = context.OperationDescription.Operation;
+            foreach (var requestFormat in context.ApiDescription.SupportedRequestFormats)
+            {
+                if (!operation.Consumes.Contains(requestFormat.MediaType, StringComparer.OrdinalIgnoreCase))
+                {
+                    operation.Consumes.Add(requestFormat.MediaType);
+                }
+            }
+
+            if (responseTypeAttributes.Count > 0)
```
pranavkm: @RSuter does this pivot seem right to you? If we do find any `SwaggerResponseAttributes` on an action, we'll exclusively use it in lieu of `SupportedResponseType` which is essentially `ConsumesAttribute`. The idea is to encourage you to use one of the two sets of attributes to represent the response format

src/NSwag.SwaggerGeneration.WebApi/OperationResponseProcessor.cs
```diff
@@ -0,0 +1,94 @@
...
+using NJsonSchema;
+using NJsonSchema.Infrastructure;
+using NSwag.SwaggerGeneration.Processors;
+using NSwag.SwaggerGeneration.Processors.Contexts;
+using NSwag.SwaggerGeneration.WebApi.Processors.Models;
+
+namespace NSwag.SwaggerGeneration.WebApi.Processors
+{
+    /// <summary>Generates the operation's response objects based on reflection and the ResponseTypeAttribute, SwaggerResponseAttribute and ProducesResponseTypeAttribute attributes.</summary>
+    public class OperationResponseProcessor : IOperationProcessor
```
pranavkm: Most of the code from this type went in to a `SwaggerResponseBuilder` type which builds the response from attributes. I share the type between the two projects. Let me know if you have any concerns or suggestions on your preferred way of organizing this.

grovesNL:
For what it's worth, there are [packages like StrongNamer](https://github.com/dsplaisted/strongnamer) if you want to avoid the strongly named assembly/snk.

src/NSwag.SwaggerGeneration.WebApi/Processors/SwaggerResponseBuilder.cs
```diff
@@ -0,0 +1,218 @@
...
+                var description = HttpUtilities.IsSuccessStatusCode(httpStatusCode) ? _successXmlDescription : string.Empty;
+                OperationResponseModels.Add(new OperationResponseModel(httpStatusCode, returnType, true, description));
+            }
+        }
+
+
+        public async Task BuildSwaggerResponseAsync(
+            ParameterInfo returnParameter)
+        {
+
```
pranavkm: Will remove extra lines (also before the method declaration)

src/NSwag.SwaggerGeneration.WebApi/Processors/SwaggerResponseBuilder.cs
```diff
@@ -0,0 +1,218 @@
...
+                        .GenerateWithReferenceAndNullability<JsonSchema4>(
+                            returnType, null, typeDescription.IsNullable, _context.SchemaResolver)
+                        .ConfigureAwait(false);
+                }
+
+                _context.OperationDescription.Operation.Responses[httpStatusCode] = response;
+            }
+
+            bool loadDefaultSuccessResponseFromReturnType;
+            if (OperationResponseModels.Count > 0)
```
pranavkm: Slightly tweaked the previous code. This is what the condition looked like in `OperationParameterProcessor`:

```C#
if (responseTypeAttributes.Any() || producesResponseTypeAttributes.Any())
{
    ...
    var operationResponses = context.OperationDescription.Operation.Responses;
    var hasSuccessResponse = operationResponses.ContainsKey("200") || operationResponses.ContainsKey("204");
 
    var loadDefaultSuccessResponseFromReturnType = !hasSuccessResponse &&
        context.MethodInfo.GetCustomAttributes()
            .Any(a => a.GetType().IsAssignableTo("SwaggerDefaultResponseAttribute", TypeNameStyle.Name)) ||
        context.MethodInfo.DeclaringType.GetTypeInfo().GetCustomAttributes()
            .Any(a => a.GetType().IsAssignableTo("SwaggerDefaultResponseAttribute", TypeNameStyle.Name));
 
    if (loadDefaultSuccessResponseFromReturnType)
        await LoadDefaultSuccessResponseAsync(context.OperationDescription.Operation, context.MethodInfo, 
            successXmlDescription, context.SchemaGenerator, context.SchemaResolver).ConfigureAwait(false);
}
else
    await LoadDefaultSuccessResponseAsync(context.OperationDescription.Operation, context.MethodInfo, 
        successXmlDescription, context.SchemaGenerator, context.SchemaResolver).ConfigureAwait(false);
```

src/NSwag.SwaggerGeneration.AspNetCore.Tests/AspNetCoreToSwaggerGenerationTests.cs
```diff
@@ -0,0 +1,446 @@
...
+            public IActionResult CreateModel(int id) => null;
+        }
+
+        private class TestModel { }
+
+        #endregion
+
+        #region ApiDescription generation
+
+        private static ApiDescriptionGroupCollection GetApiDescriptionGroups(Type controllerType)
```
rynowak: This should have some comments to explain what it's for and why it's here.

You can use this approach to cheaply get MVC to do all of the initial reflection for you, but just be aware that this isn't guaranteed to return up-to-date results - it will always be locked to 1.0.0-s behavior.

We also need good unit tests that construct `ApiDescription` directly without this.

src/NSwag.SwaggerGeneration.WebApi/Processors/SwaggerResponseBuilder.cs
```diff
@@ -0,0 +1,218 @@
...
+            {
+                // If there are some attributes declared on the controller \ action, only return a default success response
+                // if a 2xx status code isn't already defined and the SwaggerDefaultResponseAttribute is declared.
+                var operationResponses = _context.OperationDescription.Operation.Responses;
+                var hasSuccessResponse = operationResponses.ContainsKey("200") || operationResponses.ContainsKey("204");
+
```
rynowak: Why is this check limited to `200/204`? instead of any 2XX?

pranavkm: The current WebAPI generator had this and I left it as is - https://github.com/RSuter/NSwag/blob/master/src/NSwag.SwaggerGeneration.WebApi/Processors/OperationResponseProcessor.cs#L127. @RSuter?

src/NSwag.SwaggerGeneration.AspNetCore.Tests/AspNetCoreToSwaggerGenerationTests.cs
pranavkm: > We also need good unit tests that construct ApiDescription directly without this.

Sure

RSuter:
A question for you ASP.NET Core experts: Why does ASP.NET Core generate a HTTP 200 for a void response and not a 204 (like it was in Web API)?

https://github.com/RSuter/NSwag/blob/master/src/NSwag.SwaggerGeneration.WebApi/Processors/OperationResponseProcessor.cs#L218

Sample: 

```
public void MyOperation()
{
}
```

src/NSwag.SwaggerGeneration.WebApi/Processors/SwaggerResponseBuilder.cs
RSuter: I think we should use 
https://github.com/RSuter/NSwag/blob/master/src/NSwag.Core/HttpUtilities.cs#L17
here. 

Please do such changes also in NSwag.SwaggerGeneration.WebApi

rynowak:
> Why does ASP.NET Core generate a HTTP 200 for a void response and not a 204 (like it was in Web API)?

Because this is compatible with what MVC did in the part. ASP.NET Core MVC is a port of MVC - not of WebAPI. We initially had WebAPI's 204 behavior but this broke a lot of users' apps.

src/NSwag.SwaggerGeneration.WebApi/Processors/SwaggerResponseBuilder.cs
pranavkm: This type is shared between the two - so changing it here should fix both.

src/NSwag.SwaggerGeneration.AspNetCore.Tests/AspNetCoreToSwaggerGenerationTests.cs
```diff
@@ -0,0 +1,765 @@
...
+            var document = await generator.GenerateAsync(apiDescriptions);
+
+            //// Assert
+            var operation = Assert.Single(document.Operations, o => o.Path == "/" + nameof(ControllerWithParameters.ComplexFromQueryParameter)).Operation;
+            Assert.Collection(
+                operation.Parameters.OrderBy(p => p.Name),
+                parameter =>
+                {
+                    Assert.Equal(nameof(ComplexModel.Header), parameter.Name);
+                    Assert.Equal(SwaggerParameterKind.Query, parameter.Kind);
```
pranavkm: @rynowak in Mvc 1.0, sub properties of complex models always show up with a BindingSource of `ModelBinder` even if the property has specific binding source annotations. I don't see it in 2.0. Any guesses on what's up?

pranavkm:
Sorry took me a while to get this up date out. @RSuter I have a couple of questions around behavior:

1) For methods that have `SwaggerResponseAttribute` \ `ResponseTypeAttribute`, attempting to combine it with ApiDescription attributes was messy. Consequently, we prefer and only use `SwaggerResponseAttribute` \ `ResponseTypeAttribute` when it's present. https://github.com/RSuter/NSwag/pull/994/files#r145504152

2) I refactored some of the code in OperationResponseModel to allow better sharing between WebApiGenerator and AspNetCoreGenerator. Just wanted you to have a look at it and ensure things still look correct - https://github.com/RSuter/NSwag/pull/994/files#r145505220

3) There isn't always a good way to go from ApiParameterDescription to an actual reflection based object (ParameterInfo \ PropertyInfo) in particular in Mvc.Core 1.0 \ 1.1. I'm doing a best effort to discover the backing property \ parameter, but in certain cases, we might not the right attributes required to infer description or if the parameter is meant to be ignored (`SwaggerIgnoreAttribute`). See my comment about it here - https://github.com/RSuter/NSwag/pull/994/files#diff-eae5c4a75f58eb95412ee3252218e68bR55.

Beyond these, what do you think is the next set of action items for this PR and for https://github.com/RSuter/NSwag/issues/999? I'd like to start investigating `aspnetcore2swagger` command since it has the most moving parts and most unknowns in terms of implementation. What are your thoughts?

RSuter:
@pranavkm either we add experimental middleware static methods (UseSwaggerCore?) or directly implement the command... I think going directly with the command is ok, because this way we could integrate this new code into master without changing existing APIs... Do you think we could setup a Skype call sometime?

src/NSwag.SwaggerGeneration.AspNetCore/NSwag.SwaggerGeneration.AspNetCore.nuspec
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0"?>
```
RSuter: Why do we need a nuspec? Isnt this created on compile time for .net standard libs?

src/NSwag.SwaggerGeneration.AspNetCore/OperationResponseProcessor.cs
RSuter: Looks good. Maybe we also need to add the SwaggerDefaulResponseAttribute to this list...

RSuter:
@pranavkm why did you move OperationResponseProcessor and OperationResponseProcessor into the root of NSwag.SwaggerGeneration.WebApi?

pranavkm:
Thanks @RSuter!

RSuter:
Wow, you are fast :-)

RSuter:
I dont like linked files (wrong namespaces, duplicated classes in assemblies, etc). Do you think it is a bad idea to move the shared classes to NSwag.SwaggerGeneration?

pranavkm:
@RSuter that seems fine. I was trying to avoid breaking changes to the NSwag.SwaggerGeneration.WebApi binary so resorted to the source linking. Do you want to add TypeForwardededTo to the WebApi assembly to avoid breaking compiled callers or do you think there's little value in that?

RSuter:
I think this effort is not worth it - mainly because these libraries will be used mostly only at the application level (i.e. the executable, where the middlewares are registered) via NPM, MSBuild, etc... and there you will always recompile or use libs whose version match. It would be a different story for NSwag.Annotations, or NJsonSchema where the libs are potentially used in a lot more places...

RSuter:
I am thinking about these command line implementation - why cant we use TestServer to get the ApiExplorer/Description in the same process? If I understand you correctly, you want to start the application in another process and get the spec via HTTP? The problem with this approach is: 

1. You will use the NSwag version/generator installed in the application and not the one of the command line
2. What if the web application does not use NSwag at all? It would not be possible to generate a spec, but it should
3. What if the web application has not registered the ApiExplorer and we cannot access it via dependency injection?

Using TestServer would solve all that, we could create a TestServer with the dynamically loaded Startup, register ApiExplorer if necessary, get the ApiDescriptions and feed it into a generator instance of the original process (i.e. the expected version). @rynowak @pranavkm

PR #996, opened 2017-10-13
Allow environment variables in search path, assemblies

CRidge:
I need to reference assemblies located under my user profile, and the .nswag file needs to be checked into source control. To ensure other members of the team find the same assemblies when they run code generation, I need to use environment variables (%USERPROFILE%) in my search paths.

This doesn't seem to be supported, but the fix should just be a matter of adding one line, as I've done in this pull request.


RSuter:
Thank you for the PR. I don't see any problems with this, but I'll wait until the CI build is run before I merge...

PR #1002, opened 2017-10-17
Fix compilation issues of generated Angular client

terkill:
Using Angular 4.4.3 and Typescript 2.5.3 the generated Angular client does compile,

For the code snippet below, Typescript is not able to infer the correct type of `options_` (`observe` is 'response' not string and the same for `responseType`).
It fails with:
```
error TS2345: Argument of type '{ body: string; observe: string; responseType:
string; headers: HttpHeaders; }' is not assignable to parameter of type '{ body?:
any; headers?: HttpHeaders; params?: HttpParams; observe?: HttpObserve;
reportProgress?:...'.
  Types of property 'observe' are incompatible.
      Type 'string' is not assignable to type 'HttpObserve'.
```

```
        let options_ = {
            observe: "response",
            responseType: "blob",
            headers: new HttpHeaders({
                "Content-Type": "application/json",
            })
        };

        return this.http.request("get", url_, options_).flatMap((response_) => {
```

### DIFF
--- a/src/NSwag.CodeGeneration.TypeScript/Templates/AngularClientTemplate.cs
+++ b/src/NSwag.CodeGeneration.TypeScript/Templates/AngularClientTemplate.cs
@@ -693,7 +693,7 @@ if(operation.HasBody){
             
             #line default
             #line hidden
-            this.Write("        let options_ = {\r\n");
+            this.Write("        let options_ : any = {\r\n");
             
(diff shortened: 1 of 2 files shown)


RSuter:
Can't we just cast it to any instead of generating even more code?

terkill:
Sure, `response_` also need to be any though, but still far less changes.

I do not know what to do about the AppVeyor error.

RSuter:
(why didnt they add this as proper interface to the definitions?)

RSuter:
Please test everything with latest ci build

PR #1034, opened 2017-11-02
WIP aspnetcore2swaggercommand 

pranavkm:


RSuter:
Why did you close the PR?

pranavkm:
@RSuter - Here's the correct PR - https://github.com/RSuter/NSwag/pull/1046. This was an early prototype of the change and I'm not entirely sure why I sent it as a PR

RSuter:
Ok, no problem. Is there also a way to generate the spec for existing dlls, or does this only work with a .csproj?

PR #1046, opened 2017-11-06
Add support for aspnetcore2swagger command

pranavkm:


src/NSwag.AspNetCore.Launcher/Program.cs
```diff
@@ -0,0 +1,178 @@
...
+namespace NSwag.AspNetCore.Launcher
+{
+    public class Program
+    {
+        private const string EntryPointType = "NSwag.SwaggerGeneration.AspNetCore.AspNetCoreToSwaggerGeneratorCommandEntryPoint";
+        private static readonly AssemblyName AspNetCoreSwaggerGenerationAssembly = new AssemblyName("NSwag.SwaggerGeneration.AspNetCore");
+        private static readonly Version NSwagVersion = typeof(Program).GetTypeInfo().Assembly.GetName().Version;
+        // List of assemblies and versions referenced by NSwag.SwaggerGeneration.AspNetCore. This represents the minimum versions
+        // required to successfully run the tool.
+        private static readonly Dictionary<string, AssemblyLoadInfo> NSwagReferencedAssemblies = new Dictionary<string, AssemblyLoadInfo>(StringComparer.OrdinalIgnoreCase)
```
pranavkm: cc @rynowak This is the TPA list thing. The tool attempts to load versions referenced in the application first and throws if the application uses a version lower than the minimum requirement.

src/NSwag.Sample.NetCoreAngular/Program.cs
```diff
@@ -7,14 +7,15 @@ public class Program
     {
         public static void Main(string[] args)
         {
-            var host = new WebHostBuilder()
+            BuildWebHost(args).Run();
+        }
+
+        public static IWebHost BuildWebHost(string[] args) =>
```
pranavkm: @RSuter \ @rynowak, the command requires that the app is structured in a way to allow getting to the `IWebHost`. The code path's pretty similar to how TestServer \ dotnet ef is set up and I've filed a work item (https://github.com/aspnet/Hosting/issues/1255) to make our internal package that these two use publicly available. 

Mvc's newly minted test package chooses to proceed with an empty WebHostBuilder (https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNetCore.Mvc.Testing/WebApplicationTestFixture.cs#L123), when the webhost cannot be construed. This ends up being problematic since the service collection does not have nearly enough services to do meaningful work. In the aspnetcore2swagger command, I chose to throw pointing users to the 2.0 docs - https://docs.microsoft.com/en-us/aspnet/core/fundamentals/hosting?tabs=aspnetcore2x to use as a template.

src/NSwag.SwaggerGeneration.AspNetCore/AspNetCoreToSwaggerGeneratorCommandEntryPoint.cs
```diff
@@ -0,0 +1,96 @@
...
+            var assemblyName = new AssemblyName(applicationName);
+            var assembly = Assembly.Load(assemblyName);
+
+            if (assembly.EntryPoint == null)
+            {
+                throw new InvalidOperationException($"Unable to locate the program entry point for {assemblyName}.");
+            }
+
+            var entryPointType = assembly.EntryPoint.DeclaringType;
+            var buildWebHostMethod = entryPointType.GetMethod("BuildWebHost");
```
pranavkm: This would get removed once the HostingStartup.Sources package becomes available.

src/NSwag.Tests/CommandLine/AspNetCoreToSwaggerCommandLineTest.cs
```diff
@@ -0,0 +1,72 @@
+﻿using System;
+using System.Diagnostics;
+using System.IO;
+using System.Threading.Tasks;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace NSwag.Tests.CommandLine
+{
+    [TestClass]
+    public class AspNetCoreToSwaggerCommandLineTest
```
pranavkm: I'll try and get more combination of tests (tfm \ Mvc versions). This uses what was already in your repository.

RSuter:
@pranavkm What is the current state of the PR(s) and the aspnetcore2swagger?

pranavkm:
@RSuter sorry I got pulled off to some other work so I couldn't follow up with you on this. Like we last discussed, we got the command to work without adding an explicit dependency to the project. So running `dotnet aspnetcore2swagger` on a vanilla project referencing Mvc just works. We simulate the same set of steps that occurs as part of TestServer (viz discovering and running parts of your Startup \ Configure pipeline) and use that to do discovery of ApiDescription. 

I'll see if I can work with @rynowak to vet some of the design choices specific to AspNetCore this week so we could try and get this PR once you're happy with it. It might be another week - 10 days, but I'd like to take on up on adding more tests and getting some of the other work items I'd created as follow ups.

pranavkm:
Thanks @RSuter

PR #1055, opened 2017-11-10
Recognize use of Microsoft.AspNetCore.Mvc.FromFormAttribute

Itori:
With AspNetCore, you can use the attribute FromForm.
In the current version, that attribute is not recognize, so query as perform as "Query" type.

I've write test for AspNetCoreToSwaggerGeneration, but wasn't able to test it for real.
I've tested the WebApiToSwaggerGenerator, but didn't find any test testing that kind of thing.

### DIFF
--- a/src/NSwag.SwaggerGeneration.AspNetCore.Tests/AspNetCoreToSwaggerGenerationTests.cs
+++ b/src/NSwag.SwaggerGeneration.AspNetCore.Tests/AspNetCoreToSwaggerGenerationTests.cs
@@ -459,8 +459,6 @@ namespace NSwag.SwaggerGeneration.AspNetCore.Tests
             Assert.True(parameter.IsRequired);
         }
 
-
-
         [Fact]
         public async Task FormFileParametersAreDiscovered()
         {
@@ -724,7 +722,6 @@ namespace NSwag.SwaggerGeneration.AspNetCore.Tests
         {
             public BindingSource BindingSource => BindingSource.ModelBinding;
         }
-
 
         public class ReDocCodeSampleAttribute : SwaggerOperationProcessorAttribute
         {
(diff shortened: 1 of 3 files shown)


RSuter:
Hi @Itori, thank you for the PR - looks good.

RSuter:
So you did test it with the current generator WebApiToSwaggerGenerator but didnt add a unit test?

RSuter:
Is

    operationParameter.IsRequired = true; 

correct? Is a form parameter always required?

RSuter:
As you can see here: 

https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#operation-object-example

It is not always required. I think we should remove this line and leave the required handling to CreatePrimitiveParameterAsync

Itori:
Ya, didn't find a test already existing for FromHeader, and as it need to have a reference to Microsoft.AspnetCore.Mvc, i didn't know how you wanted it.

For form parameter always true, you right, it's not. And the generated code already handle this.

RSuter:
I think its ok to not a have a unit test for this because we will switch to NSwag.SwaggerGeneration.AspNetCore eventually...

For my unit tests I created an own FromFormAttribute so that I can write unit tests without referencing asp.net core packages...

Yes, can you remove this line, and I think we can merge...

RSuter:
Thanks for the PR. Do you need an immediate release?

Itori:
Nah it's ok, i modify my generated code by hand for now, and didn't need to modify it again for a few days

Itori:
Do you need an issue for release changes tracking ?

RSuter:
Ok then, as you may have seen, NSwag is released often so you probably don't need to wait long :-)

> Do you need an issue for release changes tracking ?

Hmm, what do you mean by that?

Itori:
Do you want me to open an issue so that you can have it marked as closed and assign to a release ?

RSuter:
It's ok for me...

PR #1071, opened 2017-11-22
Problems to recollet errors in HttpClientModule Angular 4.3+

SamazoOo:
I think that should solve the problems that I'm having to recollet error with HttpClientModule, the request is returning HttpResponseBase.


RSuter:
Thanks for the PR. My only question is, whether we should really use instanceof? Does it work in all browsers, etc.?

SamazoOo:
I've just tested in IE, Edge, chrome and firefox and there are no problems, it's working. I forgot to include HttpResponseBase in imports, but I did it.

Thanks for the this amazing product

RSuter:
I didn't check it, but why is this helping? Does HttpResponseBase have all needed properties or may they be undefined?

RSuter:
I've updated the templates, please retest with the latest CI artifacts: https://ci.appveyor.com/project/rsuter/nswag-25x6o/build/artifacts

SamazoOo:
It's working for me!

Issue #1004, opened 2017-10-19
Problems with Authorize and AuthenticationSchemes property in ASP.NET Core 2.0


lurumad:
Hi,

I'm working in my ASP.NET Core 2.0 Api and today I've configured the security. When I've configured the controllers with Authorize and set AuthenticationSchemes property with "Bearer", NSwag throws an exception:

System.Reflection.CustomAttributeFormatException: 'AuthenticationSchemes' property specified was not found. ---> System.Reflection.CustomAttributeFormatException: 'AuthenticationSchemes' property specified was not found.

```csharp
    [Route("api/[controller]")]
    [ApiVersion("1.0")]
    [Authorize(AuthenticationSchemes = JwtBearerDefaults.AuthenticationScheme)]
    public class CategoriesController : ControllerBase
```

Any ideas???

Regards!!!

lurumad:
Hi,

I've tried to encapsulate AuthorizeAttribute in my own 

```csharp
    public class JwtAuthorizeAttribute : JwtAuthorizeAttribute 
    {
        public JwtAuthorizeAttribute()
        {
            AuthenticationSchemes = JwtBearerDefaults.AuthenticationScheme;
        }
    }
```

But...

System.MissingMethodException: Method not found: 'Void Microsoft.AspNetCore.Authorization.AuthorizeAttribute.set_AuthenticationSchemes(System.String)'.

Regards!

lurumad:
Works with ActiveAuthenticationSchemes but is obsolete and you have to use AuthenticationSchemes:

![nswag](https://user-images.githubusercontent.com/1729519/31774430-0219a1c8-b4e6-11e7-9d7d-6630124a8b49.png)

Regards!

RSuter:
Do you have a stacktrace?

lurumad:
Hi @RSuter 

Changing the order in referencepaths works:

Initial

```xml
 <Exec Command="$(NSwagExe_Core20) webapi2swagger /assembly:$(OutDir)$(ProjectName).dll /DefaultPropertyNameHandling:CamelCase /output:swagger.json /referencepaths:&quot;%USERPROFILE%\\.nuget\\packages,C:\\Program Files\\dotnet\\store\\x64\\netcoreapp2.0&quot;" />
```
Final

```xml
[...]

PR #1094, opened 2017-12-06
Expand environment variables when using configuration file

thoemmi:
In my .NET Core 2.0 project I'm using EntityFrameworkCore, so I have to specify additional reference paths. This was already solved in https://github.com/RSuter/NSwag/issues/1004#issuecomment-337937640.

However, this doesn't work when using a config file. In my `.nswag` file I have added the paths like this:
```json
"referencePaths": [
  "C:/Program Files/dotnet/store/x64/netcoreapp2.0",
  "%USERPROFILE%/.nuget/packages"
]
```
Unfortunately, the second path isn't expanded but prefixed with the path to the `.nswag` file (because it's not rooted), i.e. it will be "fixed" to `c:\some path\%USERPROFILE%\.nuget\packages`.

This PR ensures that environment variables are expanded first before trying to make the path absolute.

/cc @roeb

### DIFF
--- a/src/NSwag.Commands/NSwagDocument.cs
+++ b/src/NSwag.Commands/NSwagDocument.cs
@@ -170,7 +170,11 @@ namespace NSwag.Commands
         /// <returns>The absolute path.</returns>
         protected override string ConvertToAbsolutePath(string pathToConvert)
         {
-            if (!string.IsNullOrEmpty(pathToConvert) && !System.IO.Path.IsPathRooted(pathToConvert))
+            if (string.IsNullOrEmpty(pathToConvert))
+                return pathToConvert;
+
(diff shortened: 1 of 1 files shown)


RSuter:
Thank you for the PR. Looks good. My main problem is that this only works in one direction - i.e. if you open a nswag document with NSwagStudio it gets converted to absolute paths, then when it's saved it is converted back to relative paths and the environment vars are lost... Would it be possible to also convert it back or somehow maintain the vars?

thoemmi:
D'oh. You're right. Loading the document in the studio should't alter it. I'll look for another solution.

RSuter:
See https://github.com/RSuter/NSwag/issues/1098

Issue #1125, opened 2018-01-04
Empty Value In Http Request Headers If Value is Empty


mvu02:
Using the NSwag generated code, we are adding a header with an empty value to the HTTP request, which occurs because we are setting the value of the optional header parameter to be null. The behavior that we want is when the optional header parameter is null, we should not be adding the header to the HTTP request at all.

An example of such a request is shown here:

POST https://google.com HTTP/1.1
Connection: Keep-Alive
Content-Type: text/plain
Accept: application/json
EmptyHeader:
Host: https://google.com

Our EmptyHeader parameter is null so NSwag is trying to run this:

request_.Headers.TryAddWithoutValidation("EmptyHeader", EmptyHeader!= null ? System.Convert.ToString(EmptyHeader, System.Globalization.CultureInfo.InvariantCulture) : null);

Which as a result ends up adding a header with an empty value. 

Is this the expected behavior for NSwag or should we be adding an if check to the template to make sure that if the value is null we should not try to add it to the header at all.

RSuter:
We should add this if check to the templates... can you create a pr?

mvu02:
@RSuter I created the PR, could you take a look? Thanks.

RSuter:
Closing as PR is merged.

PR #1126, opened 2018-01-04
Fix for Empty Http Request Headers When the Value for the Header Is Null

mvu02:
Fix for Issue: https://github.com/RSuter/NSwag/issues/1125


mvu02:
@RSuter , do we also need to check if the parameter is IsOptional in case of required nullable headers? And in the case that the parameter is not an optional parameter, we should use the original code? Or is the current changes sufficient.

RSuter:
If the parameter is not optional, we should throw an exception the same way like it's done for the other parameter types...

mvu02:
@RSuter , I couldn't find the exact case where we do a check for header parameters that are required for which we would throw an exception, and for cases where the required header parameter may be nullable, would we not want to keep the original behavior of adding a header with an empty value to the request? My changes addressed all types of header parameters instead of just optional ones, and it seems like the original code does not check for required or optional, but loops through all of them and attempts to add them to the request header.

PR #1131, opened 2018-01-05
Separate Logic Between Required and Optional Headers for Null Checking and Throwing Exceptions

mvu02:
Enhancement for Issue: https://github.com/RSuter/NSwag/issues/1125 : 

We should differentiate between required and optional headers instead of all headers in general. In the case that it is a required header, we should check if the value is null, and if so throw an exception, but when the header is optional, we just need to check if the value is null or not. If it is null then we don't need to add the header nor throw an exception.

This also includes a bug fix for PR https://github.com/RSuter/NSwag/pull/1126. The bug is at the end of line 148 where we have an extra "}}" at the end of the if statement with no matching "{{".


mvu02:
Hi @RSuter , as we discussed in https://github.com/RSuter/NSwag/pull/1126, this is the enhancement to throw an exception if the required header value is null.

PR #1147, opened 2018-01-17
Set swagger parameter defaults when generating partial controllers

petertiedemann:
We have just started using NSwag to generate ASP.NET Core controllers (we like schema-first). As part of that work i found that defaults in the swagger schema on parameters are not applied automatically by the controllers.

Consider this example Swagger snippet:
```yaml
paths:
  /lookup:
    post:
      parameters:
      - name: "index"
        in: "query"
        required: false
        default: 42
        type: "integer"
```

When i use the partial controller style, there actually an excellent opportunity to set the default before calling on to the underlying implementation. In addition, the `IXYZController` interface method does not need to accept an `int?` since a default will always be set. 

**Why is this a good thing**? Because it ensures that the backend developer uses the defaults that are expected by the caller.

**Why only partial controllers**? Initially i just wanted to inject the value directly in the method signature, but then i remembered that not all defaults can be set in like that (int can, but decimal or guid cannot for example). It _could_ be done in the base controller class by introducing a level of indirection, but that is a bit more complicated to implemented and i am lazy :)

**Any caveats?** A few, in particular the DefaultToString method that is responsible for rendering the default value in such a way that i can be embedded into the generated code, currently only deals with decimal, double, boolean and string explicitly and not all formats of those. For example datetime and guid format should probably be supported as well. I am happy to add that, but i just want to check that the PR is relevant first. UPDATE: Reading through some of the other PR's i noticed this, which i guess should be used instead?

https://github.com/RSuter/NJsonSchema/blob/master/src/NJsonSchema.CodeGeneration.CSharp/CSharpValueGenerator.cs

Below i have included the output from the unit test i added, showing the behavior as it looks after the changes in this PR:

```c#
    [System.CodeDom.Compiler.GeneratedCode("NSwag", "11.12.16.0 (NJsonSchema v9.10.19.0 (Newtonsoft.Json v9.0.0.0))")]
    public interface IController
    {
        System.Threading.Tasks.Task BarAsync(int abc, bool booldef, int intdef, double doubledef, decimal decdef, string strdef, int? bar = null);
    
    }
    
    [System.CodeDom.Compiler.GeneratedCode("NSwag", "11.12.16.0 (NJsonSchema v9.10.19.0 (Newtonsoft.Json v9.0.0.0))")]
    public partial class Controller : System.Web.Http.ApiController 
    {
        private IController _implementation; 
    
        public Controller(IController implementation)
        {
            _implementation = implementation; 
        }
    
        [System.Web.Http.HttpGet, System.Web.Http.Route("foo/bar")]
        public System.Threading.Tasks.Task Bar(int abc, bool? booldef = null, int? intdef = null, double? doubledef = null, decimal? decdef = null, string strdef = null, int? bar = null)
        {    
            return _implementation.BarAsync(abc, booldef ?? true, intdef ?? 42, doubledef ?? 0.6822871999174D, decdef ?? 79228162514264337593543950335M, strdef ?? @"default""string""", bar);
        }
    }
```
Here abc is required, and bar is not required, but has no default, the rest have various defaults.

Note that the IController interface is always called with the default values, and that the only parameter in the underlying implementation using an explicit nullable type is `bar`.


src/NSwag.CodeGeneration.CSharp.Tests/ControllerGenerationDefaultParameter.cs
```diff
@@ -0,0 +1,113 @@
...
+
+            Assert.IsTrue( code.Contains( "BarAsync(int abc, bool booldef, int intdef, double doubledef, decimal decdef, string strdef, int? bar = null);" ) );
+
+            var trimmedCode = RemoveExternalReferences( code );
+
+            Console.WriteLine( trimmedCode );
+
+            CompilerParameters parameters = new CompilerParameters { GenerateInMemory = true };
+
+            var result = new CSharpCodeProvider().CompileAssemblyFromSource( parameters, trimmedCode );
```
petertiedemann: I didn't see this elsewhere in the controller generation code, but i think its fairly useful to verify that the generated code compiles.

src/NSwag.CodeGeneration/Models/ParameterModelBase.cs
```diff
@@ -32,25 +32,69 @@ public abstract class ParameterModelBase
...
+                // it when using the controller interface indirection
+                TypeInControllerInterface = 
+                    Type.EndsWith( "?" ) ? Type.Substring( 0, Type.Length - 1 ) : Type;
+            }
+            else {
+                TypeInControllerInterface = Type;
+            }
+        }
+
+        private static string DefaultToString( SwaggerParameter parameter ) {
```
petertiedemann: As mentioned above, this one needs to be extended to cover more formats.

src/NSwag.CodeGeneration/Models/ParameterModelBase.cs
```diff
@@ -32,25 +32,69 @@ public abstract class ParameterModelBase
...
+            Type = typeName;
+            Name = parameterName;
+            VariableName = variableName;
+
+            if ( !parameter.IsRequired && parameter.Default != null ) {
+                Default = DefaultToString( parameter );
+
+                // If a default value is supplied, then we can automatically set 
+                // it when using the controller interface indirection
+                TypeInControllerInterface =
```
petertiedemann: Not sure this is the best name, its the type as used in the controller interface. Meaning that it strips the ? when there is a default.

petertiedemann:
Hmm, just noticed this https://github.com/RSuter/NSwag/pull/818 (didn't find it before writing the above code, since i was searching for defaults), but it seems to be a bit dead?

RSuter:
Thanks for the PR. Can't we use the DefaultValueGenerators instead of duplicating this logic?

https://github.com/RSuter/NJsonSchema/blob/master/src/NJsonSchema.CodeGeneration.Tests/DefaultValueGeneratorTests.cs

petertiedemann:
@RSuter Indeed we can ( replacing DefaultToString i would hope ) i think i also mentioned it in the wall-of-text describing the PR. I only found about about it after making the PR, and wanted to verify that the PR was relevant before making more changes. 

Sounds like it is relevant, so i will find some time to use the default value generator instead.

RSuter:
Yes, using the default value generator would be great...

petertiedemann:
Just returned from 3 weeks of vacation, just wanted to say that i still intend to make the changes discussed above.

petertiedemann:
@RSuter i started using value generator instead, and have rebased against master.

src/NSwag.CodeGeneration.CSharp/Models/CSharpParameterModel.cs
```diff
@@ -23,10 +24,17 @@ public class CSharpParameterModel : ParameterModelBase
...
+        public CSharpParameterModel( string parameterName,
+            string variableName,
+            string typeName,
+            SwaggerParameter parameter,
+            IList<SwaggerParameter> allParameters,
+            CodeGeneratorSettingsBase settings,
+            IClientGenerator generator )
+            : base( parameterName, variableName, typeName, 
+                parameter, allParameters, settings, generator, 
+                new CSharpValueGenerator( new CSharpTypeResolver( (CSharpGeneratorSettings) settings ),
```
petertiedemann: I assume this cast is okay?

src/NSwag.CodeGeneration/Models/ParameterModelBase.cs
```diff
@@ -29,28 +29,66 @@ public abstract class ParameterModelBase
...
+            }
+        }
+
+        /// <summary>
+        /// Value generator to use for mapping from <see cref="SwaggerParameter"/> to <see cref="string"/> value in the generated code.
+        /// </summary>
+        protected ValueGeneratorBase ValueGenerator { get; set; }
+
+        private string DefaultToString( SwaggerParameter parameter ) {
+            return ValueGenerator?.GetDefaultValue( parameter, false, parameter.ActualTypeSchema.Id, null, true );
```
petertiedemann: DefaultToString is now just a wrapper around the ValueGenerator.

petertiedemann: @RSuter i am not entirely sure i am feeding the generator the correct arguments here? It passed the test i had against the previous code base, but i would like to understand those parameters a bit better.

RSuter:
Sorry, had no time to check this.. But in general it looks good... hope to merge this in the next days

src/NSwag.CodeGeneration/NSwag.CodeGeneration.csproj
```diff
@@ -19,6 +19,7 @@
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="Newtonsoft.Json" Version="9.0.1" />
+    <PackageReference Include="NJsonSchema.CodeGeneration.CSharp" Version="9.10.37" />
```
RSuter: This is not allowed...

src/NSwag.CodeGeneration.CSharp/Models/CSharpOperationModel.cs
```diff
@@ -49,8 +49,9 @@ public class CSharpOperationModel : OperationModelBase<CSharpParameterModel, CSh
             _resolver = resolver;
 
             var parameters = _operation.ActualParameters.ToList();
-            if (settings.GenerateOptionalParameters)
-                parameters = parameters.OrderBy(p => !p.IsRequired).ToList();
+            if ( settings.GenerateOptionalParameters ) {
+                parameters = parameters.OrderBy( p => !p.IsRequired ).ThenBy( p => p.Default == null ).ToList();
```
RSuter: Is this really necessary? This is potentially a major breaking change for some users...

RSuter:
Hey @petertiedemann, thanks for the PR. You can test/use it in the latest release.

Issue #1157, opened 2018-01-27
Missing NSwag.AspNetCore.Launcher NuGet package for 11.13.0+


saxonww:
It looks like NSwag.AspNetCore.Launcher was not published to NuGet.org. This prevents use of NSwag.ConsoleCore 11.13.0 and 11.13.1.

RSuter:
This is bad. I tried to also release NSwag.AspNetCore.Launcher as NuGet package, but it doesn't work (no NuGet package is created on build).

RSuter:
@pranavkm Can you have a look at this?

pranavkm:
NSwag.ConsoleCore should bundle the contents of AspNetCore.Launcher (which it already does) and not have actual package reference to the the launcher. I'll see if I can send a quick fix for this.

pranavkm:
PR - https://github.com/RSuter/NSwag/pull/1176

cvallance:
@RSuter & @pranavkm - I think this has regressed or manifested in a different way again... see attached screen shot.

Let me know if you want some more info.

<img width="752" alt="screen shot 2018-05-18 at 8 57 26 am" src="https://user-images.githubusercontent.com/525090/40342034-8c169c7e-5ddc-11e8-952d-392820cd851d.png">

RSuter:
Strange, the package contains the NSwag.AspNetCore.Launcher.dll

![image](https://user-images.githubusercontent.com/2603405/40346986-e74fcf40-5d9e-11e8-852e-0d278dea3b2f.png)

pranavkm:
Right now the command assumes you're running it from the published results of `NSwag.ConsoleCore`? It doesn't look like that is what's happening with @cvallance's scenario - looks more along the lines of a `dotnet run` on a project referencing `NSwag.ConsoleCore` package although clearly not from the command line output. @cvallance could you share your setup?

cvallance:
So I'm on OSX. The project is a ASP.NET Core 2 application. The only reference to _NSwag_ is in the `csproj` file and it's a `DotNetCliToolReference` ... i.e. `<DotNetCliToolReference Include="NSwag.ConsoleCore" Version="11.17.9" />`.

I'm invoking it from bash using the command you can see in the screenshot - `dotnet nswag aspnetcore2swagger /project:PaymentSchedulerServer.csproj /output:testing.json`.

[...]

PR #1176, opened 2018-02-06
Remove NSwag.AspNetCore.Launcher dependency from NSwag.ConsoleCore

pranavkm:
Fixes https://github.com/RSuter/NSwag/issues/1157

### DIFF
--- a/src/NSwag.AspNetCore.Launcher/Program.cs
+++ b/src/NSwag.AspNetCore.Launcher/Program.cs
@@ -5,7 +5,7 @@ using System.Reflection;
 
 namespace NSwag.AspNetCore.Launcher
 {
-    public class Program
+    internal class Program
     {
         private const string EntryPointType = "NSwag.SwaggerGeneration.AspNetCore.AspNetCoreToSwaggerGeneratorCommandEntryPoint";
         private static readonly AssemblyName AspNetCoreSwaggerGenerationAssembly = new AssemblyName("NSwag.SwaggerGeneration.AspNetCore");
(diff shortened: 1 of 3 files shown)


src/NSwag.ConsoleCore/NSwag.ConsoleCore.csproj
```diff
@@ -16,12 +16,23 @@
...
   <ItemGroup>
     <PackageReference Include="NConsole" Version="3.9.6519.30868" />
     <PackageReference Include="System.IO.FileSystem" Version="4.3.0" />
     <PackageReference Include="System.Net.Http" Version="4.3.2" />
   </ItemGroup>
-</Project>
\ No newline at end of file
+
+  <!-- Workaround for https://github.com/NuGet/Home/issues/3891 -->
+  <PropertyGroup>
```
pranavkm: Didn't realize this produced a nupkg that was uploaded. I'm adding the outputs of the launcher to this package so you can execute the AspNetCoreToSwaggerCommand. Ideally, we'd want this to be part of the NSwag.Commands project, but this changes none of the build order.

src/NSwag.AspNetCore.Launcher/Program.cs
```diff
@@ -5,7 +5,7 @@
 
 namespace NSwag.AspNetCore.Launcher
 {
-    public class Program
+    internal class Program
```
pranavkm: Since the launcher binary will now be added as a reference when a package reference to NSwag.ConsoleCore, I made this internal to avoid polluting the namespace. It's an internal artifact, so a user project isn't expected to call this.

RSuter:
Thanks for the PR. I didnt know about PrivateAssets="All" but it makes perfectly sense...

PR #1208, opened 2018-02-28
Do not call UpdateJsonSerializerSettings if method does not exists.

coolhome:
When the code generation preference `generateUpdateJsonSerializerSettingsMethod` is set to false we should not make call to `UpdateJsonSerializerSettings`

Unable to compile client because of the following error

```
Severity	Code	Description	Project	File	Line	Suppression State
Error	CS0103	The name 'UpdateJsonSerializerSettings' does not exist in the current context
```


RSuter:
The idea is to set generateUpdateJsonSerializerSettingsMethod = false and provide it in the base class or set it to true and it is generated...

RSuter:
Sorry, this is a little bit confusing... maybe we should put this information in the wiki? Anyway, thanks for the PR!

Issue #1144, opened 2018-01-17
When GenerateSwaggerAsync() catches exception it caches it forever


prochnowc:
We have some custom `IOperationProcessor` which needs to retrieve OIDC configuration. However when this request fails (which might temporarly fail because OIDC config is fetched over http) the failure is cached forever.

In middlewares the method
`protected virtual async Task<string> GenerateSwaggerAsync(HttpContext context)`
(AspNet and AspNetCore)

contains:

```catch (Exception ex)
        {
          swaggerMiddleware._swaggerJson = ex.ToString();
        }
```

The Exception should be cached - but not forever.

RSuter:
Duplicate: https://github.com/RSuter/NSwag/issues/1151
Please discuss there...

PR #1219, opened 2018-03-12
Improved exception handling in middlewares

prochnowc:
- Do not catch exception, leave handling to framework
- Enables use of developer exception page
- Correctly responds with HTTP 500 in case of failure
- Added cache timeout when schema generation threw an exception

Solves #1144.


RSuter:
Thanks for the pr. Looks good.

prochnowc:
Glad that i could help.

Seabizkit:
so how the hell do you stop middle where from caching swagger.json, aka api is deployed.... which goes through cloud-flare, how do i add headers to the swagger ui, to stop this from being cached. version lastest.. V5-r4, does not seem possible.

icnocop:
@Seabizkit you can try to use the `CreateDocumentCacheKey` property on `OpenApiDocumentMiddlewareSettings`.

PR #1236, opened 2018-03-23
Move command entry points to NSwag.Commands

rynowak:
This removes the assembly loader from the reference graph of assemblies
like NSwag.SwaggerGeneration.AspNetCore


rynowak:
reviewed and tested this with @pranavkm

pranavkm:
👍

RSuter:
Thanks @pranavkm and @rynowak 
I will unify the Web API and ASP.NET Core settings and command class to share some logic - so it's better to wait with more PRs for now :-).

Issue #1238, opened 2018-03-25
Typescript Template: Generated client uses `new Headers` that prevents using it on node


dannsam:
I am trying to generate fetch client to work in node, that's all perfect except with the code configuring headers. NSwag creates headers using `Header` object. However, `fetch` also supports dictionaries and arrays - `HeadersInit = Headers | string[][] | { [key: string]: string };`

The change to use `Headers` was because of a bug in Typescript 2.6.1 (https://github.com/RSuter/NSwag/issues/1031) which was fixed in Typescript 2.6.2 https://github.com/Microsoft/TypeScript/issues/19656

I'd like to suggest to change it back to an object literal. That'd allow me to get it working seamlessly in node.

RSuter:
And this bug was only in v2.6.1 or also in a previous version?

dannsam:
`HeadersInit` type was introduced in TS 2.6.0, in 2.5.3 it was `any` https://github.com/Microsoft/TypeScript/blob/94b4f8b79e370020cb31995e8fb0b78f9ba94349/lib/lib.dom.d.ts#L795

The more I think about it the more it feels like a breaking change, since some people may rely on it being a `Headers` object (eg. in that comment https://github.com/RSuter/NSwag/issues/1031#issuecomment-342410823)

Would you prefer to have a new flag for `FetchClient` like (for example `UseObjectLiteralInHeaders`) or have a standalone `NodeClient` ?

RSuter:
Actually it's only broken for 2.6.0 and 2.6.1 and fixed for 2.6.2 and all other versions... So I think we should just revert this and not add a setting for two versions... I'll just add a remark on the wiki that it's broken for these two versions... otherwise we have to maintain this forever :-) (in the UI, command, settings class, templates, etc.) it's not worth it I think.

dannsam:
@RSuter I agree, a flag is not sustainable long term 👍

RSuter:
https://github.com/RSuter/NSwag/wiki/SwaggerToTypeScriptClientGenerator#fetch

TrendyTim:
Is it just me or did this change completely bork fetch (not just the fact i had to change the base class for where i was adding headers, different method than .set to add new headers)

Using NSwagStudio 11.17.15, it appears this completely broke my fetch in Chrome (67), it keeps reporting fetch failed (but im having trouble figuring out why, it only gave me provisional request headers and no response data).

[...]

PR #1239, opened 2018-03-25
Revert to use object literal instead of Headers (TS, fetch)

dannsam:
#1238


RSuter:
https://github.com/RSuter/NSwag/wiki/SwaggerToTypeScriptClientGenerator#fetch

Thanks for the PR!

Issue #239, opened 2016-08-17
Reduce generated file size


michaelaird:
The generated file can be quite large with a lot of repeated code for each operation. I think this could be greatly reduced with a couple of private (generic?) methods that get called with each operation.

rsuter:
Can you post some specific suggestions?

michaelaird:
This kind of block

```javascript
jQuery.ajax({
            url: url,
            beforeSend: this.beforeSend,
            type: "delete",
            data: content,
            dataType: "text",
            headers: {
                "Content-Type": "application/json; charset=UTF-8"
            }
        }).done((data, textStatus, xhr) => {
            this.processDeleteAccommodationReservationWithCallbacks(url, xhr, onSuccess, onFail);
        }).fail((xhr) => {
            this.processDeleteAccommodationReservationWithCallbacks(url, xhr, onSuccess, onFail);
        });
```

gets repeated over and over again. In our existing code we've abstracted that to a method like this
```javascript
public static MakeRequestGeneric<T>(data: Object, url: string, method: 'GET' | 'PUT' | 'POST' | 'DELETE'): JQueryPromise<T>
        {
            var options: JQueryAjaxSettings = <JQueryAjaxSettings>{};
            options.data = ko.toJSON(data);

            options.dataType = "json";
            options.contentType = "application/json";

            options.type = method;

            var request: JQueryPromise<T> = jQuery.ajax(url, options);
            return request;
        }
```

michaelaird:
Parameter validation might also be a good candidate

```javascript
        var url = this.baseUrl + "/api/OCC/{businessUnitId}/Booking/{bookingId}/AccommodationReservation/{reservationId}/{version}?"; 

        if (businessUnitId === undefined || businessUnitId === null)
            throw new Error("The parameter 'businessUnitId' must be defined.");
        url = url.replace("{businessUnitId}", encodeURIComponent("" + businessUnitId)); 
        if (bookingId === undefined || bookingId === null)
            throw new Error("The parameter 'bookingId' must be defined.");
        url = url.replace("{bookingId}", encodeURIComponent("" + bookingId)); 
        if (reservationId === undefined || reservationId === null)
            throw new Error("The parameter 'reservationId' must be defined.");
        url = url.replace("{reservationId}", encodeURIComponent("" + reservationId)); 
        if (version === undefined || version === null)
            throw new Error("The parameter 'version' must be defined.");
[...]

PR #1248, opened 2018-03-30
[Typescript] Reduce generated file size

Guymestef:
Proposition for #239


RSuter:
Thanks for the PR. I didn't have time to look into it yet, as it is a pretty big change... hope I'll get to it soon.

RSuter:
@all it would be great if anyone has time to review this PR!

Arikael:
I just had a quick look at the code.
If I haven't overlooked anything `Client.ProcessResponse.HandleResponse.Return.liquid` won't call the DTOs static method `fromJS` (if applicable).
Instead on [Line 1 of Client.ProcessResponse.HandleResponse.Return.liquid](https://github.com/Guymestef/NSwag/blob/master/src/NSwag.CodeGeneration.TypeScript/Templates/Client.ProcessResponse.HandleResponse.Return.liquid#L1) you have.
```
const result: any = _responseText === "" ? null : <T>{% if HandleReferences %}jsonParse{% else %}JSON.parse{% endif %}(_responseText, this.jsonParseReviver);
```
JSON.parse will not set the correct type which means:
```
const dto = <MyDto>JSON.parse(_responseText);
console.log(dto instanceof MyDTO); //logs false
```

I think it should always return objects with the correct type/prototype
While one could use `Object.assign` (recursively) I would use nswags `fromJS` because using `Object.assign` could theoretically lead to an object with properties which are not defined on `MyDto`.
Maybe, since the code gets generated according to the expected return type, this is acceptable (it would reduce the code size significantly but would only work in IE11+)

The same problem affects #1469

MgSam:
@Guymestef Are you going to fix up this PR? We're also looking for smaller file sizes.

Guymestef:
@MgSam I don't think I will do it, since it was taking too much time to be reviewed and so on, I finally wrote custom templates for my projects based on my PR...

cchatdev:
Any movement on this? @Guymestef how were you able to make your own templates?

RSuter:
See https://github.com/RSuter/NSwag/wiki/Templates

cchatdev:
@RSuter thanks. does this mean make a pull request to add a custom template or can I make and use my own via a command line config?

RSuter:
You can use your own templates via cli... i'm open to improve the existing ones, as long as they don't break existing users (ie. they dont change the public api). This PR here has conflicts, and i think also lots of breaking changes... and I currently don't have much time to invest a lot of time to improve the templates...

cchatdev:
Totally understand. Preventing breaking changes for this particular issue appears to be a large development task and in the end might not even facilitate all cases. Personally i like the command line approach where i can make a completely custom template. You, as a repo, could even host these for other to pull down and use. I'd be happy to share the Aurelia one i'd be making.

Do you have any docs for the param i need to supply for the template? I currently run this via `nswag run` which looks at my *.nswag files in the same directory.

RSuter:
Its the TemplateDirectory setting where you can specify a directory with replacements

snebjorn:
The large size of the generated file is causing significant IntelliSense slowdown in VSCode.

dylanvdmerwe:
Any feedback on this? It's the only downside to using nswag.

Shaddix:
It's not per se reducing the file size, but it's possible to move from Classes to Modules, so that the code becomes treeshakable.
Check out: https://github.com/Shaddix/react-query-swagger/#clients-as-modules

Lib is based on NSwag and is originally meant to generate react-query hooks, but you could skip the hooks via `/no-hooks` option.

dylanvdmerwe:
@Shaddix do you have an example of what this output would look like?

Shaddix:
Kind of.
In case of a single Client 'class' (if multiple they'd be split by files), here's what there is:
1. The file with types:
https://github.com/Shaddix/react-query-swagger/blob/master/examples/pet-client/src/api/no-hooks-client.ts
2. The folder with Clients: https://github.com/Shaddix/react-query-swagger/tree/master/examples/pet-client/src/api/no-hooks-client

PR #1252, opened 2018-04-03
Allow variables to contain backslashes

### LINKED ISSUE #1251: Combining MSBuild and nswag.json variables might create invalid json
I was trying to use Variables and NSwag.json from MSBuild like this:

```
<Exec Command="$(NSwagExe_Core20) webapi2swagger run nswag.json /variables:"assembly=$(TargetPath)" />
```

This will end in Invalid JSON, since $(TargetPath) naturally contains "\" which is needed to be escaped in JSON.

My proposals to solve this are:
- Either provide a way to pass overrides for nswag.json (e.g: `/overrides:"path:in:json=newValue"`) (which is still problematic, because multiple overrides have to be separated by something, which might also be part of newValue)
 - Or allow automatic string escaping in nswag.json (e.g: `nswag.exe run nswag.json /escapeVars:true /variable:"foo=b\ar"`)

I assume the last one would be easier to implement in the variable Substitution, and with some advice where to look, i'd also provide a PR for that.

glatzert:
Fixes #1251

### DIFF
--- a/src/NSwag.Commands/NSwagDocumentBase.cs
+++ b/src/NSwag.Commands/NSwagDocumentBase.cs
@@ -189,7 +189,7 @@ namespace NSwag.Commands
                         var defaultVariables = obj["defaultVariables"].Value<string>();
                         foreach (var p in ConvertVariables(defaultVariables))
                         {
-                            data = data.Replace("$(" + p.Key + ")", p.Value);
+                            data = data.Replace("$(" + p.Key + ")", JsonConvert.ToString(p.Value));
                         }
                     }
                 }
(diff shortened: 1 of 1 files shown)


RSuter:
Currently we do not have any unit tests which cover this...

glatzert:
That somewhat makes them all green as well 😶

RSuter:
Yep. I've added another JsonConvert.ToString() to the default variables...

glatzert:
I somewhat forgot that 👍

RSuter:
Is this a regression of this PR?
https://github.com/RSuter/NSwag/issues/1253#issuecomment-378261857

Issue #1275, opened 2018-04-14
TypeScript Generation: Optional args broken


FynnMazurkiewicz:
The TypeScript Client.RequestUrl.liquid template doesn't generate valid code for optional arguments.
More specifically, the following lines are generated for replacing missing arguments in TypeScript:
```javascript
else
            url_ = url_.replace("/Liquid error: Value cannot be null.
        Parameter name: input}", ""); 
```
Not only is the error message unintentional but the line-break also kills compilation.

I've already identified the error in code and proposed [TOBEREPLACED].

PR #1276, opened 2018-04-14
Fix TS request url generation bug

FynnMazurkiewicz:
Hello everyone! :blush: 

this PR attempts to fix code generation for optional arguments in TypeScript (Issue #1275).

I am neither familiar with the project nor Liquid. 
I just quickly read into liquid and NSwag internals and the triple '{' seem to be a typo.
Either way, they cause Liquid to re-evaluate the return value of reading `parameter.Name`, which is `null`.

Hoping that this will be a helpful contribution, the project seems to be just what I was looking for.
Thanks!

### DIFF
--- a/src/NSwag.CodeGeneration.TypeScript/Templates/Client.RequestUrl.liquid
+++ b/src/NSwag.CodeGeneration.TypeScript/Templates/Client.RequestUrl.liquid
@@ -17,7 +17,7 @@ url_ = url_.replace("{{ "{" }}{{ parameter.Name }}}", encodeURIComponent("" + {{
 {%     endif -%}
 {%     if parameter.IsOptional -%}
 else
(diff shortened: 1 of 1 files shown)


RSuter:
Actually the line

    url_.replace("/{{{ parameter.Name }}}", "");  

should generate 

    url_.replace("/{MyName}", "");  

so your PR is probably wrong, but looking at the other templates, should probably be: 

     url_.replace("/{{ "{" }}{{ parameter.Name }}}", "");

FynnMazurkiewicz:
Yes, this might very well be the case, like I said, I have no domain knowledge. 
This PR really just serves as a 'fix typo' button, in case this turned out to be a typo because I am so low on time.

Either way, my PR fixed the behaviour for my use case, so seems like I got lucky.

PR #1293, opened 2018-04-20
CSharp generator - check for null response content.

### LINKED ISSUE #1291: C# client generation NullReferenceException when using response_.Content
Given this generated code:

    if (status_ == "401") 
    {
        var responseData_ = await response_.Content.ReadAsStringAsync().ConfigureAwait(false); 
        throw new SwaggerException("Bad Request", (int)response_.StatusCode, responseData_, headers_, null);
    }

the `await response_.Content.ReadAsStringAsync()` will throw if `.Content` is null, which it can be if there is no body returned.

Let me know if you agree and I can PR a change to the template.

Cheers!

damianh:
Checks if the response._Content is null before attempting to read the stream. Fixes #1291


RSuter:
I'll release today or tomorrow, ok?

damianh:
Good stuff, cheers.

Issue #945, opened 2017-09-13
Epic: Support OpenApi 3


RSuter:
https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md

- 70%: Support null and oneOf

stijnherreman:
Is it possible to help you with this? I'm particularly interested in the new (or improved?) multipart support described in the spec, it'll allow adding API methods that take `multipart/mixed` content.

RSuter:
Hi @stijnherreman, yes, I'd really appreciate it! 

I think the first step is to find a way to enhance the model (NSwag.Core) so that both versions (2/3) can be serialized/deserialized from a single set of model classes (see updated issue). There are properties which have to be renamed or ignored depending on the spec version - for this we have a custom contract resolver (see https://blog.rsuter.com/advanced-newtonsoft-json-dynamically-rename-or-ignore-properties-without-changing-the-serialized-class/). 

I'm not sure yet, how we can maintain both versions in a single set of model classes (inheritance, only ignore/rename properties, property prefix?). 

I think we should start with the simpler thinks, like the new server specification (these are just new properties on SwaggerDocument). Then we can look into the new response handling, schema changes, etc...

daniel-gwilt-software:
https://github.com/RSuter/NSwag/issues/1091

RSuter:
Most recent commits for enhancing the model for OpenApi 3: 

https://github.com/RSuter/NSwag/commit/3e22a080e439741b4105d8f5c32a6430507abf4e
https://github.com/RSuter/NSwag/commit/1eebd6a978d8a02cf861cffe189d4c7b4954350a
https://github.com/RSuter/NSwag/commit/8dfa231cd5d9b80e20f53f03a16c3c7df56106bd
https://github.com/RSuter/NSwag/commit/aa3d161550232c4b23ba98f39e22ffda3923eb00
https://github.com/RSuter/NSwag/commit/ef1d3d185d832e16206ade0bde068629e72b9efd

RSuter:
Most unit tests for the OpenApi 3 serialization are here: https://github.com/RSuter/NSwag/tree/master/src/NSwag.Core.Tests/Serialization

If someone (@stijnherreman ?) wants to help, we need more unit tests, some OpenApi 3 properties are still missing, etc...

[...]

PR #1308, opened 2018-05-01
Fixes null ref exception when generating 3.0 clients with request bodies

crfloyd:
Fixes issue where request bodies cause null ref exceptions when generating client because ActualName instead of Name should be used when creating the parameter name since this provides a fallback to 'body' in the event that Name is null.


RSuter:
Thanks for the PR!

RSuter:
The generators eventually need to only use the OpenAPI 3.0 properties/models (which are set by the Swagger 2.0 properties) and support additional OpenAPI 3.0 features, see https://github.com/RSuter/NSwag/issues/945

RSuter:
Currently, OpenAPI 3.0 support is not on my priority list, as I want to first consolidate other stuff. But I'm happy to accept PRs - best would be with unit tests to avoid regressions in the future.

PR #1319, opened 2018-05-14
NSwagExe_Core20 broken in 11.7.4

thoemmi:
In 11.7.4 `NSwagExe_Core20` is broken, because it gets overwritten with the value for `NSwagExe_Core21`. I suggest to release a hotfix as soon as possible.


RSuter:
Sorry for that. Will release in the next hours...

PR #1330, opened 2018-05-18
Add support for Angular 6 Singleton Provider (non-test)

nartc:
Just a proposal for adding Singleton Provider option for Angular template. 

Please let me know how I did. I'm more than welcome to receive feedbacks or guidance.


RSuter:
Thanks, looks good, I've just split up the models, so i have to clean up, but everything looks good. I'll share an update here soon...

Issue #1334, opened 2018-05-21
SwaggerToCSharpWebApiControllerGenerator Wrap request responses... doesn't generate compiling code


werner-1:
When generation a C# Web API Controller and setting `"wrapResponses": true` (selecting "Wrap success responses to allow full response access") the generated code for a method is as follows

    var result = await _implementation.TransactionsAsync(reference, transaction).ConfigureAwait(false);
    var status = (System.Net.HttpStatusCode)System.Enum.Parse(typeof(System.Net.HttpStatusCode), result.StatusCode);
    HttpResponseMessage response = Request.CreateResponse(status);

but result.StatusCode is an integer if `"generateResponseClasses": true`

This does not compile.

PR #1335, opened 2018-05-21
Fix C# Web API Controller generator template when "wrapResponses": true

werner-1:
Fix Issue #1334
Don't try to parse an int to an enum in generated Web API Controller.


RSuter:
Thx for the pr!

PR #1340, opened 2018-05-22
Use BaseDirectory to look for DLLs in AspNetCoreToSwaggerCommand

RSuter:


src/NSwag.Commands/Commands/SwaggerGeneration/AspNetCore/AspNetCoreToSwaggerCommand.cs
```diff
@@ -70,11 +70,14 @@ public class AspNetCoreToSwaggerCommand : SwaggerGeneratorCommandBase<AspNetCore
 
                 var cleanupFiles = new List<string>();
 
-                var toolDirectory = Path.GetDirectoryName(typeof(AspNetCoreToSwaggerCommand).GetTypeInfo().Assembly.Location);
                 var args = new List<string>();
                 string executable;
 
 #if NET451
+                var toolDirectory = AppDomain.CurrentDomain.BaseDirectory;
+                if (!Directory.Exists(toolDirectory))
```
pranavkm: In general, this is super unlikely to fail. The BaseDirectory would almost always exist unless you're doing some crazy AppDomain shenanigans.

PR #1343, opened 2018-05-24
Use TestServer to generate spec from ASP.NET Core assemblies

RSuter:
Uses TestServer to start an ASP.NET Core app from assemblies and generate a spec with AspNetCoreToSwaggerGenerator (AspNetCoreToSwaggerCommand.cs)

Loading a .NET Core 1.1 DLL fails because of this: https://github.com/aspnet/Mvc/issues/5690

Is there a workaround?

/cc @pranavkm @rynowak


pranavkm:
> Loading a .NET Core 1.1 DLL fails because of this: aspnet/Mvc#5690

Sadly not. We did not get bug reports to port the fix back to 1.1 and it's unlikely we'd address it at this point. Would targeting a 2.0 app suffice for test scenairos?

robertmuehsig:
FIY: This seems to work https://github.com/aspnet/Home/issues/3132#issuecomment-391988723

RSuter:
Too bad. The problem is that NSwag.SwaggerGeneration.AspNetCore references ASP.NET Core 1.1 ATM (as it is also used in the middleware for 1.1+ and it works there). The only problem is the TestServer feature (used in CLI). Because it references 1.1, it probably won't even work when selecting an ASP.NET Core 2.1 assembly... Maybe we can change this, so that the feature only works for ASP.NET Core 2.0+.

RSuter:
Is this only a problem when using ApplicationParts in the app?

robertmuehsig:
No idea - it fails on startup and when I manually add the ApplicationParts (whatever that is) it works.

RSuter:
@robertmuehsig so, you're already using the new TestServer feature and new generator via CLI with an ASP.NET Core assembly and it works when manually adding the assemblies to ApplicationParts?

robertmuehsig:
Well - I had (more or less) the same problem with my application and saw that your issue was linked, but I'm not using TestServer.

RSuter:
@robertmuehsig Ok, thanks for the info then. Will have to investigate further... have a nice weekend.

PR #1379, opened 2018-06-13
Updates for Version Detection & Addition of Bearer Authentication 

martinholden-skillsoft:
Changed SwaggerDocument.cs to parse the JSON of the Swagger document to get the version versus the contains function that doesn't take into account whitespace. Perhaps a little more overhead but better results.

Added support for OpenAPI Bearer authentication see
https://swagger.io/docs/specification/authentication/bearer-authentication/


RSuter:
I implemented this just yesterday: https://github.com/RSuter/NSwag/commit/f92e130fffba8cdb11ad6a4854e33b5de1fc4149

RSuter:
However your version detection has not changed now... but we need a better solution than deserializing twice - performance is already a problem for big documents and this will make it even worse...

PR #1385, opened 2018-06-15
Improve .NET Core 2.1 sample

RSuter:
Update sample so that 200 and 204 responses are included. 

@pranavkm @rynowak is it correct that the 200/204 responses have to be specified when other status codes are specified with the ProducesResponseTypeAttribute? Is there a reason why you didn't implement the sample controller as in this PR? Does the controller now look as intended?

### DIFF
--- a/src/NSwag.Sample.NETCore21/Controllers/PetController.cs
+++ b/src/NSwag.Sample.NETCore21/Controllers/PetController.cs
@@ -20,6 +20,7 @@ namespace NSwag.Sample
         [HttpPost]
         [Consumes("application/json")]
         [Produces("application/json")]
+        [ProducesResponseType(typeof(void), (int)HttpStatusCode.NoContent)]
         [ProducesResponseType(typeof(SerializableError), (int)HttpStatusCode.BadRequest)]
         public async Task<IActionResult> AddPet([FromBody] Pet pet)
         {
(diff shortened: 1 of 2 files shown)


pranavkm:
>  is it correct that the 200/204 responses have to be specified when other status codes are specified with the ProducesResponseTypeAttribute?

Yup. 200 is the only inferred status code that gets applied when no `Produces` \ `ProducesResponseType` attribute is applied to an action. Once an action is associated with one or more status codes, we expect the user to specify the remainder.

src/NSwag.Sample.NETCore21/Controllers/PetController.cs
```diff
@@ -20,6 +20,7 @@ public class PetController : ControllerBase
         [HttpPost]
         [Consumes("application/json")]
         [Produces("application/json")]
+        [ProducesResponseType(typeof(void), (int)HttpStatusCode.NoContent)]
```
pranavkm: Might be nit-picky, but you could use `Microsoft.AspNetCore.Http.StatusCodes.` here to avoid casting to int.

src/NSwag.Sample.NETCore21/Controllers/PetController.cs
```diff
@@ -87,6 +91,7 @@ public class PetController : ControllerBase
         [HttpPost("{petId}")]
         [Consumes("application/www-form-url-encoded")]
         [Produces("application/json")]
+        [ProducesResponseType(typeof(void), (int)HttpStatusCode.NoContent)]
```
pranavkm: You could do `ProducesResponseType((int)HttpStatusCode.NoContent)` and avoid specifying `typeof(void)`

RSuter:
@pranavkm thanks for the review, will fix soon...

PR #1396, opened 2018-06-19
Fix return type generation for nested Task<ActionResult<T>> types

x-kej:
Currently an ASP.Net return type of Task<ActionResult<T>> will generate a TypeScript class of ActionResultOfT which doesn't match the response which is just a T, since it only removes the outermost generic type. This PR will keep peeling off Task<> and ActionResult<> wrappers until it finds an underlying type.


RSuter:
Good find. Thx for the PR!

PR #1469, opened 2018-07-18
[Review] Reduce file size of the Angular typescript client

yhnavein:
File reduction is achieved by ignoring all the existing processing of the responses. Very simple construction can be used instead, which is just enough for us. 

Example with a diff:
```diff
-    return this.http.request("get", url_, options_).pipe(_observableMergeMap((response_ : any) => {
-        return this.processGetPermissions(response_);
-    })).pipe(_observableCatch((response_: any) => {
-        if (response_ instanceof HttpResponseBase) {
-            try {
-                return this.processGetPermissions(<any>response_);
-            } catch (e) {
-                return <Observable<string[]>><any>_observableThrow(e);
-            }
-        } else
-            return <Observable<string[]>><any>_observableThrow(response_);
-    }));
+    return this.http.get<string[]>(url_, options_);
-}
-
-protected processGetPermissions(response: HttpResponseBase): Observable<string[]> {
-    const status = response.status;
-    const responseBlob = 
-        response instanceof HttpResponse ? response.body : 
-        (<any>response).error instanceof Blob ? (<any>response).error : undefined;
-
-    let _headers: any = {}; if (response.headers) { for (let key of response.headers.keys()) { _headers[key] = response.headers.get(key); }};
-    if (status === 200) {
-        return blobToText(responseBlob).pipe(_observableMergeMap(_responseText => {
-        let result200: any = null;
-        result200 = _responseText === "" ? null : <string[]>JSON.parse(_responseText, this.jsonParseReviver);
-        return _observableOf(result200);
-        }));
-    } else if (status !== 200 && status !== 204) {
-        return blobToText(responseBlob).pipe(_observableMergeMap(_responseText => {
-        return throwException("An unexpected server error occurred.", status, _responseText, _headers);
-        }));
-    }
-    return _observableOf<string[]>(<any>null);
-}
```

So it's 1 line instead of existing 35.

I know that this may be a breaking change for someone, so this change is done in a way that it is possible to enable / disable the new shorter template using a new flag: `UseAngularResultProcessing`. By default (`false`) it's an existing behaviour.

Additional thing that happened in this PR is using the same flag to adjust behaviour of the Result Type generated in the Observable. In our codebase it was been generating always as a nullable Observable, which was not true (and more importantly it was just a PITA)

Please let me know your thoughts.

BTW - this PR is similar to the #1248 but in this, a different approach is used.

### DIFF
--- a/src/NSwag.CodeGeneration.TypeScript/Models/TypeScriptFileTemplateModel.cs
+++ b/src/NSwag.CodeGeneration.TypeScript/Models/TypeScriptFileTemplateModel.cs
@@ -86,6 +86,9 @@ namespace NSwag.CodeGeneration.TypeScript.Models
         /// <summary>Gets a value indicating whether required types should be imported.</summary>
         public bool ImportRequiredTypes => _settings.ImportRequiredTypes;
 
+        /// <summary>Gets or sets a value indicating whether to use a simple Angular's result processing (without all these additional processXXX methods) (default: false)</summary>
+        public bool UseAngularResultProcessing => _settings.UseAngularResultProcessing;
+
         /// <summary>Gets a value indicating whether to call 'transformOptions' on the base class or extension class.</summary>
         public bool UseTransformOptionsMethod => _settings.UseTransformOptionsMethod;
 
--- a/src/NSwag.CodeGeneration.TypeScript/Models/TypeScriptOperationModel.cs
+++ b/src/NSwag.CodeGeneration.TypeScript/Models/TypeScriptOperationModel.cs
@@ -6,6 +6,7 @@
 // <author>Rico Suter, mail@rsuter.com</author>
 //-----------------------------------------------------------------------
 
+using System;
 using System.Linq;
 using NJsonSchema;
 using NJsonSchema.CodeGeneration;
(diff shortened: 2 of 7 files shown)


yhnavein:
Guys, any chance it will be merged?
I'm tired of maintaining my own fork of nswag, releasing it, etc.

RicoSuter:
I think there are too many breaking changes in this - there are lots of changes. I’m quite busy at the moment but hopefully will find time to do a review

ApocDev:
Any update on this one? The changes don't actually seem to be breaking (mostly whitespace changes)

This vastly improves Angular clients, and removes superfluous types in the process.

Jehoel:
@yhnavein 

> I'm tired of maintaining my own fork of nswag, releasing it, etc.

You can use your own templates with NSwag + NSwagStudio by supplying the path to a private `templates` folder in the NSwag project properties - this is is how I work on my own templates (with my own proprietary extensions) without needing to keep my own fork. My own fork now is only for changes I'd like to see adopted upstream.

dylanvdmerwe:
This is still something that would be a great enhancement to the Typescript files!

PR #1462, opened 2018-07-15
Overload for FromJsonAsync with JsonReferenceResolver factory support

Em1ss1oN:
Proposed fix for https://github.com/RSuter/NSwag/issues/1458 with overload for `SwaggerDocument.FromJsonAsync()` with `JsonReferenceResolver` factory parameter. Also referenced NJsonSchema.Yaml in NSwag.Core.Yaml project, and used `JsonAndYamlReferenceResolver`.

PR #1470, opened 2018-07-18
Use yaml and json resolver, update NJS

RSuter:


RSuter:
Merged https://github.com/RSuter/NSwag/pull/1462 instead

PR #1482, opened 2018-07-26
remove extra parentheses

Y2zz:


RSuter:
Are you sure this doesnt break things?

Y2zz:
I found it when I used the Fetch template and I tested the build results and it seems to solve my problem.

Y2zz:
image

RSuter:
I've updated the template in this PR so that other output types don't break. Can you retest with that?

Y2zz:
Still not working, but I found the problem.
Because prior to use Angular template, when modified to Fetch templates, profiles, attributes appear in rxjsversion

image

Y2zz:
Forgot to say, I am using NSwagStudio for editing.

RSuter:
I've added an additional IsAngular check, so it must now use the correct path...

Y2zz:
I just tried it, still invalid

RSuter:
You tried with 

```
{% if Framework.IsFetchOrAurelia or operation.RequestAngularBlobs or Framework.Angular.UseHttpClient or operation.RequestAngularJSBlobs -%}
{%     if Framework.IsAngular and Framework.UseRxJs6 and (Framework.Angular.UseHttpClient or operation.RequestAngularBlobs) -%}
}));
{%     else -%}
});
{%     endif -%}
{% endif -%}
```
?

Y2zz:
image

Yes...

RSuter:
How did you try this? Do you override the template via TemplateDirectory? I'm pretty sure the changed template is not picked up correctly...

Y2zz:
Sorry, it’s my problem, my configuration file is overwritten, it works fine after specifying the template folder.

PR #1521, opened 2018-08-08
  Use ApiParameterDescription.IsRequired from API Explorer with fallback

### LINKED ISSUE #1519: Nullable properties are required in swagger.json
I'm using `UseSwaggerUi3WithApiExplorer` and `Page` and `PageSize` inputs are set as required in swagge.json.

Shouldn't they be optional since they are nullable?
How Can I make them optional? 

I don't want to bring them out of GetListCommand

    [HttpGet]
    public ActionResult GetList([FromQuery]GetListCommand getListCommand)
    {
        return Ok(await _mediator.Send(getListCommand));
    }

    public class GetListCommand
    {
        public int? Page { get; set; }
        public int? PageSize { get; set; }
    }

RSuter:
Closes #1519

### DIFF
--- a/src/NSwag.SwaggerGeneration.AspNetCore/Processors/OperationParameterProcessor.cs
+++ b/src/NSwag.SwaggerGeneration.AspNetCore/Processors/OperationParameterProcessor.cs
@@ -300,11 +300,8 @@ namespace NSwag.SwaggerGeneration.AspNetCore.Processors
             {
                 operationParameter.Default = extendedApiParameter.ParameterInfo.DefaultValue;
             }
-            else
-            {
-                operationParameter.IsRequired = true;
-            }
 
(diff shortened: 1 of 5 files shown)


src/NSwag.SwaggerGeneration.AspNetCore/Processors/OperationParameterProcessor.cs
```diff
@@ -329,6 +326,40 @@ private class ExtendedApiParameterDescription
 
             public IEnumerable<Attribute> Attributes { get; set; }
 
+            public bool IsRequired
+            {
+                get
+                {
+                    // available in asp.net core >= 2.2
+                    if (ApiParameter.HasProperty("IsRequired"))
```
pranavkm: Neat.

Issue #1159, opened 2018-01-29
Issue with generating SDK from .NET assembly.(Dot Net Core)


Boriszn:
Hi Rico, 

Could you please help.

When I try to generate SDK from .net core assembly project it rises an exception: 

```javascript
tem.IO.FileNotFoundException: Could not load file or assembly 'Microsoft.AspNetCore.JsonPatch, Version=1.1.1.0, Culture=neutral, PublicKeyToken=adb9793829ddae60' or one of its dependencies. The system cannot find the file specified.
File name: 'Microsoft.AspNetCore.JsonPatch, Version=1.1.1.0, Culture=neutral, PublicKeyToken=adb9793829ddae60'

Check if .NET Core is installed and 'dotnet' is globally available.
```
But the assembly exists.

RSuter:
Try adding

%USERPROFILE%/.nuget/packages

to the ReferencePaths

RSuter:
v11.14.0: 

We improved the assembly loader (mainly for .NET Core) so that it runs in a more isolated space and the loaded DLLs should better match the requested versions. Please test this with your projects to ensure that we didnt introduce regressions. 

Important if you have DLL loading problems: 

- Check that the DLL in question is available in the output directory (where the selected assembly is located) or in ReferencePaths
- Check that you selected the correct runtime: Either in NSwagStudio UI, as /runtime parameter in the NPM CLI or execute the correct exe binary

For more information regarding assembly loading: https://github.com/RSuter/NSwag/wiki/Assembly-loading

Main commit: https://github.com/RSuter/NSwag/commit/04576e4d1b714ac5bd9b5df11d72469a23f7c0ff#diff-14dafe6661bab407ae5c0d7095ccd1f4

slang25:
We found referencing the .nuget folder problematic as you cannot guarantee you'll pick up the correct version of the assembly.

Try:
```xml
<Copy SourceFiles="@(Reference)" DestinationFolder="$(OutDir)WebApi2SwaggerReferencePath" />
```

And point to `WebApi2SwaggerReferencePath` in `ReferencePaths`.

The later cleanup with:
```xml
<RemoveDir Directories="$(OutDir)WebApi2SwaggerReferencePath" />
```

RSuter:
Correct, in versions prior to v11.14 it was often the problem that the wrong DLL was loaded. In the latest version (v11.14) I try to find the best version based on the requested version: 

[...]

PR #1531, opened 2018-08-10
Changed instances of @(Reference) to @(ReferencePath)

cvallance:
Updated the samples to use `@(ReferencePath)` as per the comment/conversation here:
https://github.com/RSuter/NSwag/issues/1159#issuecomment-407633719

Also updated documentation here:
https://github.com/RSuter/NSwag/wiki/MSBuild


RSuter:
Thanks for the pr 👍

PR #1534, opened 2018-08-11
Clears schemes if no host should be used. Fixes #1517.

### LINKED ISSUE #1517: Removing hostname doesn’t remove service scheme in TypeScript client generation
I’m using NSwagStudio 11.18.2.0. If I generate a TypeScript client for http://petstore.swagger.io/v2/swagger.json with the default settings the generated client starts like this:
	
	export class Client {
	    baseUrl: string;
	    beforeSend: any = undefined;
	    protected jsonParseReviver: ((key: string, value: any) => any) | undefined = undefined;
	
	    constructor(baseUrl?: string) {
	        this.baseUrl = baseUrl ? baseUrl : "https://petstore.swagger.io/v2";
	    }

If I change the service host name in the setting to ‘.’ the generated code will then instead be:

	export class Client {
	    baseUrl: string;
	    beforeSend: any = undefined;
	    protected jsonParseReviver: ((key: string, value: any) => any) | undefined = undefined;
	
	    constructor(baseUrl?: string) {
	        this.baseUrl = baseUrl ? baseUrl : "https:///v2";
	    }

**https:///v2** is not a valid URL :-/ Shouldn’t **https://** be removed as well so the default URL will only be **/v2**?

(Btw, this is my third issue in a really short time. I’m sorry about that, I just really like this tool :-)

pekspro:


RSuter:
Thanks for the pr 👍

PR #1537, opened 2018-08-14
Fixes #1536

### LINKED ISSUE #1536: Problems sending an array of objects in TypeScript
The URL that is created when an array of objects is going to be send with TypeScript is invalid. There is a loop that adds each property to the query string. But it also adds each function to the query string which is bad.

In this loop there is also a cast is also missing when running TypeScript in strict mode.

I will send a PR for this in a few minutes :-)

pekspro:
Regarding the casting, it could also be solved by adding:

    [key: string]: any;

on each generated DTO.


RSuter:
Thanks for the PR! Is this with the cast for strict mode?

pekspro:
The fix makes sure that functions aren’t serialized to the query string (with “hasOwnProperty”). It also fixes a missing cast for strict mode (with (\<any>item)[attr]).

That said, there is still more to do to make the “pet store” compile is strict mode.

Issue #1509, opened 2018-08-05
Generated TypeScript files doesn’t compile with compiler setting “strict:” true


pekspro:
I tried to generate TypeScript files from the petstore swagger file (http://petstore.swagger.io/v2/swagger.json) with NSwagStudio 11.18.1.0. The generated files don’t compile with “strict”: true defined in the TypeScript configuration file:

	...
	  "compilerOptions": {
	    "strict": true,
	    "sourceMap": true,
	    "target": "es2015"
	  },
	...

It complains on functions that returns null where null can’t be returned and some other stuff.

Would be nice if that was supported :-) I’m using TypeScript 2.8.

RSuter:
Did you change the typescript version to the latest available version?

pekspro:
Yes, I selected version 2.7 (nswagstudio).

(Thanks for a super fast reply by the way :-)

RSuter:
Can you post examples where the output is not strict compliant? (it seems TypeScript got stricter :-))

pekspro:
I think a have solved the problem with functions returning null in JQueryCallbacks in #1542.

I haven’t been able to replicate the other problems. All others where related to Promises so I guess it was something in my system that caused these problems. Sorry about that :-)

RSuter:
Can we close this one?

RSuter:
BTW: You can test your custom templates by overriding them: https://github.com/RSuter/NSwag/wiki/Templates

pekspro:
I’m guilty of this, I’m closing it myself :-)

PR #1542, opened 2018-08-17
Fixes "return null" problem in issue #1509.

pekspro:
Changed so that functions that could return a value also could return null.


RSuter:
I think you should change `{% if operation.HasResultType %}` to `{% if operation.HasResultType and SupportsStrictNullChecks %}` so that this is only rendered for supported TS versions...

RSuter:
(I know that probably no one still uses that old versions :-)

RSuter:
What do you think? 
Anyway thanks for the PR!

pekspro:
Ah, you’re right :-) I didn’t know such check existed :-) I will change it later today or tomorrow.

RSuter:
Ok, I've made the change. Thanks for the PR!

PR #1561, opened 2018-08-28
Add local nuget repository as refPaths

yhnavein:
When working with Assembly Loader for dotnet core we found out that we have a lot of issues with missing libraries in the build directory. Fix that is working for us was to add a local nuget packages folder to the `referencePaths` array in the `nswag.json` config file.

However this is problematic as:
1) The path that is given there do not expand env variables
2) The path is different per operating system.

To fix that we used an approach to add a new configuration parameter in the `SwaggerGenerator -> WebApiToSwagger` section named a `loadDefaultNugetCaches` which is false by default (it means by default it does not change the current NSwag behaviour)

When you set a `loadDefaultNugetCaches` to `true` then depending on your OS, a proper path to the local Nuget packages is added to the `referencePaths` section in the background.

And this resolves our issue with the NSwag assembly loader for dotnet core apps.

### DIFF
--- a/samples/WithoutMiddleware/Sample.AspNetCore20/nswag_assembly.nswag
+++ b/samples/WithoutMiddleware/Sample.AspNetCore20/nswag_assembly.nswag
@@ -22,7 +22,8 @@
       "assemblyPaths": [
         "bin/$(configuration)/netcoreapp2.0/Sample.AspNetCore20.dll"
       ],
-      "referencePaths": []
+      "referencePaths": [],
+      "useNuGetCache": false
     }
   },
   "codeGenerators": {}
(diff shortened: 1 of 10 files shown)


RSuter:
Thanks for the PR. 

However I do not recommend adding the nuget cache directory as ReferencePath (slow and problems with different versions and sometimes the package is still missing), have you tried to copy all dlls to a directory and add that to ReferencePath: 

    <Copy SourceFiles="@(ReferencePath)" DestinationFolder="$(OutDir)References" />

See https://github.com/RSuter/NSwag/wiki/MSBuild

However, if there is a need for this PR, ill merge. I think the name UseNuGetCache instead of LoadDefaultNugetCaches would be better as it is not actually loading the cache - what do you think?

yhnavein:
Hi,

I don't like the solution of copying files in the build step. This is complicated when this has to be built the same on Windows and Linux and Mac.
Additionally when I did initial test some time ago even copying files were not sufficient in our case, so I would prefer use this functionality in the meantime we find a proper solution.

I agree with the name `UseNuGetCache`. It is indeed better than mine. I will change it later.

Issue #1591, opened 2018-09-10
Invalid C# property names generated for deep object properties


sweetlandj:
When a deep object property name contains an invalid C# character (e.g. ".") the resulting C# client code also contains the invalid character.  For example, given the following schema:

```
{
   'openapi' : '3.0',
   'info' : {
      'version' : '1.0.2',
       'title' : 'Test API'
   },
   'paths': {
      '/journeys': {
         'get': {
            'tags': [
               'CheckIn'
            ],
            'summary': 'Retrieve journeys',
            'operationId': 'retrieveJourneys',
            'description': '',
            'parameters': [
                {
                   'name': 'lastName',
                   'in': 'query',
                   'schema': { 'type': 'string' }
                },
                {
                   'name': 'eTicketNumber',
                   'in': 'query',
                   'schema': { 'type': 'string' }
                },
                {
                   'name': 'options',
                   'in': 'query',
                   'style': 'deepObject',
                   'explode': true,
                   'schema': { '$ref': '#/components/schemas/Options' }
                }
            ],
            'responses': {}
         }
      }
   },
   'components':{
      'schemas': {
         'Options': {
            'type':'object',
            'properties': {
               'optionalOrder.id': {
                  'schema': { 'type': 'string' }
               },
               'optionalFrequentFlyerCard.id':{
                  'schema': { 'type': 'string' }
               },
               'optionalDepartureDate':{
                  'schema': { 'type': 'string' }
               },
               'optionalOriginLocationCode':{
                  'schema': { 'type': 'string' }
               }
            }
         }
      }
   }
}
```

The generated C# code contains the statement:

```
if (options.OptionalOrder.Id != null)
{
    urlBuilder_.Append("Options[OptionalOrder.Id]=").Append(System.Uri.EscapeDataString(ConvertToString(options.OptionalOrder.Id, System.Globalization.CultureInfo.InvariantCulture))).Append("&");
}
```

Which does not compile.  Also worth noting is that in the generated Options schema, the invalid property names are translated into valid C# identifiers, `OptionalOrderId` and `OptionalFrequentFlyerCardId`, via the configured `IPropertyNameGenerator` .  Therefore, the resulting code _should_ be:

```
if (options.OptionalOrderId != null)
{
    urlBuilder_.Append("options[optionalOrder.Id]=").Append(System.Uri.EscapeDataString(ConvertToString(options.OptionalOrderId, System.Globalization.CultureInfo.InvariantCulture))).Append("&");
}
```

[...]

PR #1592, opened 2018-09-10
Use IPropertyNameGenerator to determine property names of complex que…

sweetlandj:
This is a fix for issue #1591.

It extends `ParameterModelBase` by changing the return type of `PropertyNames` and `CollectionPropertyNames` from `IEnumerable<string>` to `IEnumerable<PropertyModel>`, a new class that exposes both the original `JsonProperty` key (the schema property name) and a generated property name suitable for use as a C# or TypeScript identifier.  `PropertyModel` also contains an overridden `ToString` method that will return the generated property name, which _should_ preserve backward compatibility with existing usages of `{{property}}` in templates.

The property name is now generated via the `IPropertyNameGenerator` provided as part of the `CodeGeneratorSettingsBase` supplied to the `ParameterModelBase` constructor.  This should make them consistent with the component/schema classes in cases where schema references are used.

A new test was added to the `ParameterTests` class in NSwag.CodeGeneration.CSharp.Tests to verify expected behavior.


src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.QueryParameter.liquid
```diff
@@ -6,16 +6,16 @@ urlBuilder_.Append("{{ parameter.Name }}=").Append(System.Uri.EscapeDataString({
 foreach (var item_ in {{ parameter.VariableName }}) { urlBuilder_.Append("{{ parameter.Name }}=").Append(System.Uri.EscapeDataString(ConvertToString(item_, System.Globalization.CultureInfo.InvariantCulture))).Append("&"); }
 {% elseif parameter.IsDeepObject -%}
 {%     for property in parameter.PropertyNames -%}
-if ({{parameter.Name}}.{{property}} != null)
+if ({{parameter.Name}}.{{property.PropertyName}} != null)
```
RSuter: I think we should rename PropertyName to Name (same as in the parameter object on the left)

src/NSwag.CodeGeneration/Models/ParameterModelBase.cs
```diff
@@ -44,6 +45,12 @@ public abstract class ParameterModelBase
             Type = typeName;
             Name = parameterName;
             VariableName = variableName;
+
+            var propertyNameGenerator = settings?.PropertyNameGenerator ?? new DefaultPropertyNameGenerator();
```
RSuter: Is this (and DefaultPropertyNameGenerator.cs) really needed? Isnt a generator already registered/set by default?

RSuter: Yes, I think you can remove DefaultPropertyNameGenerator as there is already one for CSharp and TypeScript which is always set: 

https://github.com/RSuter/NJsonSchema/blob/master/src/NJsonSchema.CodeGeneration.CSharp/CSharpGeneratorSettings.cs#L41

Just throw an exception here, e.g.: 

    settings?.PropertyNameGenerator ?? throw new InvalidOperationException("PropertyNameGenerator not set.");

RSuter:
Thanks for the PR. Looks really good! Is it ready to be merged?

sweetlandj:
Thanks!  Yeah it should be ready to go.  I created packages locally and they seem to behave as expected.

Issue #1599, opened 2018-09-12
API Explorer: Fix required/optional parameters


RSuter:
- Query parameters
- Header parameters
- Path parameters: 
    - Current: Always required
- Body parameters: 
    - Current: Always required, see https://github.com/RSuter/NSwag/issues/1600

Code to fix (currently API Explorer's IsRequired behavior is ignored): [OperationParameterProcessor.cs](https://github.com/RSuter/NSwag/blob/master/src/NSwag.SwaggerGeneration.AspNetCore/Processors/OperationParameterProcessor.cs#L326)
Parameter unit tests: [NSwag.SwaggerGeneration.AspNetCore.Tests/Parameters](https://github.com/RSuter/NSwag/tree/master/src/NSwag.SwaggerGeneration.AspNetCore.Tests/Parameters)

PR: https://github.com/RSuter/NSwag/pull/1608

rynowak:
Rico is there anything you need from us related to this? As I understand it, API Explorer reflects how MVC will behave. If you think there are bugs, lets us know and we can investigate for 2.2.0

RSuter:
@rynowak I have to add more systematic unit tests for this. I wouldn't say there are bugs per se, just unexpected behavior... and maybe lots of breaking changes if I let API Explorer completely handle IsRequired. Will update you as soon as i know more...

RSuter:
@rynowak From #1556: All parameters are by default optional in `public Task AffectCamera(string id, string cameraId)`, correct? Because I think typical parameters are required (otherwise the generator has lots of nullable parameters => ugly) I have to add [BindRequired] everywhere? Is that correct?

rynowak:
When using `[ApiController]` query parameters are optional by default - path parameters are required unless they are optional in the route definition. This reflects how the framework behaves.

If you want a different convention (query parameters are required by default, you can implement that with a convention, and ApiExplorer will respect that).

RSuter:
Ah thanks for this info, then ill switch to the correct implementation and provide an optional convention for backward campatibility...

PR #1606, opened 2018-09-13
Optional/required parameters: Use RequireParametersWithoutDefault fallback setting and API Explorer

RSuter:
Closes https://github.com/RSuter/NSwag/issues/1599


RSuter:
I've added tests for complex body parameters... check my controller in the PR. The tests now reflect the current implementation but they are not how you'd expect them, right?

scottdewald:
> I've added tests for complex body parameters... check my controller in the PR. The tests now reflect the current implementation but they are not how you'd expect them, right?

Looks great. Thanks a lot!

RSuter:
Yeah, but the problem is that this is not how api explorer reports isrequired (currently required = !has default). And the goal is to completely use api explorer...

RSuter:
IMPORTANT: This is not about whether FromBody is required or not but about if a body parameter is required or not...

RSuter:
If I enable api explorer required handling, body parameters are always required. @rynowak is there way to mark body parameters as optional, e.g. [BindOptional]?

rynowak:
With `[ApiController]` body are parameters are required and that's enforced by the framework. We didn't add a way to make body parameters optional because no one has asked for it.

RSuter:
@scottdewald please check PR again, it now adds body parameter tests (always required in the tests) plus a new setting for the "only parameters with default are optional" fallback... is this what you'd expect?

PR #1612, opened 2018-09-17
C# Client: Enhance exception message with status and response

### LINKED ISSUE #1605: HTTP response missing in SwaggerException.Message
When a unit test throws an exception, [xUnit](https://xunit.github.io/) outputs Exception.Message. But SwaggerException.Message does not contain the HTTP status code or the HTTP response, which are crucial to describe the exception.

```
Xavo.DemandClient.SwaggerException : MyMessage
   at SwaggerTests.SwaggerTests.SwaggerTest() in C:\...\SwaggerTests.cs:line 12
```

The [docs](https://docs.microsoft.com/en-us/dotnet/api/system.exception.message?view=netcore-2.1) say this on `Exception.Message`:

> Gets a message that describes the current exception.

I think SwaggerException.Message should return both the HTTP response and the HTTP status code. Using `SwaggerException.ToString()` returns the HTTP response, but doubt that I could convince xUnit developers to switch to using `Exception.ToString` instead of `Exception.Message`.

For reference, here is the unit test which generates the output above:

```csharp
using System.Collections.Generic;
using Xavo.MyControllerClient;
using Xunit;

namespace SwaggerTests
{
    public class SwaggerTests
    {
        [Fact]
        public void SwaggerTest()
        {
            throw new SwaggerException("MyMessage", 401, "HTTP not found response", new Dictionary<string, IEnumerable<string>>(), null);
        }
    }
}
```

cbuchacher:
When a unit test throws an exception, [xUnit](https://xunit.github.io/) outputs
Exception.Message. But SwaggerException.Message does not contain the HTTP
status code or the HTTP response, which are crucial to describe the exception.

    Xavo.DemandClient.SwaggerException : MyMessage
       at SwaggerTests.SwaggerTests.SwaggerTest() in C:\...\SwaggerTests.cs:line 12

Enhance the exception message by adding the HTTP status code and HTTP response.

Although the exception contains The HTTP status code and HTTP response as
properties, the unit tests cannot be taught to output them without explicitly
handling them in every single test. An alternative approach is to proxy the
swagger clients in order to intercept method invocations and wrap the exception
in a custom exception (using DynamicProxy or DynamicObject+ImpromptuInterface).
This is however not possible because the methods are neither virtual nor do
they implement an interface. It also seems to be impossible to intercept the
test methods themselves.

Closes #1605.

### DIFF
--- a/src/NSwag.CodeGeneration.CSharp/Templates/File.liquid
+++ b/src/NSwag.CodeGeneration.CSharp/Templates/File.liquid
@@ -118,7 +118,7 @@ namespace {{ Namespace }}
         public System.Collections.Generic.Dictionary<string, System.Collections.Generic.IEnumerable<string>> Headers { get; private set; }
 
         public {{ exceptionClassName }}(string message, int statusCode, string response, System.Collections.Generic.Dictionary<string, System.Collections.Generic.IEnumerable<string>> headers, System.Exception innerException) 
-            : base(message + "\n\nStatus: " + statusCode + "\nResponse: \n" + response, innerException)
+            : base(message + "\n\nStatus: " + statusCode + "\nResponse: \n" + response.Substring(0, response.Length >= 512 ? 512 : str.Length), innerException)
         {
             StatusCode = statusCode;
             Response = response; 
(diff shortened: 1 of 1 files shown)


RSuter:
I've switched to string concatenation (to support older C# compilers), added some line breaks, and truncate the response, so that the message can't get too big... is that ok?

cbuchacher:
Looks good to me. Thanks.

PR #1621, opened 2018-09-21
Added more Swagger UI 3 display settings

APErebus:
There is a number of Swagger UI 3 display settings that aren't currently configurable.

I specifically wanted the ability to hide the Models section at the bottom, which if you pass `-1` to `defaultModelsExpandDepth` will hide it.

I also included a couple of other options that seemed appropriate to include.

I also included a custom plugin to allow the disabling of the `Try it out` options. This wasn't something I wanted in our UI (as our auth is a bit fiddly), so this just removes that button.

### DIFF
--- a/src/NSwag.AspNet.Owin/SwaggerUi3/index.html
+++ b/src/NSwag.AspNet.Owin/SwaggerUi3/index.html
@@ -45,6 +45,22 @@ window.onload = function() {
   }
   var urls = {Urls};
 
+  const disableTryItOutPlugin = function() {
+    return {
+      statePlugins: {
+      spec: {
+        wrapSelectors: {
+          allowTryItOutFor: function() {
+            return function() {
+                return {EnableTryItOut};
+              }
+            }
+          }
+        }
+      }
+    }
+  }
+
   // Build a system
   var ui = SwaggerUIBundle({
     url: url,
@@ -54,6 +70,9 @@ window.onload = function() {
     docExpansion: "{DocExpansion}",
     apisSorter: "{ApisSorter}",
     operationsSorter: "{OperationsSorter}",
+    defaultModelsExpandDepth: {DefaultModelsExpandDepth},
+    defaultModelExpandDepth: {DefaultModelExpandDepth},
+    tagSorter: "{TagSorter}",
 
     dom_id: '#swagger-ui',
     deepLinking: true,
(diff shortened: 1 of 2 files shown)


src/NSwag.AspNetCore/SwaggerUi3Settings.cs
```diff
@@ -39,6 +39,18 @@ public class SwaggerUi3Settings<T> : SwaggerUiSettingsBase<T>
...
+        public int DefaultModelsExpandDepth { get; set; } = 1;
+
+        /// <summary>The default expansion depth for the model on the model-example section in Swagger UI 3.</summary>
+        public int DefaultModelExpandDepth { get; set; } = 1;
+
+        /// <summary>Specifies the tag sorter in Swagger UI 3</summary>
+        public string TagSorter { get; set; } = "none";
+
+        /// <summary>Specifies whether the "Try it out" option is disabled in Swagger UI 3.</summary>
+        public bool DisableTryItOut { get; set; } = false;
```
RSuter: I think it would be better to rename this to EnableTryItOut and set to true (avoid negative bools).

src/NSwag.AspNet.Owin/SwaggerUi3/index.html
```diff
@@ -45,6 +45,18 @@
   }
   var urls = {Urls};
 
+  const disableTryItOutPlugin = function() {
+    return {
+      statePlugins: {
+      spec: {
+        wrapSelectors: {
+            allowTryItOutFor: () => () => !{DisableTryItOut}
```
RSuter: Please use ES3 javascript to support all browsers (ie old school functions instead of lambda)

RSuter:
Thanks for the PR! Looks good. I’ve added two comments...

src/NSwag.AspNet.Owin/SwaggerUi3/index.html
APErebus: Yep! Will do

src/NSwag.AspNetCore/SwaggerUi3Settings.cs
APErebus: Can do.

APErebus:
@RSuter updated. Thanks for the feedback

PR #1647, opened 2018-10-03
Owin SwaggerMiddleware IIS issue with SwaggerUI and Application folder

GFlisch:
When deploying on IIS configured with an application folder, the middleware is too strict and doesn't take into account the application folder in the path checks.


RSuter:
I think there’s an additional basepath (or similar) setting for that

PR #1655, opened 2018-10-05
Add CreateWebHostBuilderMethod & AspNetCoreEnvironment setting, implement UseDocumentProvider & DocumentName

### LINKED ISSUE #1654: Discussion: aspnetcore2swagger configuration
Hi @RSuter 

I was digging into how nswag launches a web application, because I was hitting nullref in some places. My discovery was that the [CreateTestServer](https://github.com/RSuter/NSwag/blob/master/src/NSwag.Commands/Commands/SwaggerGeneration/SwaggerGeneratorCommandBase.cs#L206-L209) method here just launches a test server using the Startup class of the target project, this mean the code into the Program.cs file is not run and AFIK the `new WebHostBuilder().UseStartup(startupType)` is not setting up all the defaults like `WebHost.CreateDefaultBuilder` is doing e.g. the only configuration provider (the root cause for my nullrefs) configured when `Startup` is invoked by the **aspnetcore2swagger** command is the [EnvironmentVariablesConfigurationProvider](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.configuration.environmentvariables.environmentvariablesconfigurationprovider?view=aspnetcore-2.1).

What do you think about changing the `CreateTestServer` method from `new WebHostBuilder().UseStartup(startupType)` to `WebHost.CreateDefaultBuilder().UseStartup(startupType)` to have aspnetcore2swagger invoke and run our applications with an environment more close to the development one?

### LINKED ISSUE #1698: Console / NSwag.MSBuild - Adding TypeMappers to AspNetCoreToSwaggerGeneratorSettings
There seems to missing a way to configure `AspNetCoreToSwaggerGeneratorSettings` dynamically when generating code using the MSBuild pipeline.

We have custom `TypeMapper`s that we are adding during `Startup`. However those mappings are not applied when generating clients using MSBuild, causing discrepancies between our actual schema and the clients generated.

We are applying the following mappers (to support NodaTime's Instant and LocalDate):

```csharp
var typeMappers = settings.TypeMappers;

typeMappers.Add(new PrimitiveTypeMapper(
    typeof(LocalDate),
    jsonSchema =>
    {
        jsonSchema.Type = JsonObjectType.String;
        jsonSchema.Format = JsonFormatStrings.Date;
    }));

typeMappers.Add(new PrimitiveTypeMapper(
    typeof(Instant),
    jsonSchema =>
    {
        jsonSchema.Type = JsonObjectType.String;
        jsonSchema.Format = JsonFormatStrings.DateTime;
    }));
```

Is there an undocumented way of applying the same configuration during generation?

RSuter:
Closes #1654, closes #1698

- [x] Add CreateWebHostBuilderMethod setting for scenarios where no (default) Startup.cs is available
    - StartupType setting is ignored when CreateWebHostBuilderMethod is defined
- [x] Add UseDocumentProvider setting support
    - Add DocumentName setting (default: v1)
- [x] Add AspNetCoreEnvironment setting

Breaking changes: 

- UseDocumentProvider will be set to true by default for new documents (previously it was false)
    - Old nswag.json documents without UseDocumentProvider setting will be transformed to false
    - To keep old behavior in the aspnetcore2swagger command you need to explicitly set `/UseDocumentProvider:false`
- Used NetFx in commands upgraded from 4.5.1 to 4.6.1

Usage of UseDocumentProvider:

image


src/NSwag.Commands/Commands/SwaggerGeneration/SwaggerGeneratorCommandBase.cs
```diff
@@ -203,25 +208,54 @@ public async Task<TSettings> CreateSettingsAsync(AssemblyLoader.AssemblyLoader a
...
             {
-                // TODO: Implement proper auto discovery
-                var assemblies = await LoadAssembliesAsync(AssemblyPaths, assemblyLoader).ConfigureAwait(false);
-                startupType = assemblies.First().ExportedTypes.First(t => t.Name == "Startup");
+                // Load configured CreateWebHostBuilder method from program type
+                var segments = CreateWebHostBuilderMethod.Split('.');
+                var programTypeName = string.Join(".", segments.Take(segments.Length - 1));
+                var programType = assemblyLoader.GetType(programTypeName);
+                var method = programType.GetRuntimeMethod(segments.Last(), new[] { typeof(string[]) });
+                return new TestServer((IWebHostBuilder)method.Invoke(null, new object[] { new string[0] }));
```
ilmax: do we want to handle also the case the method doesn't have parameters?

src/NSwag.Commands/Commands/SwaggerGeneration/SwaggerGeneratorCommandBase.cs
```diff
@@ -203,25 +208,54 @@ public async Task<TSettings> CreateSettingsAsync(AssemblyLoader.AssemblyLoader a
...
-            if (string.IsNullOrEmpty(StartupType))
+            if (!string.IsNullOrEmpty(CreateWebHostBuilderMethod))
             {
-                // TODO: Implement proper auto discovery
-                var assemblies = await LoadAssembliesAsync(AssemblyPaths, assemblyLoader).ConfigureAwait(false);
-                startupType = assemblies.First().ExportedTypes.First(t => t.Name == "Startup");
+                // Load configured CreateWebHostBuilder method from program type
+                var segments = CreateWebHostBuilderMethod.Split('.');
+                var programTypeName = string.Join(".", segments.Take(segments.Length - 1));
+                var programType = assemblyLoader.GetType(programTypeName);
```
ilmax: Do we want to trow a meaningful exception when the type is not found?

src/NSwag.Commands/Commands/SwaggerGeneration/SwaggerGeneratorCommandBase.cs
RSuter: Good point, was not sure if this is a possible case...

ilmax:
@RSuter Sorry for being late, I was able to try this one a bit more in depth, so I will write my findings here:

**MsBuild integration**
I was able to run the program, but there are a couple of issues here:

1. It will and run the program, but reading from configuration does not work because [WebHost.CreateDefaultBuilder](https://github.com/aspnet/MetaPackages/blob/f245512f6e68d65309b65528d479f32b34c67718/src/Microsoft.AspNetCore/WebHost.cs#L154) uses the current directory (which is e.g. bin/Debug/netcoreapp2.1 in my case) so I had to override it in Program.cs with:
```csharp
.ConfigureAppConfiguration((context, config) =>
                {
                    config.SetBasePath(@"TheProperPath");
                })
                .UseStartup<Startup>();
```
2. after fixing this one, the app launches, but I got a BadImageFormatException
```
 System.BadImageFormatException: Could not load file or assembly 'System.Data.SqlClient, Version=4.4.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'. Reference assemblies should not be loaded for execution.  They can only be loaded in the Reflection-only loader context. (Exception from HRESULT: 0x80131058)
  File name: 'System.Data.SqlClient, Version=4.4.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' ---> System.BadImageFormatException: Cannot load a reference assembly for execution.
     at Microsoft.EntityFrameworkCore.SqlServerRetryingExecutionStrategy.ShouldRetryOn(Exception exception)
     at Microsoft.EntityFrameworkCore.Storage.ExecutionStrategy.ExecuteImplementation[TState,TResult](Func`3 operation, Func`3 verifySucceeded, TState state)
     at Microsoft.EntityFrameworkCore.Storage.ExecutionStrategy.Execute[TState,TResult](TState state, Func`3 operation, Func`3 verifySucceeded)
     at Microsoft.EntityFrameworkCore.ExecutionStrategyExtensions.Execute[TState,TResult](IExecutionStrategy strategy, Func`2 operation, Func`2 verifySucceeded, TState state)
     at Microsoft.EntityFrameworkCore.Storage.RelationalDatabaseCreator.EnsureCreated()
     at Test.Startup.Configure(IApplicationBuilder app, IHostingEnvironment env, ILoggerFactory loggerFactory, IReflectedPermissionService reflectedPermissionService) in C:\Code\TicketingSolution\PortalAPI\Startup.cs:line 149
  --- End of stack trace from previous location where exception was thrown ---
     at Microsoft.AspNetCore.Hosting.ConventionBasedStartup.Configure(IApplicationBuilder app)
     at Microsoft.Extensions.DependencyInjection.IServiceCollectionExtensions.InjectApiVersionRoutePolicy.<>c__DisplayClass2_0.<Configure>b__0(IApplicationBuilder app)
     at Microsoft.AspNetCore.Hosting.Internal.WebHost.BuildApplication()
```

Please note I'm using the copy @ReferencePaths trick in my csproj e.g. 

```xml
<Copy SourceFiles="@(ReferencePath)" DestinationFolder="$(OutDir)References" />
```

**NSwagStudio**
After setting the Environment via command line
```
set ASPNETCORE_ENVIRONMENT=Development
NSwagStudio.exe
```
Setting the new field appropriately and the correct runtime, I was able to generate the swagger document!

RSuter:
> It will and run the program, but reading from configuration does not work because WebHost.CreateDefaultBuilder uses the current directory (which is e.g. bin/Debug/netcoreapp2.1 in my case) so I had to override it in Program.cs with:

I think there is an option to set the working directory (WorkingDirectory setting?) - have you tried that?

ilmax:
@RSuter yes, I've tried setting the correct location with:

```
.ConfigureAppConfiguration((context, config) =>
	{
		config.SetBasePath(@"TheProperPath");
	})
	.UseStartup<Startup>();
```

But it still result in an exception (see message above)

RSuter:
Im talking about the workingDirectory setting of the nswag cli

ilmax:
Oh, sorry I didn't get that one, I can try again and get back to you

RSuter:
I'll add the AspNetCoreEnvironment setting in the next hours, fix your suggestions and merge this PR

ilmax:
👍

src/NSwag.Commands/Commands/SwaggerGeneration/SwaggerGeneratorCommandBase.cs
RSuter: Added support for method without params + improved exception handling

src/NSwag.Commands/Commands/SwaggerGeneration/SwaggerGeneratorCommandBase.cs
RSuter: Added support for method without params + improved exception handling

RSuter:
@ilmax ok, PR is ready and I now start running some integration tests with it... it should now completely cover your scenarios. Also see updated PR description.

If possible, please test with the following settings:

- use new CreateWebHostBuilderMethod setting
- use new AspNetCoreEnvironment setting
- use WorkingDirectory setting

If everything works as expected, we can merge.

ViRuSTriNiTy:
How exactly do i use the setting `CreateWebHostBuilderMethod `? I cannot find any documentation. The comment in the sources stating `The CreateWebHostBuilder method in the form 'assemblyName:fullTypeName.methodName' or 'fullTypeName.methodName'."` is not saying much, i tried tousand different values getting various exceptions. I assume you had an example why testing this? Where can i find it?

Issue #1626, opened 2018-09-21
Add an `IDocumentProvider` interface and service implementing it


dougbu:
The overall aim of in-progress work in ASP.NET Core 2.2 is to give customers simple ways to annotate their project files for automated code generation during every build (as source API descriptions change). See ASP.NET Core Mvc's [feature/client.code.generation](https://github.com/aspnet/Mvc/tree/feature/client.code.generation) branch and the aspnet/Mvc#7947 high-level issue.

The ASP.NET Core work includes a "generic" document-fetching tool which can get the Open API / Swagger document from a web site project. This tool can download from a URI (with the site running in `TestHost`) but is easier to use when the document provider (e.g. NSwag.AspNetCore) adds a service implementing `Microsoft.Extensions.ApiDescription.IDocumentProvider`. In the NSwag case, the retrieved document would reflect the user's configuration -- unlike when using the `aspnetcore2swagger` CLI tool.

Specific work in this repo would include:
- define chosen `IDocumentProvider` interface in the NSwag.AspNetCore package
  - the generic tool in the Microsoft.Extensions.ApiDescription.Client package does not define this `interface`, allowing providers to choose their favourite signature
- implement `IDocumentProvider` in the same package
- update the `AddSwagger(...)` extension method to register the new `IDocumentProvider` service
  - will likely also require a few other new services in DI, making at least the necessary configuration information available

RSuter:
- AddSwagger() should register this new interface
- UseSwagger*() methods should add the settings to IDocumentProvider service
    - Add a new "option" so that the registered document can have a name/identifier
- IDocumentProvider should serve all documents registered by the UseSwagger*() methods
    - Should provide an api to enumerate all registered documents and name
    - Should provide an api to generate multiple/all documents or just one document by name

rynowak:
Hi Rico,

Some questions about this:

> AddSwagger() should register this new interface
> UseSwagger*() methods should add the settings to IDocumentProvider service

Currently you don't require `AddSwagger` to be called when using `UseSwagger()` - so is it your expectation that those cases would degrade gracefully when `AddSwagger` wasn't called?

> Add a new "option" so that the registered document can have a name/identifier

Does this go on `SwaggerSettings<>`? On `SwaggerGeneratorSettings`? Or external to both:
```
ISwaggerGeneratorSettingsRegistry
{
    SwaggerGeneratorSettings this[string documentName] { get; set; }
}
```

rynowak:
[...]

Issue #984, opened 2017-10-09
Epic: NSwag v12


RSuter:
Next major version v13: https://github.com/RSuter/NSwag/issues/1700

**Breaking changes:**

There are breaking changes but most of them can be fixed by using the renamed method or change a setting. 

Settings

- **The default settings have been updated**, see PR #1719 for details
    - When using an nswag.json file, the old settings are retained and this is not a breaking change.
    - When using the CLI directly (e.g. swagger2csharp command) then these defaults **may be a breaking change** and the previous settings have to be set explicitely to generate the previous output.

Runtime (CLI, NPM, NSwagStudio)

- **Default runtime set to NetCore21 (before WinX64)**
    - Change this in NPM command line with the `/runtime:WinX64` parameter
    - The WinX86/WinX64 command line ASP.NET Core dependencies have been upgraded to 2.1.x (#1696)
    - Upgrade ASP.NET Core dependencies (x64/x86) (#1696) 
    - For more information, see https://github.com/RSuter/NSwag/wiki/Assembly-loading

NSwag.AspNetCore (middlewares)

- **The API of the ASP.NET Core middlewares have been changed** (easy fix)
    - All public **and** not obsolete methods are now completely based on [AspNetCoreToSwaggerGenerator](https://github.com/RSuter/NSwag/wiki/AspNetCoreToSwaggerGenerator)
        - The old [WebApiToSwaggerGenerator](https://github.com/RSuter/NSwag/wiki/WebApiToSwaggerGenerator) based methods are still available but deprecated and will be removed in v13 (#1700)
    - `AddSwaggerDocument()/AddOpenApiDocument()`
        - Document configuration moved to `AddSwaggerDocument(configure)` in `ConfigureServices()` (before in `UseSwaggerWithApiExplorer()`)
    - `UseSwaggger()`
        - `UseSwagggerWithApiExplorer()` is now just `UseSwagger()` in `Configure()`
        - `SwaggerRoute` renamed to `Path` (#1728)
    - `UseSwaggerUi*()`
        - Web UI methods now **only register the UI** and do not serve `swagger.json` anymore (use an additional call to `UseSwagger()`)
        - `UseSwaggerUi3WithApiExplorer()` no longer exists, use `UseSwagger()` and `UseSwaggerUi3()` in `Configure()`
        - `UseSwaggerReDoc()` renamed to `UseReDoc()`
        - `SwaggerUiRoute` renamed to `Path` (#1728)
[...]

PR #1658, opened 2018-10-09
Add `IDocumentProvider` service and its implementation

dougbu:
- #1626
- does not "provide an api to generate multiple / all documents"; could be layered on this work
  - otherwise, meets requirements listed in #1626
- assumes the following answers to open questions in #1626 (generally means everything is opt-in)
  - without an `AddSwagger()` call, `IDocumentProvider` service will not exist
  - registered document name / identifier is new `SwaggerSettings<>.DocumentName` property
    - document is not registered unless `DocumentName` is non-`null` i.e. `IDocumentProvider` won't know about it
    - `SwaggerSettings<>.MiddlewareBasePath` and `SwaggerRoute` are independent of `DocumentName`
  - middleware does not depend on the `IDocumentProvider` or registrar services; duplicating some code
- many thanks to @rynowak for writing this proposal

### DIFF
--- a/src/NSwag.AspNetCore/DocumentRegistry.cs
+++ b/src/NSwag.AspNetCore/DocumentRegistry.cs
@@ -1,0 +1,52 @@
+﻿//-----------------------------------------------------------------------
+// <copyright file="DocumentRegistry.cs" company="NSwag">
+//     Copyright (c) Rico Suter. All rights reserved.
+// </copyright>
+// <license>https://github.com/NSwag/NSwag/blob/master/LICENSE.md</license>
+// <author>Rico Suter, mail@rsuter.com</author>
+//-----------------------------------------------------------------------
+
+using System;
+using System.Collections.Generic;
+
+namespace NSwag.AspNetCore
+{
+    internal class DocumentRegistry
+    {
+        private readonly Dictionary<string, RegisteredDocument> _documents;
+
+        public DocumentRegistry()
+        {
+            _documents = new Dictionary<string, RegisteredDocument>(StringComparer.Ordinal);
+        }
+
+        public RegisteredDocument this[string documentName]
+        {
+            get
+            {
+                if (documentName == null)
+                {
+                    throw new ArgumentNullException(nameof(documentName));
+                }
+
+                _documents.TryGetValue(documentName, out var document);
+                return document;
+            }
+
+            set
+            {
+                if (documentName == null)
+                {
+                    throw new ArgumentNullException(nameof(documentName));
+                }
+
+                if (value == null)
+                {
+                    throw new ArgumentNullException(nameof(value));
+                }
+
+                _documents[documentName] = value;
+            }
+        }
+    }
+}
--- a/src/NSwag.AspNetCore/IDocumentProvider.cs
+++ b/src/NSwag.AspNetCore/IDocumentProvider.cs
@@ -1,0 +1,20 @@
+﻿//-----------------------------------------------------------------------
+// <copyright file="IDocumentProvider.cs" company="NSwag">
+//     Copyright (c) Rico Suter. All rights reserved.
+// </copyright>
+// <license>https://github.com/NSwag/NSwag/blob/master/LICENSE.md</license>
+// <author>Rico Suter, mail@rsuter.com</author>
+//-----------------------------------------------------------------------
+
+using System.IO;
+using System.Threading.Tasks;
+
+namespace Microsoft.Extensions.ApiDescription
+{
+    // This service will be looked up by name from the service collection when using
+    // the Microsoft.Extensions.ApiDescription tool
+    internal interface IDocumentProvider
+    {
+        Task GenerateAsync(string documentName, TextWriter writer);
+    }
+}
--- a/src/NSwag.AspNetCore/NSwagDocumentProvider.cs
+++ b/src/NSwag.AspNetCore/NSwagDocumentProvider.cs
@@ -1,0 +1,116 @@
+﻿//-----------------------------------------------------------------------
+// <copyright file="NSwagDocumentProvider.cs" company="NSwag">
+//     Copyright (c) Rico Suter. All rights reserved.
+// </copyright>
+// <license>https://github.com/NSwag/NSwag/blob/master/LICENSE.md</license>
+// <author>Rico Suter, mail@rsuter.com</author>
+//-----------------------------------------------------------------------
+
+using System;
+using System.IO;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.AspNetCore.Mvc.ApiExplorer;
+using Microsoft.Extensions.ApiDescription;
+using Microsoft.Extensions.Options;
+using NSwag.SwaggerGeneration;
+using NSwag.SwaggerGeneration.AspNetCore;
+using NSwag.SwaggerGeneration.WebApi;
+
+namespace NSwag.AspNetCore
+{
+    internal class NSwagDocumentProvider : IDocumentProvider
+    {
+        private readonly IApiDescriptionGroupCollectionProvider _apiDescriptionGroupCollectionProvider;
+        private readonly IOptions<MvcOptions> _mvcOptions;
+        private readonly IOptions<MvcJsonOptions> _mvcJsonOptions;
+        private readonly DocumentRegistry _registry;
+
+        public NSwagDocumentProvider(
+            DocumentRegistry registry,
+            IApiDescriptionGroupCollectionProvider apiDescriptionGroupCollectionProvider,
+            IOptions<MvcOptions> mvcOptions,
+            IOptions<MvcJsonOptions> mvcJsonOptions)
+        {
+            if (apiDescriptionGroupCollectionProvider == null)
+            {
+                throw new ArgumentNullException(nameof(apiDescriptionGroupCollectionProvider));
+            }
+
+            if (mvcOptions == null)
+            {
+                throw new ArgumentNullException(nameof(mvcOptions));
+            }
+
+            if (mvcJsonOptions == null)
+            {
+                throw new ArgumentNullException(nameof(mvcJsonOptions));
+            }
+
+            if (registry == null)
+            {
+                throw new ArgumentNullException(nameof(registry));
+            }
+
+            _apiDescriptionGroupCollectionProvider = apiDescriptionGroupCollectionProvider;
+            _mvcOptions = mvcOptions;
+            _mvcJsonOptions = mvcJsonOptions;
+            _registry = registry;
+        }
+
+        public async Task<SwaggerDocument> GenerateAsync(string documentName)
+        {
+            if (documentName == null)
+            {
+                throw new ArgumentNullException(nameof(documentName));
+            }
+
+            var documentInfo = _registry[documentName];
+            if (documentInfo == null)
+            {
+                throw new InvalidOperationException($"No registered document found for document name '{documentName}'.");
+            }
+
+            documentInfo.Settings.ApplySettings(_mvcJsonOptions.Value.SerializerSettings, _mvcOptions.Value);
+
+            SwaggerDocument document;
+            if (documentInfo.Settings is AspNetCoreToSwaggerGeneratorSettings aspnetcore)
+            {
+                var generator = new AspNetCoreToSwaggerGenerator(aspnetcore, documentInfo.SchemaGenerator);
+                document = await generator.GenerateAsync(_apiDescriptionGroupCollectionProvider.ApiDescriptionGroups);
+            }
+            else if (documentInfo.Settings is WebApiToSwaggerGeneratorSettings webapi)
+            {
+                var generator = new WebApiToSwaggerGenerator(webapi, documentInfo.SchemaGenerator);
+                document = await generator.GenerateForControllersAsync(webapi.ControllerTypes);
+            }
+            else
+            {
+                throw new InvalidOperationException($"Unsupported settings type '{documentInfo.GetType().FullName}");
+            }
+
+            documentInfo.PostProcess?.Invoke(document);
+
+            return document;
+        }
+
+        // Called by the Microsoft.Extensions.ApiDescription tool
+        async Task IDocumentProvider.GenerateAsync(string documentName, TextWriter writer)
+        {
+            if (documentName == null)
+            {
+                throw new ArgumentNullException(nameof(documentName));
+            }
+
+            if (writer == null)
+            {
+                throw new ArgumentNullException(nameof(writer));
+            }
+
+            var document = await GenerateAsync(documentName);
+
+            var json = document.ToJson();
+            await writer.WriteAsync(json);
+        }
+    }
+}
--- a/src/NSwag.AspNetCore/RegisteredDocument.cs
+++ b/src/NSwag.AspNetCore/RegisteredDocument.cs
@@ -1,0 +1,61 @@
+﻿//-----------------------------------------------------------------------
+// <copyright file="RegisteredDocument.cs" company="NSwag">
+//     Copyright (c) Rico Suter. All rights reserved.
+// </copyright>
+// <license>https://github.com/NSwag/NSwag/blob/master/LICENSE.md</license>
+// <author>Rico Suter, mail@rsuter.com</author>
+//-----------------------------------------------------------------------
+
+using System;
+using System.Collections.Generic;
+using NSwag.SwaggerGeneration;
+using NSwag.SwaggerGeneration.AspNetCore;
+using NSwag.SwaggerGeneration.WebApi;
+
+namespace NSwag.AspNetCore
+{
+    // Captures the state required to produce a document. Mediates between two supported
+    // document generators.
+    internal class RegisteredDocument
+    {
+        public static RegisteredDocument CreateAspNetCoreGeneratorDocument(
+            AspNetCoreToSwaggerGeneratorSettings settings,
+            SwaggerJsonSchemaGenerator schemaGenerator,
+            Action<SwaggerDocument> postProcess)
+        {
+            return new RegisteredDocument()
+            {
+                PostProcess = postProcess,
+                SchemaGenerator = schemaGenerator,
+                Settings = settings,
+            };
+        }
+
+        public static RegisteredDocument CreateWebApiGeneratorDocument(
+            WebApiToSwaggerGeneratorSettings settings,
+            IEnumerable<Type> controllerTypes,
+            SwaggerJsonSchemaGenerator schemaGenerator,
+            Action<SwaggerDocument> postProcess)
+        {
+            return new RegisteredDocument()
+            {
+                ControllerTypes = controllerTypes,
+                PostProcess = postProcess,
+                SchemaGenerator = schemaGenerator,
+                Settings = settings,
+            };
+        }
+
+        private RegisteredDocument()
+        {
+        }
+
+        public IEnumerable<Type> ControllerTypes { get; private set; }
+
+        public Action<SwaggerDocument> PostProcess { get; private set; }
+
+        public SwaggerJsonSchemaGenerator SchemaGenerator { get; private set; }
+
+        public SwaggerGeneratorSettings Settings { get; private set; }
+    }
+}
--- a/src/NSwag.AspNetCore/SwaggerExtensions.cs
+++ b/src/NSwag.AspNetCore/SwaggerExtensions.cs
@@ -72,8 +72,30 @@ namespace NSwag.AspNetCore
         {
             var settings = new SwaggerSettings<WebApiToSwaggerGeneratorSettings>();
             configure?.Invoke(settings);
-            app.UseMiddleware<WebApiToSwaggerMiddleware>(settings.ActualSwaggerRoute, controllerTypes, settings, schemaGenerator ?? new SwaggerJsonSchemaGenerator(settings.GeneratorSettings));
-            return app;
+
+            schemaGenerator = schemaGenerator ?? new SwaggerJsonSchemaGenerator(settings.GeneratorSettings);
+
+            return UseSwaggerCore(app, controllerTypes, settings, schemaGenerator);
+        }
+
+        public static IApplicationBuilder UseSwaggerCore(
+            IApplicationBuilder app,
+            IEnumerable<Type> controllerTypes,
+            SwaggerSettings<WebApiToSwaggerGeneratorSettings> settings,
+            SwaggerJsonSchemaGenerator schemaGenerator)
+        {
+            if (settings.DocumentName != null)
+            {
+                var registry = app.ApplicationServices.GetService<DocumentRegistry>();
+                if (registry != null)
+                {
+                    // Register our settings in the registry if possible. The middleware doesn't actually depend
+                    // on any services, it does things the old way.
+                    registry[settings.DocumentName] = RegisteredDocument.CreateWebApiGeneratorDocument(settings.GeneratorSettings, controllerTypes, schemaGenerator, settings.PostProcess);
+                }
+            }
+
+            return app.UseMiddleware<WebApiToSwaggerMiddleware>(settings.ActualSwaggerRoute, controllerTypes, settings, schemaGenerator);
         }
 
         /// <summary>Adds the Swagger generator that uses Api Description to perform Swagger generation.</summary>
@@ -87,7 +109,29 @@ namespace NSwag.AspNetCore
         {
             var settings = new SwaggerSettings<AspNetCoreToSwaggerGeneratorSettings>();
             configure?.Invoke(settings);
-            return app.UseMiddleware<AspNetCoreToSwaggerMiddleware>(settings, schemaGenerator ?? new SwaggerJsonSchemaGenerator(settings.GeneratorSettings));
+
+            schemaGenerator = schemaGenerator ?? new SwaggerJsonSchemaGenerator(settings.GeneratorSettings);
+
+            return UseSwaggerWithApiExplorerCore(app, settings, schemaGenerator);
+        }
+
+        private static IApplicationBuilder UseSwaggerWithApiExplorerCore(
+            IApplicationBuilder app,
+            SwaggerSettings<AspNetCoreToSwaggerGeneratorSettings> settings,
+            SwaggerJsonSchemaGenerator schemaGenerator)
+        {
+            if (settings.DocumentName != null)
+            {
+                var registry = app.ApplicationServices.GetService<DocumentRegistry>();
+                if (registry != null)
+                {
+                    // Register our settings in the registry if possible. The middleware doesn't actually depend
+                    // on any services, it does things the old way.
+                    registry[settings.DocumentName] = RegisteredDocument.CreateAspNetCoreGeneratorDocument(settings.GeneratorSettings, schemaGenerator, settings.PostProcess);
+                }
+            }
+
+            return app.UseMiddleware<AspNetCoreToSwaggerMiddleware>(settings, schemaGenerator);
         }
 
         /// <summary>Adds services required for swagger generation.</summary>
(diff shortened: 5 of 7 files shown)


src/NSwag.AspNetCore/IDocumentProvider.cs
```diff
@@ -0,0 +1,20 @@
...
+//-----------------------------------------------------------------------
+
+using System.IO;
+using System.Threading.Tasks;
+
+namespace Microsoft.Extensions.ApiDescription
+{
+    // This service will be looked up by name from the service collection when using
+    // the Microsoft.Extensions.ApiDescription tool
+    internal interface IDocumentProvider
```
RSuter: I think we should rename it to IOpenApiDocumentProvider because it's specific to Swagger/OpenAPI3 documents. Otherwise I'd introduce a type enum just to be future proof.

RSuter: What do you think?

src/NSwag.AspNetCore/RegisteredDocument.cs
```diff
@@ -0,0 +1,61 @@
...
+                SchemaGenerator = schemaGenerator,
+                Settings = settings,
+            };
+        }
+
+        private RegisteredDocument()
+        {
+        }
+
+        public IEnumerable<Type> ControllerTypes { get; private set; }
```
RSuter: think we should support only the AspNetCoreToSwaggerGenerator and ignore the Web API (legacy) generator. I think this property would not be needed this way...

src/NSwag.AspNetCore/SwaggerExtensions.cs
```diff
@@ -98,7 +142,15 @@ public static class SwaggerExtensions
         /// </remarks>
         public static IServiceCollection AddSwagger(this IServiceCollection serviceCollection)
         {
-            return serviceCollection.AddSingleton<IConfigureOptions<MvcOptions>, NSwagConfigureMvcOptions>();
+            serviceCollection.AddSingleton<IConfigureOptions<MvcOptions>, NSwagConfigureMvcOptions>();
```
RSuter: Do you know why this is needed? Doesn't this overwrite/hide client settings?

src/NSwag.AspNetCore/IDocumentProvider.cs
```diff
@@ -0,0 +1,20 @@
...
+//     Copyright (c) Rico Suter. All rights reserved.
+// </copyright>
+// <license>https://github.com/NSwag/NSwag/blob/master/LICENSE.md</license>
+// <author>Rico Suter, mail@rsuter.com</author>
+//-----------------------------------------------------------------------
+
+using System.IO;
+using System.Threading.Tasks;
+
+namespace Microsoft.Extensions.ApiDescription
```
RSuter: Is it a good idea to add the Microsoft namespace to this package? This may confuse people. Or will this be moved into another library eventually?

src/NSwag.AspNetCore/IDocumentProvider.cs
```diff
@@ -0,0 +1,20 @@
...
+{
+    // This service will be looked up by name from the service collection when using
+    // the Microsoft.Extensions.ApiDescription tool
+    internal interface IDocumentProvider
+    {
+        Task GenerateAsync(string documentName, TextWriter writer);
```
RSuter: How do you know what document to use when you can't enumerate it? Does it have to be specified in the csproj?

src/NSwag.AspNetCore/NSwagDocumentProvider.cs
```diff
@@ -0,0 +1,116 @@
...
+
+            _apiDescriptionGroupCollectionProvider = apiDescriptionGroupCollectionProvider;
+            _mvcOptions = mvcOptions;
+            _mvcJsonOptions = mvcJsonOptions;
+            _registry = registry;
+        }
+
+        public async Task<SwaggerDocument> GenerateAsync(string documentName)
+        {
+            if (documentName == null)
```
RSuter: Maybe it would be better to allow null and throw an exception when more than one document is added with null? This way a single document - probably the most used scenario - is automatically added.

src/NSwag.AspNetCore/IDocumentProvider.cs
dougbu: This namespace is a marker for use in the tool we'll ship. It can take many forms i.e. is up to the document generator provider. So, linking it to the `Microsoft.Extensions.ApiDescription.Design` package, at least by inference, makes sense.

src/NSwag.AspNetCore/IDocumentProvider.cs
dougbu: Though I could be convinced, I'm fairly resistant to this idea.

Nothing in the `Microsoft.Extensions.ApiDescription.Design` package is specific to Open API or the earlier Swagger formats. All that's required is that the service return a document the code generator can handle. And, the various documents are all self-describing.

If NSwag supports new formats, I suggest using distinct `DocumentName`s for the different formats.

src/NSwag.AspNetCore/IDocumentProvider.cs
dougbu: The tool in the `Microsoft.Extensions.ApiDescription.Design` package relies on the user including `DocumentName` metadata on the relevant items in their `.csproj` file. If the metadata doesn't exist, the tool falls back to the name `"v1"`. If that doesn't match a document, it simply fails.

src/NSwag.AspNetCore/RegisteredDocument.cs
dougbu: :ok:

src/NSwag.AspNetCore/NSwagDocumentProvider.cs
dougbu: Te Microsoft tool won't pass `null` into `IDocumentProvider.GenerateAsync(...)` and instead uses `"v1"` by default. However, I could change registration to use `"v1"` when the new `DocumentName` property is `null`. Thoughts?

Separately, the registrar currently has a "last one wins" semantic. Not sure we should require distinct `DocumentName` properties in the initial release. Wouldn't that be a breaking change?

src/NSwag.AspNetCore/SwaggerExtensions.cs
dougbu: `NSwagConfigureMvcOptions` and the `NSwagMvcApplicationModelConvention` it adds existed before this PR. The convention just changes `ApplicationModel.ApiExplorer.IsVisible` to `true`.

@rynowak in what scenarios would that property change break something?

src/NSwag.AspNetCore/NSwagDocumentProvider.cs
RSuter: That would be a breaking change.

RSuter: Probably this should be a breaking change in v12. Good thing is that it's easy to communicate how to fix (exceptions). #984

dougbu: To clarify the resolution for this PR, I believe I should
1. Leave this `null` check alone
2. Change the registrations to use `"v1"` when `DocumentName` isn't set
3. Leave the "last one wins" semantic alone pending work in the v12 epic

On (2), would you prefer the `DocumentName` property simply have a default value of `"v1"`?

dougbu:
:up::date:d to address your comments @RSuter. I look forward to any alternative suggestions or further thoughts you have.

src/NSwag.AspNetCore/SwaggerExtensions.cs
rynowak: I am a little confused maybe? This was here before.

My understanding of what this does aligns with Dougs. 

That said, with 2.1+ we now enable ApiExplorer visibility by default for `[ApiController]` controller types. https://github.com/aspnet/Mvc/blob/release/2.1/src/Microsoft.AspNetCore.Mvc.Core/Internal/ApiBehaviorApplicationModelProvider.cs#L64

RSuter: I saw that this was already there but wondered if this construct is a problem somehow?

src/NSwag.AspNetCore/NSwagDocumentProvider.cs
RSuter: I think we should do 2 and 3 and just use a default value of "v1"

src/NSwag.AspNetCore/IDocumentProvider.cs
RSuter: Sounds good.

src/NSwag.AspNetCore/SwaggerExtensions.cs
rynowak: Oh I see. I would suggest updating your convention to look a little bit more like ours https://github.com/aspnet/Mvc/blob/release/2.1/src/Microsoft.AspNetCore.Mvc.Core/Internal/ApiBehaviorApplicationModelProvider.cs#L61

In particular if the visiblity is null, then you can set it to true. This allows users to opt controllers out of ApiExplorer explicitly.

src/NSwag.AspNetCore/IDocumentProvider.cs
```diff
@@ -0,0 +1,20 @@
...
+//-----------------------------------------------------------------------
+
+using System.IO;
+using System.Threading.Tasks;
+
+namespace Microsoft.Extensions.ApiDescription
+{
+    // This service will be looked up by name from the service collection when using
+    // the Microsoft.Extensions.ApiDescription tool
+    public interface IDocumentProvider
```
RSuter: So this interface is the only public API which is relevant for you, is this correct? 

Shouldn't it be renamed to IOpenApiDocumentProvider or also use a document type enum - because there might be other document types in the future....

rynowak: This doesn't need to be public

rynowak: This is a contract that we're implementing via late binding - it's not going to have types like enums - because there's no library. It's a way for us to consume a document in a generic way.

If you want a different abstraction for inside NSwag that sounds good, but it's not what we will consume.

RSuter: Ok, if we can make it internal and it doesnt give name collisions problems with other libraries (e.g. Swashbuckle in the same app) I think it's fine.

src/NSwag.AspNetCore/IDocumentProvider.cs
dougbu: Tool can now find this `internal` type. Turned out the issue wasn't the visibility modifier at all.

src/NSwag.AspNetCore/NSwagDocumentProvider.cs
dougbu: Done

src/NSwag.AspNetCore/DocumentRegistry.cs
```diff
@@ -0,0 +1,52 @@
...
+
+namespace NSwag.AspNetCore
+{
+    internal class DocumentRegistry
+    {
+        private readonly Dictionary<string, RegisteredDocument> _documents;
+
+        public DocumentRegistry()
+        {
+            _documents = new Dictionary<string, RegisteredDocument>(StringComparer.Ordinal);
```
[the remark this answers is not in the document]
RSuter: I'd keep it as is

src/NSwag.AspNetCore/IDocumentProvider.cs
RSuter: Resolve as it is internal only now.

src/NSwag.AspNetCore/SwaggerExtensions.cs
RSuter: Created new issue: https://github.com/RSuter/NSwag/issues/1671 for later

RSuter:
@dougbu is this now ready to be merged?

dougbu:
> is this now ready to be merged?

It won't harm anything to merge this (though squashing the 10 commits might help 😃.

But, based on our offline thread about the `Startup.Configure(...)` method, we might go with a different design -- especially w.r.t. the document registry. Suggest holding off until you choose an option.

RSuter:
@dougbu I made a PR against the branch in this PR, https://github.com/dougbu/NSwag/pull/1

I think this should cover scenario 3 (change middleware api). Is this ok and will work with your other code?

dougbu:
@RSuter well, I was just testing the changes I'd made. I'll push to my remote branch so that you can compare and contrast&hellip;

RSuter:
I'm now trying to merge all my work but it's now a big mess. But there are thee main problems: 

- With your approach, we cannot add a default document because we cannot detect when the user didnt provide one (i.e. AddOpenApi() is not able to detect that AddDocument is not called
- AddOpenApi and AddDocument can add a document ATM which doesnt make sense from an API perspective, see


```
            services
                .AddOpenApi(settings => settings.DocumentName = "v1") // Why the settings here?
                .AddDocument(settings =>
                {
                    settings.DocumentName = "v2";
                    settings.GeneratorSettings.SchemaType = SchemaType.OpenApi3;
                });
```

- If we break all users, I want to split the swagger.json and swagger UI middlewares into separate Use* functions to reduce the number of methods and not do two things with one method... 

Will update my PR against your PR soon.

dougbu:
> With your approach, we cannot add a default document because we cannot detect when the user didn't provide one (i.e. AddOpenApi() is not able to detect that AddDocument is not called

`AddOpenApi()` adds the first document. `AddDocument(...)` is only for additional documents.

> `// Why the settings here?`

Because users may want to configure the first, possibly only, document.

> separate Use* functions to reduce the number of methods and not do two things with one method.

I tried that approach but found it falls down because the UI middleware needs to know where the generator middleware is hosted. But, this will be much simpler to do once I move to an `IOptions<T>` approach.

> I think documentName should not be part of the options/settings as it must be clear that it's only used to connect AddSwagger with UseSwagger (which IMO is not really nice, but there is not a better solution with how ASP.NET Core is designed)

:ok: though I'm not quite sure I understand the concern.

All middleware in my approach handles all the documents. Well, except the Redoc UI one. Names are primarily important for routing and the `IDocumentProvider` calls.

> keep the name Swagger

Sure.

src/NSwag.Sample.OpenApi/NSwag.Sample.OpenApi.csproj
```diff
@@ -0,0 +1,19 @@
+﻿<Project Sdk="Microsoft.NET.Sdk.Web">
```
dougbu: Now that I look again at the existing samples, this one isn't needed. And, I started with the wrong template here, ignoring the lack of actual APIs exposed and bulking the project up with irrelevant files.

src/NSwag.AspNetCore/OpenApiExtensions.cs
```diff
@@ -0,0 +1,239 @@
...
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.ApiDescription;
+using Microsoft.Extensions.DependencyInjection.Extensions;
+using Microsoft.Extensions.FileProviders;
+using NSwag.AspNetCore;
+using NSwag.AspNetCore.DependencyInjection;
+using NSwag.AspNetCore.Middlewares;
+using NSwag.SwaggerGeneration.AspNetCore;
+
+namespace Microsoft.Extensions.DependencyInjection
```
dougbu: The namespace here is very important. Users should be able to do mainline scenarios as soon as they add the NSwag.AspNetCore package to their project. No need to add a `using NSwag.*` statement in their `Startup` class.

RSuter:
> AddOpenApi() adds the first document. AddDocument(...) is only for additional documents.

I think this API is really confusing because AddOpenApi() and AddDocument() are doing the same but have a different name... That's why I'd like to have this nested configuration, i.e. AddOpenApi() adds the required services and AddDocument (as in `AddOpenApi(options => options.AddDocument()`) actually adds a document - if none is specified a default one is regeistered by AddOpenApi. 

> 🆗 though I'm not quite sure I understand the concern.

I think it should be like this in a common scenario (simplest case): 

AddSwagger() // registers services
UseSwagger() // register swagger.json middleware
UseSwaggerUi3() // register swagger.json consuming UI

This way each call has its own defined responsibility...

> But, this will be much simpler to do once I move to an IOptions<T> approach.

How should this solve this structural problem? Can you give an example?

RSuter:
And let's not use OpenApi in the method and stick with Swagger for now... otherwise it gets really confusing because OpenApi does not enforce OpenAPI v3.0 and v2.0 is still supported so we have two methods which should do the same but internally dont (one registers the services the other doesnt).

src/NSwag.AspNetCore/OpenApiExtensions.cs
RSuter: Without a call to AddSwagger() the interface will not be registered and thus the whole thing wont work, right?

dougbu: That's a different concern. Everything fails nice and loudly when `AddSwagger()` is not called but the `Use*(...)` methods are.

dougbu:
> AddOpenApi() and AddDocument() are doing the same but have a different name

Not quite. `AddOpenApi(...)` does the whole shebang and `AddDocument(...)` just extends the available documents. Note `AddDocument(...)` isn't an extension method on the `IServiceCollection`.

> I think it should be like this in a common scenario (simplest case)

That will work fine using my approach as well. The `AddSwagger()` call registers the default document, `UseSwagger()` uses the default `IOptions<SwaggerMiddlewareOptions>` and `UseSwaggerUi3()` uses the default `IOptions<SwaggerUi3Options>` and also the `SwaggerRoute` set in `IOptions<SwaggerMiddlewareOptions>`.

Users will be able to mess themselves up by calling `UseSwaggerUi3()` without `UseSwagger()` or by passing `SwaggerMiddlewareOptions` into `UseSwagger()` that doesn't have the same route as the default. But, both cases are avoidable e.g. by having the middleware flip a bit in `SwaggerMiddlewareOptions` to indicate they're actually in use.

> How should this solve this structural problem?

Users would be able to stick with the usual ASP.NET Core configuration patterns. (MVC and EF get a bit more complicated because they aren't themselves exposed as middleware.)

For example, `ConfigureServices(...)` might contain
``` c#
services.AddSwagger();
services.Configure<SwaggerMiddlewareOptions>(o => o.SwaggerRoute = "/openApi/{documentName}/swagger.json");
services.Configure<SwaggerUi3Options>(o => o.SwaggerUiRoute = "/myUi");
```
then the `Use*()` methods in `Configure(...)` stitch everything together.

RSuter:
```
services.AddSwagger();
services.Configure<SwaggerMiddlewareOptions>(o => o.SwaggerRoute = "/openApi/{documentName}/swagger.json");
services.Configure<SwaggerUi3Options>(o => o.SwaggerUiRoute = "/myUi");
```

This is really bad as you can have only one SwaggerMiddlewareOptions or SwaggerUi3Options option... 

What if I need two Swagger UIs?
E.g one with the category a, one with category b and one with both (dropdown). See sample in my updated PR: https://github.com/dougbu/NSwag/pull/1

RSuter:
… also SwaggerMiddlewareOptions has a PostProcess (Action(request, document)) which cannot be specified per document this way...

RSuter:
> Users will be able to mess themselves up by calling UseSwaggerUi3() without UseSwagger() or by passing SwaggerMiddlewareOptions into UseSwagger() that doesn't have the same route as the default. But, both cases are avoidable e.g. by having the middleware flip a bit in SwaggerMiddlewareOptions to indicate they're actually in use.

Yes, but Swashbuckle does it the same way - and I'd rather trade this problem for more flexibility. If you don't change the routes it just works - otherwise you have to be aware that you can misconfigure the thing... 

For me these (AddSwagger, UseSwagger and UseSwaggerUi) are 3 different services... you can use UseSwaggerUi without the other two (e.g. with an external spec)

RSuter:
> Not true because the UseSwagger(...) and UseSwaggerUi3(...) methods have overloads accepting unique options.

Ah yes, makes sense - so either you configure it for all in ConfigureServices() or individually in Configure(), right?

> This is why I put the PostProcess property in the document settings.

Adding PostProcess to the generator settings was a mistake in the first place so I want to fix this... PostProcess should be only a middleware level feature to transform a document based on the current HttpContext (e.g. set correct hosts, etc.) and thus only be available in SwaggerMiddlewareOptions or SwaggerUi3Options  (and not in the generator settings itself - use a document processor for that instead). 

Can't we just merge my PR and add your IOptions support to that (applies for all and {documentName} is replaced)?

src/NSwag.AspNetCore/OpenApiExtensions.cs
```diff
@@ -0,0 +1,239 @@
...
+            return builder;
+        }
+
+        /// <summary>
+        ///
+        /// </summary>
+        /// <param name="builder"></param>
+        /// <param name="configure"></param>
+        /// <returns></returns>
+        public static IApplicationBuilder UseOpenApi(
```
rynowak: This should be in `Microsoft.Extensions.Builder`

src/NSwag.AspNetCore/Middlewares/AspNetCoreToOpenApiMiddleware.cs
```diff
@@ -0,0 +1,92 @@
...
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Routing;
+using Microsoft.AspNetCore.Routing.Internal;
+using Microsoft.AspNetCore.Routing.Template;
+using Microsoft.Extensions.ObjectPool;
+
+namespace NSwag.AspNetCore.Middlewares
+{
+    internal class AspNetCoreToOpenApiMiddleware
```
dougbu: This new middleware supports all registered documents in one instance, similar to the existing `SwaggerUi3Settings<T>.SwaggerRoutes` approach. Is there a reason you'd rather not do this?

dougbu:
> Ah yes, makes sense - so either you configure it for all in ConfigureServices() or individually in Configure(), right?

Exactly 😃

> PostProcess should be only a middleware level feature

That's fine though it seems better for `PostProcess` to operate on multiple documents rather than users having to configure the individual documents as well as a middleware per document to host it.

dougbu:
Latest commit means I can't apply `IOptions<T>` and related work I spent the day doing. Can that be backed out @RSuter?

RSuter:
Ok.

dougbu:
:up::date: only remaining WIP bits are a couple of names and a few empty doc comments.

From my perspective, this PR has the following advantages:
- uses `IOptions<T>` for `UseSwagger(...)` and `UseSwaggerUi4(...)` scenarios
- changes fewer files
- reduces the middleware depth when multiple documents are registered (by default)
- supports post-processing in `IDocumentProvider` scenarios
- places extensions and the `ISwaggerBuilder` interface in appropriate namespaces
- does not nest lambdas when using `AddSwagger(...)` or `AddOpenApi(...)`
  - note MVC uses nested lambdas only for changing `MvcOptions`, which would be relevant if `SwaggerDocumentProvider` took options
- leaves `SwaggerDocumentRegistry` with a single purpose
- leaves existing code unchanged except for a few nits

dougbu:
@RSuter please note the "Reviewers" comments in a few files

dougbu:
@RSuter I should have mentioned: If you choose this PR over #1678 and want me to, I'll squash the commits, perform any renaming, and fill in the blank doc comments. LMK

dougbu:
> I hope this is ok for you?

@RSuter yes, this is fine. Much better than fine: You've now got an `IDocumentProvider` implementation and have moved NSwag.AspNetCore to be more familiar to ASP.NET Core developers with plans to do more in v12!

I was somewhat sorry to see the middleware pipeline still contains multiple `SwaggerMiddleware` instances and that users don't *immediately* get `IOptions<T>` goodness. But, these are your decisions about your code. And, I agree this took more of both of our time than we'd hoped.

One more real concern, echoing my earlier "supports post-processing in `IDocumentProvider` scenarios" point: This lack in NSwag.AspNetCore may mean the Microsoft document tool will get a different document than the `SwaggerMiddleware` would provide. That is, users may do more than adjust to the request host and schemes in their `SwaggerMiddlewareSettings.PostProcess` actions. Can / should `SwaggerGeneratorSettings.DocumentProcessors` and `OperationProcessors` be used for that purpose?

RSuter:
> This lack in NSwag.AspNetCore may mean the Microsoft document tool will get a different document than the SwaggerMiddleware would provide. That is, users may do more than adjust to the request host and schemes in their SwaggerMiddlewareSettings.PostProcess actions. 

You are right, this also bothers me. The following code is actually not executed when running via command line: 
https://github.com/RSuter/NSwag/blob/master/src/NSwag.AspNetCore/Middlewares/SwaggerMiddleware.cs#L87-L91

The idea is that the PostProcess here should only be used to change BasePath, Schemes and Host based on the request, but of course it cannot be enforced. How would you proceed? 

- Remove these lines completely (breaking change for some, host and schemes may be wrong), 
- remove just PostProcess (but then you cannot change host, schemes, etc.) or 
- keep as is and add documentation that PostProcess should only be used to change host, schemes etc?

> Can / should SwaggerGeneratorSettings.DocumentProcessors and OperationProcessors be used for that purpose?

Yes, exactly - you should use a custom document processor for that - but you don't have access to a request there (because a spec is independent of an actual HTTP request).

dougbu:
> keep as is and add documentation that `PostProcess` should only be used to change host, schemes etc?

If it were me, this is the route I'd go because adjusting these values to match the request is so useful. Suggest also mentioning the processing order of `OperationProcessor`s, `DocumentProcessor`s and `PostProcess`.

Separately, including *all* schemes, hosts and `PathBase` values may be done using the [`IServerAddressesFeature`](https://github.com/aspnet/Hosting/blob/5a3c6645667ec24e09d6200fb1e949f2a2417b9a/src/Microsoft.AspNetCore.Hosting.Server.Abstractions/Features/IServerAddressesFeature.cs) in `IWebHost.ServerFeatures`.

>  you should use a custom document processor for that

Reasonable.

In retrospect, the main thing my trial addition of a `PostProcess` action to `SwaggerDocumentSettings` improved was ease of use for simple cases. Implementing `IDocumentProcessor` is slightly more involved but covers more cases because the `DocumentProcessorContext` provides more context.

RSuter:
> In retrospect, the main thing my trial addition of a PostProcess action to SwaggerDocumentSettings improved was ease of use for simple cases. Implementing IDocumentProcessor is slightly more involved but covers more cases because the DocumentProcessorContext provides more context.

Yes, but I don't like having two concepts for the same scenario... we could just provide a generic `GenericDocumentProcessor(Action<Context> action)` so that it's very easy to add a lambda/inline based document processor - this way you dont need a new class...

> If it were me, this is the route I'd go because adjusting these values to match the request is so useful. Suggest also mentioning the processing order of OperationProcessors, DocumentProcessors and PostProcess.

Ok, then we'll do that. But in the case of using the command line you need to define these properties via config anyway so why not force it also for the hosted scenario?

> Separately, including all schemes, hosts and PathBase values may be done using the IServerAddressesFeature in IWebHost.ServerFeatures.

Good point.

dougbu:
> so why not force it also for the hosted scenario?

I'm missing something. Force what?

If you mean forcing use of config, aren't `IDocumentProcessor`s and `PostProcess` already added through config?

RSuter:
I mean force the user to use the config to get the host etc so that it works in the host scenario and cli scenario.. but this might upset users :-)

dougbu:
I agree setting the document properties statically would be a regression.

After talking more internally, I'm sorry to say the `IServerAddressesFeature` isn't going to be helpful until after the production server is started, possibly after the first request. Even then, for the middleware case, you'll be able to detect that HTTP is enabled in addition to HTTPs but that won't be 100% reliable. (E.g. for an HTTP request -- the service may be deployed behind a TLS gateway and not itself support HTTPS. Or the service may support both schemes but be behind a reverse proxy that supports just one.) And, `IServerAddressesFeature` won't improve `IDocumentProvider` scenarios.

Issue #1602, opened 2018-09-13
204 nocontent exception


ahmettahasakar:
Hello

If i return a 204 nocontent, the typescript throws a swagger exception. However, it is a successfull return code. blobtotext works and then the pipe throws the exception. Am i missing something?

Thank you

RSuter:
Can you post some sample Swagger spec + TS output?

ahmettahasakar:
```
"delete": {
        "tags": [
          "HaileeResponses"
        ],
        "operationId": "HaileeResponses_Delete",
        "parameters": [
          {
            "type": "string",
            "name": "id",
            "in": "path",
            "required": true,
            "format": "guid",
            "x-nullable": false
          }
        ],
        "responses": {
          "200": {
            "x-nullable": true,
            "description": "",
            "schema": {
              "type": "file"
            }
          },
          "204": {
            "description": ""
          },
          "400": {
            "description": "Runtime error",
            "schema": {
              "type": "object",
              "additionalProperties": false,
              "properties": {
                "error": {
                  "$ref": "#/definitions/ErrorViewModel"
                }
              }
            }
          }
        }
      }
```

```
delete(id: string): Observable<FileResponse | null> {
        let url_ = this.baseUrl + "/HaileeResponses/{id}";
        if (id === undefined || id === null)
            throw new Error("The parameter 'id' must be defined.");
        url_ = url_.replace("{id}", encodeURIComponent("" + id)); 
        url_ = url_.replace(/[?&]$/, "");

        let options_ : any = {
            observe: "response",
            responseType: "blob",
            headers: new HttpHeaders({
                "Content-Type": "application/json", 
                "Accept": "application/json"
            })
        };

        return this.http.request("delete", url_, options_).pipe(_observableMergeMap((response_ : any) => {
            return this.processDelete(response_);
        })).pipe(_observableCatch((response_: any) => {
            if (response_ instanceof HttpResponseBase) {
                try {
                    return this.processDelete(<any>response_);
                } catch (e) {
                    return <Observable<FileResponse | null>><any>_observableThrow(e);
                }
            } else
                return <Observable<FileResponse | null>><any>_observableThrow(response_);
        }));
    }

    protected processDelete(response: HttpResponseBase): Observable<FileResponse | null> {
        const status = response.status;
        const responseBlob = 
            response instanceof HttpResponse ? response.body : 
            (<any>response).error instanceof Blob ? (<any>response).error : undefined;

[...]

Issue #1259, opened 2018-04-05
Merge 2xx responses with the same type and handle them in one block


RSuter:
merge 2xx responses with the same type and handle them in one if (status... || block

jeremyVignelles:
If 204 (No content) is in the possible answers, just make the result nullable.

The point is : 2xx status codes should not throw in the client, because they semantically indicate success.

What would be the behavior if two different types are declared, e.g. for 200 and 201 ?

- Will code generation for C# crash, since typescript can handle that?
- Will that automatically create a wrapper?

Blackbaud-MitchellThomas:
My team was caught off guard to see a 2xx response, in our case a 204, throw an exception, especially with a message in the exception of "Success", which is kind of a bizarre thing by itself. I would consider this more of a bug than an enhancement (how it's currently tagged).

We are totally fine with different blocks for each status code even with the same response type. Merging the code blocks is an optimization that I wouldn't hitch to a solution for this, especially with the complexity it would add.

Is anyone working on this?

RSuter:
> Is anyone working on this?

Problaby not

Blackbaud-MitchellThomas:
I'll see if I can help out!

jchannon:
Just hit this by inspecting the generated code. Why is the default behaviour to throw if not 200 or 204. Seems a very odd decision.  I would personally prefer if there was an option such as `ThrowExceptions` that hides this default feature. As a client I don't think I should be expecting an exception, it should be my responsibility to check the status code and process that how I want.

<img width="305" alt="screen shot 2019-02-21 at 20 41 47" src="https://user-images.githubusercontent.com/105126/53202561-c1cf0700-361e-11e9-840f-c33a95e9abb9.png">

kensykora:
The fact that this is tagged enhancement is really bizarre to me. `202 Accepted` is super common in rest patterns for long running requests. 

See https://github.com/microsoft/api-guidelines/blob/vNext/Guidelines.md#132-stepwise-long-running-operations

To me this is a bug with the library. I understand it'd cause backward compatibility issues, but a code generation flag, as was suggested earlier seems like a very reasonable solution. The fact that we need to handle this "normally" with a try/catch is very frustrating to me.

RicoSuter:
> As a client I don't think I should be expecting an exception, it should be my responsibility to check the status code and process that how I want.

The idea is to hide HTTP specific stuff (avoid leaky abstractions).

[...]

PR #1668, opened 2018-10-15
Add AllowAllSuccessCodes-Option to TS client generator.

FabianTe:
This is a (TypeScript-only) fix for issue #1602

Enabling this option (`AllowAllSuccessCodes`, false by default) will cause **every** 2xx status code to be handled as a successful one.

An Example: An operation that returns either a DTO with status code 200 or nothing with status code 204.

Before:

```typescript
if (status === 200) {
	return response.text().then((_responseText) => {
		let result200: any = null;
		result200 = _responseText === "" ? null : <MyDto[]>JSON.parse(_responseText);
		return new SwaggerResponse(status, _headers, result200);
	});
}
if (status === 204) {
	return response.text().then((_responseText) => {
		return throwException("A server error occurred.", status, _responseText, _headers);
	});
}
```

After:

```typescript
// ...
if (status === 200) {
	return response.text().then((_responseText) => {
		let result200: any = null;
		result200 = _responseText === "" ? null : <MyDto[]>JSON.parse(_responseText);
		return new SwaggerResponse(status, _headers, result200);
	});
}
if (status === 204) {
	return response.text().then((_responseText) => {
		return new SwaggerResponse(status, _headers, <any>null);
	});
}
// ...
```

### DIFF
--- a/src/NSwag.CodeGeneration.TypeScript/Models/TypeScriptOperationModel.cs
+++ b/src/NSwag.CodeGeneration.TypeScript/Models/TypeScriptOperationModel.cs
@@ -64,12 +64,37 @@ namespace NSwag.CodeGeneration.TypeScript.Models
         {
             get
             {
-                var response = GetSuccessResponse();
-                var isNullable = response?.IsNullable(_settings.CodeGeneratorSettings.SchemaType) == true;
+                string resultType = "";
+                bool isNullable = false;
 
-                var resultType = isNullable && SupportsStrictNullChecks && UnwrappedResultType != "void" && UnwrappedResultType != "null" ?
-                    UnwrappedResultType + " | null" :
-                    UnwrappedResultType;
+                if (!_settings.AllowAllSuccessCodes)
+                {
+                    var response = GetSuccessResponse();
+                    isNullable = response?.IsNullable(_settings.CodeGeneratorSettings.SchemaType) == true;
+
+                    resultType = UnwrappedResultType;
+
+                } else
+                {
+                    foreach (string _responseStatus in _operation.ActualResponses.Keys)
+                    {
+                        SwaggerResponse _response = _operation.ActualResponses[_responseStatus];
+                        if (_responseStatus.StartsWith("2"))
+                        {
+                            var _isNullable = _response.IsNullable(_settings.CodeGeneratorSettings.SchemaType);
+                            if (_isNullable)
+                                isNullable = true;
+                            var _type = _generator.GetTypeName(_response.ActualResponseSchema, _isNullable, "Response");
+                            resultType += (resultType != "" ? " | " : "") + _type;
+                        }
+                        
+                        if (resultType == "")
+                            resultType = "void";
+                    }
+                }
+
+                if (isNullable && SupportsStrictNullChecks && UnwrappedResultType != "void" && UnwrappedResultType != "null")
+                    resultType += " | null";
 
                 if (WrapResponse)
                 {
--- a/src/NSwag.CodeGeneration.TypeScript/Models/TypeScriptResponseModel.cs
+++ b/src/NSwag.CodeGeneration.TypeScript/Models/TypeScriptResponseModel.cs
@@ -9,6 +9,7 @@
 using NJsonSchema;
 using NJsonSchema.CodeGeneration.TypeScript;
 using NSwag.CodeGeneration.Models;
+using System.Linq;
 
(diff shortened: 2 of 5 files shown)


RSuter:
The problem with that is that it works only "by accident" - i.e. when all 2xx responses have either the same response type or null - I think we should merge all responses with the same type as described in this issue: https://github.com/RSuter/NSwag/issues/1259

FabianTe:
> I think we should merge all responses with the same type as described in this issue: #1259

I'd like to clarify that with this PR I only want to create a fix / workaround for the problem discussed in #1602 In our project we are dealing with controllers that return multiple 2xx status codes in one operation and the generated TS client simply isn't able to deal with these operations, even after manipulating the templates.

Since a comprehensive fix is more complicated, I intend to provide a workaround until the overall problem can be solved another way.

> The problem with that is that it works only "by accident" - i.e. when all 2xx responses have either the same response type or null

I added a commit that deals with this. Now the generator will add the types of all 2xx responses to the operation result type.

Hope this is helps clarifying things 😉

FabianTe:
Have you thought about including this? We are going to depend on this (or any other) fix / workaround, otherwise the TS client simply isn't usable.

bret-bailey:
@RSuter - just ran across this issue - all DELETE api calls return 204 - but the generated code handles this as an exception - it seems that the only 'fix' for us is to force the backend to return 200 despite there being 'no content' - 

is there another solution?  how can I get the generated code to handle both 200 & 204 the same way?  
e.g. if (status == 200 || status == 204) {

`
    protected processDeleteServerUsingDELETE(response: HttpResponseBase): Observable<ResponseEntity> {
        const responseBlob =
            response instanceof HttpResponse ? response.body :
            (<any>response).error instanceof Blob ? (<any>response).error : undefined;

        let _headers: any = {}; if (response.headers) { for (let key of response.headers.keys()) { _headers[key] = response.headers.get(key); }};
        if (status === 200) {
            return blobToText(responseBlob).pipe(_observableMergeMap(_responseText => {
            let result200: any = null;
            let resultData200 = _responseText === "" ? null : JSON.parse(_responseText, this.jsonParseReviver);
            result200 = resultData200 ? ResponseEntity.fromJS(resultData200) : new ResponseEntity();
            return _observableOf(result200);
            }));
        } else if (status === 204) {
            return blobToText(responseBlob).pipe(_observableMergeMap(_responseText => {
            return throwException("A server error occurred.", status, _responseText, _headers);
            }));
        } else if (status === 401) {
            return blobToText(responseBlob).pipe(_observableMergeMap(_responseText => {
            return throwException("A server error occurred.", status, _responseText, _headers);
            }));
        } else if (status === 403) {
            return blobToText(responseBlob).pipe(_observableMergeMap(_responseText => {
            return throwException("A server error occurred.", status, _responseText, _headers);
            }));
            return blobToText(responseBlob).pipe(_observableMergeMap(_responseText => {
            }));
        }
        return _observableOf<ResponseEntity>(<any>null);
    }
`

related swagger doc json:
`
      "delete": {
        "tags": [
          "server-controller"
        ],
        "summary": "deleteServer",
        "operationId": "deleteServerUsingDELETE",
        "produces": [
          "application/json"
        ],
        "parameters": [
          {
            "name": "serverId",
            "in": "path",
            "description": "serverId",
            "required": true,
            "type": "integer",
            "format": "int64"
          },
          {
            "name": "name",
            "in": "query",
            "required": false,
            "type": "string"
          }
        ],
        "responses": {
          "200": {
            "description": "OK",
            "schema": {
              "$ref": "#/definitions/ResponseEntity"
            }
          },
          "204": {
            "description": "No Content"
          },
          "401": {
            "description": "Unauthorized"
          },
          "403": {
            "description": "Forbidden"
          }
        },
        "security": [
          {
            "JWT Bearer Token": [
              "all"
            ]
          }
        ],
        "deprecated": false
      }
    },
`

bobwah:
Any updates on this?

ecolara3:
**BUMP for attention!**

RicoSuter:
Can someone resolve the conflicts so that we can look into this?

RicoSuter:
I think this should be handled via the 2xx merge config (not yet implemented). 
/cc @jeremyVignelles

FabianTe:
Please excuse my lack of activity. I don't work at the company this project anymore. I'll close because I will not work on this PR anymore.

PR #1688, opened 2018-10-26
Quick fix: Correct typo in `NSwag.Sample.NETCore20.Startup`

dougbu:
- change /openapi/v1/openapi.json to host the `"openapi"` document


RSuter:
Thanks for the fix

Issue #1355, opened 2018-05-30
Epic: Improve API version support


RSuter:

commonsensesoftware:
I just stumbled across a number of posts and issues that all seem to lead back to here.

The implementation of [ApiVersionProcessor.cs](../blob/master/src/NSwag.SwaggerGeneration/Processors/ApiVersionProcessor.cs) is not correct and will miss many scenarios. [API Versioning](https://github.com/Microsoft/aspnet-api-versioning) does not actually care about attributes and that is only one way that API versions can be provided. API versions can also be provided by conventions or through custom extensions.

The [API Explorer](https://www.nuget.org/packages/Microsoft.AspNetCore.Mvc.Versioning.ApiExplorer) extensions provided by API Versioning already does all of the difficult work of collating API descriptions by API version so they can be integrated with other libraries such as Swagger generators like NSwag. Things already work well with Swashbuckle without any cross-dependencies or understanding of what each library does. I would be curious and interested in understanding why the same model doesn't work for NSwag. I'm open to discussing how the integration can be made better.

commonsensesoftware:
If a service author is versioning by URL segment **and** they don't want the route parameter or API parameter desciption included, they just need to specify that in the API explorer options a la:

```c#
AddVersionedApiExplorer( options => options.SubstituteApiVersionInUrl = true );
```

This removes all API version parameters and substitutes the route parameter with a value formatted according to the `options.SubstitutionFormat` (defaults to `"VVV"`, which is the most common). If the URL segment method is not used, these options have no effect.

It's also worth noting that the **apiVersion** route constraint key is configurable starting in API Versioning 3.0+ via `options.RouteConstraintName`, even though it's not likely to be changed.  The name of route parameter is completely user-defined. While it's common and used in many examples, there is no requirement that the route parameter name needs to be **version**. The route parameter name is only meaningful to the service author. ASP.NET Core does the work of parsing the route template, matching the route constraint, and passing the name and value of the parameter to the constraint for processing. In other words, a route parameter defined as `{v:apiVersion}` or `{ver:apiVersion}` is perfectly valid. The ASP.NET API Explorer strips out the route constraint definition, so the generated path is left with `{v}` and `{ver}` respectively. You _should_ be able to make this work if you can determine which API parameter description matches the API version parameter and use it's name to replace the token. While it's possible, it seems like it's something difficult and brittle for NSWag to determine and would be better served by making the API Versioning API Explorer support do the work instead (which it does).

[...]

PR #1701, opened 2018-10-31
Add ApiGroupNames setting, remove ApiVersionProcessor and improve api versioning

RSuter:
- [x] Add ApiGroupNames setting
- [x] Remove ApiVersionProcessor (in AspNetCoreToSwaggerGeneratorSettings)
- [x] Remove IncludedVersions setting on aspnetcore2swagger / nswag.json (CLI)

Should solve lots of api versioning problems, see https://github.com/RSuter/NSwag/issues/1355

**This is a breaking change for users of the old ApiVersionProcessor:** Use AddVersionedApiExplorer() in ConfigureServices() and define ApiGroupNames instead of IncludedVersions

```
services.AddApiVersioning(options =>
{
    options.AssumeDefaultVersionWhenUnspecified = true;
    options.ApiVersionReader = new UrlSegmentApiVersionReader();
})
.AddMvcCore()
.AddVersionedApiExplorer(options =>
{
    options.GroupNameFormat = "VVV";
    options.SubstituteApiVersionInUrl = true;
});
```

Without UseDocumentProvider: 

image

With UseDocumentProvider: 

image


RSuter:
@commonsensesoftware now one of the problems i have is that even if i set `SubstituteApiVersionInUrl = true;` the parameter `api-version` is still declared. How can I disable the reporting of this parameter? It doesnt make sense to report it because all placeholders have been removed...

RSuter:
@commonsensesoftware thanks for the info. PR looks fine now...

OculiViridi:
@RSuter I can't find the `AddVersionedApiExplorer()` method.

RSuter:
It's in the Microsoft.AspNetCore.Mvc.Versioning.ApiExplorer package

See sample project: 
https://github.com/RSuter/NSwag/blob/master/src/NSwag.SwaggerGeneration.AspNetCore.Tests.Web/NSwag.SwaggerGeneration.AspNetCore.Tests.Web.csproj

Furynation:
> 
> 
> @RSuter I can't find the `AddVersionedApiExplorer()` method.

@OculiViridi Make sure you are adding MvcCore directly before calling `AddVersionedApiExplorer()`:
`services.AddMvcCore();`
`services.AddVersionedApiExplorer();`

If you need the full MVC package, then:
`services.AddMvc()..SetCompatibilityVersion(CompatibilityVersion.Version_2_1);
services.AddMvcCore().AddVersionedApiExplorer();`

acds:
Hi All, Per https://github.com/RicoSuter/NSwag/issues/2325

I'm still having issues with 2.2 and API versioning. 

When trying to use `services.AddMvcCore().AddVersionedApiExplorer();` the `AddVersionedApiExplorer` does not resolve.

I was hoping migration to NSwag would be easy...

commonsensesoftware:
You shouldn't have to use `AddMvcCore` anymore. This was improved in API Versioning quite a long while ago. Calling `services.AddVersionedApiExplorer()` will implicitly call the dependent `AddApiExplorer()` extension.

In attempting to stitch the two issues together, it would appear that your group names are incongruent. That would explain why things do not match up. Swagger generators and API Versioning do **not** know about each other. API Versioning collates and groups actions according to their API versions. Swagger generators simply use the grouped actions defined by the API Explorer - however that magically happened.

I see that your document has the **Name** and **Version** of `v1.0`. The configured group name format; however is `VVV`. This format code will yield `major[.minor][-status]`. This would yield an API version of `1.0` as simply `1`.  Note how the `v` is **not** included. The `v` is not (nor ever is) part of the API version. Some people like having the `v` prefix, but you need to include it in the format; for example, `'v'VVV`. This particular built-in format code uses an optional minor version. The numeric, minor version is considered optional if it equals `0`. This would yield a result of `v1` in your example. To match the format in your Swagger document, use `'v'VV` which will format `1.0` as `v1.0`.

For more information, review [API Version Formatting](https://github.com/Microsoft/aspnet-api-versioning/wiki/Version-Format). I hope that helps.

acds:
@commonsensesoftware Thanks for the reply. I read the docs as linked.

Yes, in my eagerness to migrate to NSwag, I'd not correctly decoded, the other implementation.

with
```C#

 services.AddVersionedApiExplorer(options =>
 {
     options.GroupNameFormat = "VVV";
     options.SubstituteApiVersionInUrl = true;
});

...

services.AddApiVersioning(options =>
{
    options.ReportApiVersions = true;
    options.AssumeDefaultVersionWhenUnspecified = true;
    options.DefaultApiVersion = new ApiVersion(1, 0);
});
```
Swashbuckles:
```C#
services.AddSwaggerGen(c =>
{
    c.SwaggerDoc("v1", new Info { Title = "My Title API", Version = "v1" });
});
```
translates to:
```C#
services.AddOpenApiDocument(document =>
{
    document.DocumentName = "v1";
    document.Version = "v1";
    document.Title = "My Title API";
    document.ApiGroupNames = new[] {"1"};
});    
```
In the hopes that this help other migrators.

As a though... it might be a good idea to retrieve the `IApiDescriptionGroupCollectionProvider` via:
```C#
var explorer = services.BuildServiceProvider().GetRequiredService<IApiDescriptionGroupCollectionProvider>();
```
I'd figure that a useful model would be to loop though each group identified in the explore and add a `services.AddOpenApiDocument` for each group identified such that it was always a complete set and this code needs no further configuration as the API evolves... 

However it seems the explorer at the point provides an empty collection, and figure this is because the startup has not proceeded far enough in the build process as yet to have the API Version information generated from the MVC code...where the Swashbuckle implementation would otherwise do this in the `UseSwagger()` or `UseSwaggerUi3()` in this case...

commonsensesoftware:
@acds, the **IApiVersionDescriptionProvider** service defined by API Versioning will give you a simpler and fully aggregated view of the information you're looking for. The Swashbuckle setup used to be similar to the same setup you are describing here using `services.BuildServiceProvider()`. Swashbuckle ended up adding _Options_ support, which enabled configuring generation later in the pipeline. You can see how these two concepts come together in [this example](https://github.com/microsoft/aspnet-api-versioning/blob/master/samples/aspnetcore/SwaggerSample/ConfigureSwaggerOptions.cs).

I don't know enough about NSwag to tell you if this same concept is supported for configuring `services.AddOpenApiDocument`. If it's not, then that's the approach for a feature I would consider requesting.

I hope that's useful.

ThomasVague:
Is there any updates on iterating through the IApiVersionDescriptionProvider? I've been looking for a more elegant solution, but this seems to be the only way still?

`var explorer = services.BuildServiceProvider().GetRequiredService<IApiDescriptionGroupCollectionProvider>();`

commonsensesoftware:
@ThomasVague Are you looking for something beyond `IApiDescriptionGroupCollectionProvider.ApiVersionDescriptions`? That will enumerate all available `ApiVersionDescription` instances in order. If NSwag were to expose this information directly, I presume it should need to provide some type of adapter to be bridge it in a generic way; otherwise, you can also get this information from API Versioning from the DI container at various points in the application.

commonsensesoftware:
I can't speak to the speak hooks for NSwag, but it should certainly possible to hook up the API Versioning metadata in a similar way as Swashbuckle. There are no special handling or conditions for specific OpenAPI generators, including Swashbuckle. If there are some community examples of NSwag with API Versioning, I'm willing to link to them in the documentation.

Unfortunately, you can't really group further; at least, not without some serious work. The issue isn't the code, NSwag, or API Versioning. The OpenAPI (formerly Swagger) UI only affords for a single level of grouping. Numerous changes would have to be made in order to make it work - likely more than you're willing to do. Another alternative is that you can combine the components together as a single string. For example:

- `v1/Namespace1`
- `v1/Namespace2`
- `v2/Namespace1`
- `v2/Namespace2`
- `v2/Namespace3`

That's probably not what you're hoping for, but that will work with minimal effort. Yet one more option is to create your own UI and then you can group however you like.

ThomasVague:
That's actually exactly what I'm looking for. Only a single level, but grouped in versions and namespaces. I've been looking around, but haven't found any examples to do grouping in that manner. 

Currently my code looks like this : 

```c#
services.AddVersionedApiExplorer(o =>
	{
		o.SubstituteApiVersionInUrl = true;
		o.ApiVersionParameterSource = new UrlSegmentApiVersionReader();
		o.GroupNameFormat = "'v'VVV";
	});

	var versionDescriptionProvider = services.BuildServiceProvider().GetService<IApiVersionDescriptionProvider>();

	foreach (var apiVersionDescription in versionDescriptionProvider?.ApiVersionDescriptions)
	{
		services.AddOpenApiDocument((document, serviceProvider) =>
		{
			document.DocumentName = "v" + apiVersionDescription.ApiVersion.MajorVersion;
			document.ApiGroupNames = new string[] { "v" + apiVersionDescription.ApiVersion.MajorVersion };
			document.Version = apiVersionDescription.ApiVersion.MajorVersion + "." + apiVersionDescription.ApiVersion.MinorVersion;
			document.Description = settings.Description;

			if (apiVersionDescription.IsDeprecated)
			{
				document.Description += " This API version has been deprecated.";
			}

			document.PostProcess = doc =>
			{
				doc.Info.Contact = new()
				{
					Name = settings.ContactName,
					Email = settings.ContactEmail,
					Url = settings.ContactUrl
				};
				doc.Info.License = new()
				{
					Name = settings.LicenseName,
					Url = settings.LicenseUrl
				};
			};


			document.AddSecurity(JwtBearerDefaults.AuthenticationScheme, new OpenApiSecurityScheme
			{
				Name = "Authorization",
				Description = "Input your Bearer token to access this API",
				In = OpenApiSecurityApiKeyLocation.Header,
				Type = OpenApiSecuritySchemeType.Http,
				Scheme = JwtBearerDefaults.AuthenticationScheme,
				BearerFormat = "JWT",
			});

			document.OperationProcessors.Add(new AspNetCoreOperationSecurityScopeProcessor());
			document.OperationProcessors.Add(new SwaggerGlobalAuthProcessor());

			document.TypeMappers.Add(new PrimitiveTypeMapper(typeof(TimeSpan), schema =>
			{
				schema.Type = NJsonSchema.JsonObjectType.String;
				schema.IsNullableRaw = true;
				schema.Pattern = @"^([0-9]{1}|(?:0[0-9]|1[0-9]|2[0-3])+):([0-5]?[0-9])(?::([0-5]?[0-9])(?:.(\d{1,9}))?)?$";
				schema.Example = "02:00:00";
			}));

			document.OperationProcessors.Add(new SwaggerHeaderAttributeProcessor());

			var fluentValidationSchemaProcessor = serviceProvider.CreateScope().ServiceProvider.GetService<FluentValidationSchemaProcessor>();
			document.SchemaProcessors.Add(fluentValidationSchemaProcessor);
		});
	}
```
```c#
app.UseOpenApi(settings =>
	{
	});
	app.UseSwaggerUi3(settings =>
	{
		settings.DefaultModelsExpandDepth = -1;
		settings.DocExpansion = "none";
		settings.TagsSorter = "alpha";
	});
```

commonsensesoftware:
>**Sidebar:** You don't have to write a bunch of code to produce various API formats. `ApiVersion` intrinsically supports formatting for this exact reason (see [version formatting](https://github.com/dotnet/aspnet-api-versioning/wiki/Version-Format)).

```c#
document.DocumentName = apiVersionDescription.ApiVersion.ToString("'v'V");
document.ApiGroupNames = new[] { document.DocumentName };
document.Version = apiVersionDescription.ApiVersion.ToString("VV");
```

### Custom Grouping
Repeating it here would be verbose, but there are a couple of options for custom grouping to achieve the results you want. The simplest way is likely using your own `IApiDescriptionProvider` that [re-]groups API descriptions the way you want. Here's a few of the links that explain it in more detail and with code:

- [Swagger: I need versioning AND grouping](https://stackoverflow.com/questions/68788850/swagger-i-need-versioning-and-grouping)
- [How to support multiple swagger files per version/group?](https://github.com/dotnet/aspnet-api-versioning/issues/516)

Issue #1704, opened 2018-11-01
swagger2tsclient: Functions throwException and blobToText generated with out SwaggerException


FreeFrags:
When i select a few DTO objects to be generated, with out an API the Functions throwException and blobToText are generated but the SwaggerException is not.

I think these functions are not needed either if you dont generate an API.

I propose to disable the generation of these 2 functions when no api is generated.

I attached a nswag config and a swaggerSpec to the issue so you can reproduce the error.
[config.nswag.txt](https://github.com/RSuter/NSwag/files/2538032/config.nswag.txt)
[SwaggerSpec.json.txt](https://github.com/RSuter/NSwag/files/2538033/SwaggerSpec.json.txt)

PR #1705, opened 2018-11-01
Only generate the blobToText and throwException if it is needed

FreeFrags:
A proposal to fixing the following issue:
https://github.com/RSuter/NSwag/issues/1704

### DIFF
--- a/src/NSwag.CodeGeneration.TypeScript/Models/TypeScriptFileTemplateModel.cs
+++ b/src/NSwag.CodeGeneration.TypeScript/Models/TypeScriptFileTemplateModel.cs
@@ -132,9 +132,14 @@ namespace NSwag.CodeGeneration.TypeScript.Models
             !_settings.TypeScriptGeneratorSettings.ExcludedTypeNames.Contains("FileResponse") &&
             _document.Operations.Any(o => o.Operation.ActualResponses.Any(r => r.Value.Schema?.ActualSchema.Type == JsonObjectType.File));
 
-        /// <summary>Gets a value indicating whether the SwaggerException class is required.</summary>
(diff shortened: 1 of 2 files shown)


RSuter:
I think we need a new property, eg RequireClientFunctions without the swaggerexception check for the functions:

https://github.com/RSuter/NSwag/blob/b5a2a2d7195019b846e31b0d36c7058abbab9a1d/src/NSwag.CodeGeneration.TypeScript/Models/TypeScriptFileTemplateModel.cs#L136

FreeFrags:
i might be mistaken but since the client functions require the SwaggerException class,i think that the RequiresClientFunctions function needs to check if RequiresSwaggerExceptionClass returns true. 

Which will then result in the same check as the RequiresSwaggerExceptionClass. 

I will modify the code so you can see what i mean.

FreeFrags:
Would this be the preferred solution?

RSuter:
Yes, that makes sense

FreeFrags:
All right thank you very much, is there anywhere we can see what the estimated date or the open issues remaing till a next release are?

RSuter:
I am currently merging some PRs, test everything and will release a new version in the next days.

RSuter:
I think it should be

```
        /// <summary>Gets a value indicating whether the client functions are required.</summary>
        public bool RequiresClientFunctions => 
            _settings.GenerateClientClasses &&
            !string.IsNullOrEmpty(Clients);

        /// <summary>Gets a value indicating whether the SwaggerException class is required. Note that if RequiresClientFunctions returns true this returns true since the client functions require it. </summary>
        public bool RequiresSwaggerExceptionClass =>
            RequiresClientFunctions &&
            !_settings.TypeScriptGeneratorSettings.ExcludedTypeNames.Contains("SwaggerException");
```

template looks correct.

Issue #1707, opened 2018-11-01
Typo in Swagger configuration parameter: tagSorter -> tagsSorter


WesleySSmith:
There is a typo in the name of a swagger configuration parameter that is causing it to have no effect.  In _src\NSwag.AspNetCore\SwaggerUi3\index.html_ and _src\NSwag.AspNet.Owin\SwaggerUi3\index.html_, there is this:

`    tagSorter: "{TagSorter}",`

it should be:

`    tagsSorter: "{TagSorter}",`

This config setting is documented as "tagsSorter" here:  https://github.com/swagger-api/swagger-ui/blob/master/docs/usage/configuration.md

I don't know if you'd also like to fix the other instances where TagSorter should be TagsSorter, for example in the property of SwaggerUi3Settings.

On a related note, NSwag seems to also still support the "apisSorter" configuration option, which is no longer in use, at least according to this swagger-ui commit:

https://git.r3ktm8.de/root/swagger-ui/commit/571d65a5829e5b4d352fe126f251e045a1cef16f

RSuter:
Can you create a pr where you remove the apisSorter, and rename the property + fix the bug? I think the break is fine as it gives instant, easy to fix compile errors and there are breaking changes anyway in the next version

PR #1708, opened 2018-11-01
Rename TagSorter -> TagsSorter to match what swagger-ui needs.  #1707

WesleySSmith:
Proposed fix for #1707


RSuter:
Looks good, thanks for the fix.

PR #1728, opened 2018-11-09
Add TransformToExternalPath setting 

RSuter:
- Add TransformToExternalPath setting

Transform the internal route to an external route based on the request headers from nginx:

```
app.UseSwaggerUi3(config => config.TransformToExternalPath = (internalUiRoute, request) => 
{ 
    // The header X-External-Path is set in the nginx.conf file 
    var externalPath = request.Headers.ContainsKey("X-External-Path") ? request.Headers["X-External-Path"].First() : "";  
    return externalPath + internalUiRoute; 
}); 
```

**Breaking changes:**

- Rename SwaggerRoute/SwaggerUiRoute to DocumentPath/Path


TwoAngel:
Soo, u removed "UseSwaggerUi3WithApiExplorer" and instead??

RicoSuter:
AddSwaggerDocument()
UseOpenApi()
UseSwaggerUi3()

PR #1752, opened 2018-11-19
[WIP] VersionedApiExplorer example

LuukN2:
A W.I.P. generator for WebApi and WebApi with OData using the VersionedApiExplorer.

No generator is necessary for AspNetCore because versioning replaces the interfaces used by the Mvc Apiexplorer.


LuukN2:
Awesome, I'll finish the processors and make a test project. I will try to help out at versioning to get the ODataQueryableAttributes supported so an odata specific processor will not be necessary here.

Should I move over the files from NSwag.AspNet.Owin over to the new project since it keeps a reference to NSwag.SwaggerGeneration.WebApi?

RSuter:
> Should I move over the files from NSwag.AspNet.Owin over to the new project since it keeps a reference to NSwag.SwaggerGeneration.WebApi?

No, that's fine. The new projects can use existing ones...

LuukN2:
I added some tests and it seems to work in my production environment. I made no changes to any of the existing projects.

RicoSuter:
Whats the state of this pr?

RicoSuter:
TBH, is it really work investing more time in the legacy ASP.NET MVC support (pre .NET Core)? There the generator is reflection based and bad anyway...

LuukN2:
Currently this will support Web Api and odata using the versioned api explorers, I don't know how many people would need support for the old Mvc.

Supporting OData and Web Api on .NET framework with the latest packages is important for us. I'm not quite sure how many other have this requirement, especially with old mvc which doesn't have an api explorer.

I'd say it should be fine to support only Web Api and OData since the api explorers should make it relatively low effort to maintain when things update.

RicoSuter:
Ok, yes - it's a completely new library and should not interfere something else... there will be some breaking changes soon (v13): https://github.com/RicoSuter/NSwag/pull/2178

We might need to fix some things in the PR then...

RicoSuter:
Will not continue to support legacy technologies (full .net, asp.net etc)

Issue #1653, opened 2018-10-05
TypeScript client: DTO fields are not initialized


ili:
Hi!
After updating to current version toolchain (from  v11.12.9.0) we'v faced with breaking change - now DTO peoperties are not initialized by default:

Old ( v11.12.9.0):
```js
export class SelectRequestBaseOfInstallationFilter implements ISelectRequestBaseOfInstallationFilter {
   filter: InstallationFilter = new InstallationFilter(); // INITED
   page: Paginator = new Paginator(); //INITED

    // skipped constructor & etc..
}
```

Current (v11.20.1.0):
```js
export abstract class SelectRequestBaseOfInstallationFilter implements ISelectRequestBaseOfInstallationFilter {
    filter: InstallationFilter; // NOT INITED
    page: Paginator; // NOT INITED

    // skipped constructor & etc..
}
```

RSuter:
Thats probably because the spec has changed. Can you post the old and new spec? Null handling has been improved for openapi 3 output...

RSuter:
I get this typescript which looks fine for me (with TS 2.7 settings): 

- The properties are required (!)
- The properties are initialized in the ctor when no data is provided

![image](https://user-images.githubusercontent.com/2603405/46537604-fdf10d80-c8b1-11e8-9ad1-5c952801452d.png)

ili:
Up )
Teammates are angry with me, the point is only one to migrate to the next Angular version :)

Please help :)

RSuter:
> In some reason filter property is not initialized.

Can you post the declaration of the filter property and filter schema of your spec or the whole spec?

ili:
Here it is [swagger.txt](https://github.com/RSuter/NSwag/files/2541184/swagger.txt)

RSuter:
Looks good to me: 

![image](https://user-images.githubusercontent.com/2603405/47903022-66b4b100-de83-11e8-9647-78281be9de56.png)

You can run my config: 

[nswag.zip](https://github.com/RSuter/NSwag/files/2541667/nswag.zip)

ili:
Shame on me!

I'v gave you wrong swagger definition! 

Here is one, where the problem is reproduced (this is to the same API but generated with newer NSwag version).
[swagger_new.txt](https://github.com/RSuter/NSwag/files/2551613/swagger_new.txt)

ili:
[...]

PR #1768, opened 2018-11-26
Test for #1653

ili:
Test demonstrates that properties are not initialized:
```ts
class GenericRequestBaseOfRequestBody implements IGenericRequestBaseOfRequestBody {
    request!: RequestBody;

    constructor(data?: IGenericRequestBaseOfRequestBody) {
        if (data) {
            for (var property in data) {
                if (data.hasOwnProperty(property))
                    (<any>this)[property] = (<any>data)[property];
            }
        } 
        // here should be this.request initialization
    }

    init(data?: any) {
        if (data) {
            this.request = data["Request"] ? RequestBody.fromJS(data["Request"]) : new RequestBody();
        }
    }

    static fromJS(data: any): GenericRequestBaseOfRequestBody {
        data = typeof data === 'object' ? data : {};
        let result = new GenericRequestBaseOfRequestBody();
        result.init(data);
        return result;
    }

    toJSON(data?: any) {
        data = typeof data === 'object' ? data : {};
        data["Request"] = this.request ? this.request.toJSON() : <any>undefined;
        return data; 
    }
}
```


src/.editorconfig
```diff
@@ -0,0 +1,10 @@
+; What is EditorConfig? http://editorconfig.org/
```
RSuter: Remove and .gitignore this file

ili: this is great thing :)
it sets editor properties, for ex. i do use tabs for spacing, and it is configured in my VS, this file allows solution wide configuration without need to reconfigure VS properties

ili: Done

ili:
I'v checked with current master branch, test succeed!

Thanks a lot!

RSuter:
@ili so this is already fixed?

ili:
@RSuter Yes! )

PR #1812, opened 2018-12-11
Fix for .net core 2.2 support in npm package

MarcosMeli:
Adding .net core 2.2 as default runtime
Adding 22 to supportedCoreVersions array

### DIFF
--- a/src/NSwag.Npm/bin/nswag.js
+++ b/src/NSwag.Npm/bin/nswag.js
@@ -1,8 +1,8 @@
 #!/usr/bin/env node
 "use strict";
 
-var defaultCoreVersion = "21";
-var supportedCoreVersions = ["10", "11", "20", "21"];
+var defaultCoreVersion = "22";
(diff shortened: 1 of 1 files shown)


RSuter:
Had to change the default back to 2.1 as it would be a breaking change...

MarcosMeli:
Perfect @RSuter. I thought that .net core 2.2 runtime will handle 2.1 assemblies with no problem
Good to know for future releases
Thanks

RSuter:
> I thought that .net core 2.2 runtime will handle 2.1 assemblies with no problem

Maybe yes, but you need to have 2.2 installed on the machine (which might not be the case). Also i dont want to take this risk :-)

MarcosMeli:
> > I thought that .net core 2.2 runtime will handle 2.1 assemblies with no problem
> 
> Maybe yes, but you need to have 2.2 installed on the machine (which might not be the case). Also i dont want to take this risk :-)

I completely agree with you

PR #1872, opened 2019-01-08
Path param with model binder attribute

Tibo-lg:
When using a `ModelBinderAttribute` on a path parameter, the api explorer sets the `BindingSource` to `Custom` which leads to it appearing as a query parameter in the swagger documentation, even though it is properly treated as a path parameter by asp. A workaround is to also add the `[FromRoute]` tag, but this PR adds a small check to detect when a parameter has a custom binding source but is actually coming from the path to avoid having to use this workaround.

I also added a test for it to highlight the issue.


RSuter:
Thank you for the PR!

PR #1882, opened 2019-01-13
Controlling generation of CancellationToken in interface

Librazy:
Added CancellationTokenParameterOption with has 3 vaild value:

1. Both.
  This is the default option and has same behavior as before. Generate both method in interface. If GenerateOptionalParameters, generate single method with CancellationToken being optional.
2. Yes.
  Generate the overload with CancellationToken only.
2. No.
  Generate the overload without CancellationToken only. This option force it to generate a implementation method without CancellationToken in class.


src/NSwagStudio/ViewModels/CodeGenerators/SwaggerToCSharpClientGeneratorViewModel.cs
```diff
@@ -39,6 +40,11 @@ public SwaggerToCSharpClientCommand Command
             .Select(t => (CSharpClassStyle)Enum.Parse(typeof(CSharpClassStyle), t))
             .ToArray();
 
+        /// <summary>Gets the list of cancellation token options. </summary>
+        public CancellationTokenParameterOption[] CancellationTokenParameterOptions { get; } = Enum.GetNames(typeof(CancellationTokenParameterOption))
```
mihailpd: `{ get; } = ...` is C# 6.0 feature

RSuter:
Thanks for the PR. I still need to find time to review it... hopefully soon.

brechtvhb:
Hello Rico,

Did you already find the time to review the PR?  I need this feature too.

RicoSuter:
It makes sense to add this option, but i dont like the yes/no/both enum - but i dont know yet a better solution/api. Any idea?

Librazy:
The other way is divide this into two bool argument but one combination don't make sense, so I still prefer the Y/N/B enum. Actually it's a nullable `bool?` before but I found that will reduce readability so I extracted an 3-option enum.

brechtvhb:
A boolean is sufficient for me; Either ignore the cancellation token or not. I can't see a use case to render the function twice once with and once without cancellationToken.

RicoSuter:
> I can't see a use case to render the function twice once with and once without cancellationToken.

Isnt this the current behavior? Or is it with ct and with default value?

RicoSuter:
Otherwise I’d probably agree that a setting like GenerateCancellationTokens would be fine - but need to check the scenarios first

melucas:
Could this also be added to the CSharp Client?

RyanMarcotte:
I see this hasn't been looked at in quite a while and has conflicts with the main branch now...  Was looking for a way to remove the "no `CancellationToken`" methods from generated clients and this would do the trick.  Any chance the merge conflicts could be fixed and this feature merged in?

RicoSuter:
I think instead of adding more configs, we should just add the cancellation tokens with default value (`CancellationToken cancellationToken = default(CancellationToken)`) in the templates and all is fine, no?

RyanMarcotte:
Thinking about it more, even with a mandatory `CancellationToken` parameter developers could still just pass `CancellationToken.None` into methods anyway.  Seems like code reviews are the only way to force a "don't use `CancellationToken.None` convention.

> I think instead of adding more configs, we should just add the cancellation tokens with default value (`CancellationToken cancellationToken = default(CancellationToken)`) in the templates and all is fine, no?

This is sufficient.

RicoSuter:
You can generate both (with and without ct) so this can be closed.

PR #1928, opened 2019-01-28
Generate Authorization attributes in controller.

alex-leroux:
Hi,
I need Authorization attributes to be generated in the Controller.liquid template, if specified in the swagger definition. This pull request indicates a way that this might work. It only caters for operation level security scheme mappings, in ASP Net Core, for api key and basic schemes types.

A complete solution would probably require support for ASP Net, for controller level security mappings, and for OAuth2 and OpenID Connect schemes types, and also be guarded by an opt-in setting.

Can it work this way, or do you see this developing in some other direction, in NSwag.CodeGeneration.CSharp?

### DIFF
--- a/src/NSwag.CodeGeneration.CSharp.Tests/Controllers/ControllerGenerationDefaultParameterTests.cs
+++ b/src/NSwag.CodeGeneration.CSharp.Tests/Controllers/ControllerGenerationDefaultParameterTests.cs
@@ -1,4 +1,7 @@
-﻿using System.Text.RegularExpressions;
+﻿using System;
+using System.Collections.Generic;
+using System.Collections.ObjectModel;
+using System.Text.RegularExpressions;
 using NJsonSchema;
 using Xunit;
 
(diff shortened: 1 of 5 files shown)


alex-leroux:
I have now updated this pull request to include a setting which controls generation of Authorization attributes, as well as generating Authorization attributes for security mapped to the API root level.

RSuter:
Why did you close?

alex-leroux:
I no longer need these changes, to continue with my use of the package.
I have done this by extending the controller generator class, though it is messier for me because I have to duplicate your templates.
The changes in my pull request may however be of some use. I can recreate the pull request if you want to look at it?

RSuter:
I think it would make sense to have this behind a setting (disabled by default).

alex-leroux:
Hi Rico, I have reopened this pull request.
Please take a look and see if it would be helpful for you to pull these changes.

RicoSuter:
On hold until someone requests this

pwfcurry:
Can this be moved into OperationModelBase.cs, so it's accessable to the typescript templates?

I would like to add docs related to a route's security to the generated clients.

GeirGrusom:
May I request this? :) I need authorization attributes based on whether or not security is enabled on the end points.

PR #1933, opened 2019-01-30
allowAnonymous calls in OperationSecurityScopeProcessor

Opvolger:
Thanks for NSwag. We have added this addition to the code because we also have anonymous calls.

### DIFF
--- a/src/NSwag.SwaggerGeneration/Processors/Security/OperationSecurityScopeProcessor.cs
+++ b/src/NSwag.SwaggerGeneration/Processors/Security/OperationSecurityScopeProcessor.cs
@@ -6,11 +6,11 @@
 // <author>Rico Suter, mail@rsuter.com</author>
 //-----------------------------------------------------------------------
 
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
 using System.Threading.Tasks;
-using NSwag.SwaggerGeneration.Processors;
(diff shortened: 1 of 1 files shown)


src/NSwag.SwaggerGeneration/Processors/Security/OperationSecurityScopeProcessor.cs
```diff
@@ -68,5 +73,14 @@ protected virtual IEnumerable<string> GetScopes(SwaggerOperationDescription oper
                 .SelectMany(a => ((string)a.Roles).Split(','))
                 .Distinct();
         }
+
+        /// <summary>Gets all customAttributes from a method.</summary>
+        /// <param name="methodInfo">The method information.</param>
+        /// <returns>The Attributes of the method.</returns>
+        private IEnumerable<Attribute> GetCustomAttributes(MethodInfo methodInfo)
+        {
+            return methodInfo.GetCustomAttributes().Concat(
```
RSuter: Maybe use GetCustomAttributes(true) (true => inherited attributes) instead of concat?

Opvolger: I will try it, a busy day today. I'll come back to this.
I had copied this piece of code from your (line 58/59)

KirschbaumP:
Is there a way to have routes without authentification?

Issue #1547, opened 2018-08-22
SwaggerToCSharpControllerGenerator: Generate Task<ActionResult<T>> for ASP.NET Core


altso:
Please add an ability to generate controller actions that return `ActionResult<T>`/`Task<ActionResult<T>>` for ASP.NET Core.

Ref: #1386

RSuter:
What’s exactly the benefit of that?

altso:
I believe it's the same as with `System.Net.Http.HttpResponseMessage` for ASP.NET WebAPI which is allowed right now (https://github.com/RSuter/NSwag/blob/master/src/NSwag.CodeGeneration.CSharp/Templates/Controller.liquid#L94).

It gives you a better flexibility around http status codes.

```csharp
[HttpPut("resource/{id}")]
public async Task<ActionResult<Resource>> UpdateResource(int id, [FromBody] Resource resource)
{
    if (/* resource exists */)
    {
        if (/* custom validation (e.g. check for fields that could not be updated) */)
        {
            if (/* user is permitted to update this specific resource */)
            {
                // update resource and return an updated copy (e.g. last modified date field)
                var updatedResource = _resourceService.Update(resource);
                return Ok(updatedResource);
            }

            return Forbid();
        }

        return BadRequest();
    }

    return NotFound();
}
```

optuuttila:
I actually played around this a bit and came up with this fork:
https://github.com/optuuttila/NSwag/commit/cef9ecc1693bde731a636c52173562a832e3ef6a

It results in following kind of generated code:
Interface:
        System.Threading.Tasks.Task<Microsoft.AspNetCore.Mvc.ActionResult<MyModel>> MyMethodAsync(string param1, string param2);

Controller:
        public System.Threading.Tasks.Task<Microsoft.AspNetCore.Mvc.ActionResult<MyModel>> MyMethod(string param1, string param2)
        {
            return _implementation.MyMethodAsync(param1, param2);
        }

Obviously the code would need to be finalized (a PR for the NJsonSchema.CodeGeneration.CSharp package), but before doing any of that I just wanted to ask whether this is anything along the lines you would accept as an PR?

RSuter:
Why in NJS? Just create a PR with WIP in the name and ill look into it, you can still improve it etc.

optuuttila:
Ok, I'm very new to contributing to open source projects and I have no idea what NJS means, but I'll start creating the PR's. Thanks.

RSuter:
[...]

PR #1942, opened 2019-02-06
Support to ASP.Net Core ActionResult Type on controller generator

optuuttila:
This PR adds support to ASP.Net Core ActionResult Type on controller generator.
This is related to issue https://github.com/RSuter/NSwag/issues/1547

~This also has dependency to pending PR at https://github.com/RSuter/NJsonSchema/pull/895 which adds a new setting to the code generator settings class for this feature.~


src/NSwag.CodeGeneration.CSharp/Models/CSharpOperationModel.cs
```diff
@@ -124,6 +130,12 @@ public override string ResultType
         {
             get
             {
+                if (UseActionResultType) {
+                    return SyncResultType == "void"
+                        ? "IActionResult"
```
RSuter: Please use full type name

src/NSwag.CodeGeneration.CSharp/Models/CSharpOperationModel.cs
```diff
@@ -70,6 +70,9 @@ public class CSharpOperationModel : OperationModelBase<CSharpParameterModel, CSh
                 }
             }
 
+            //TODO: FORK the NJsonSchema.CodeGeneration.CSharp and add own setting for this
+            UseActionResultType = true;
```
RSuter: I think this should be a setting here: https://github.com/RSuter/NSwag/blob/master/src/NSwag.CodeGeneration.CSharp/SwaggerToCSharpControllerGeneratorSettings.cs

src/NSwag.CodeGeneration.CSharp/Models/CSharpOperationModel.cs
```diff
@@ -58,7 +58,7 @@ public class CSharpOperationModel : OperationModelBase<CSharpParameterModel, CSh
                 {
                     parameters = parameters
                         .OrderBy(p => p.Position ?? 0)
-                        .OrderBy(p => !p.IsRequired)
+                        .ThenBy(p => !p.IsRequired)
```
RSuter: I think this should not be changed, as all required parameters must come first for correctly working null defaults...

RSuter:
Thanks for the PR! I've added some comments...
I think we need to improve ASP.NET Core support in general (i.e. a setting where you can specify whether it's an ASP.NET Core target and maybe remove the AspNetNamespace setting) so that correct code is generated... or maybe an AspNetCoreVersion setting so you can decide on that what namespace, ActionResult etc to use?

src/NSwag.CodeGeneration.CSharp/Models/CSharpOperationModel.cs
optuuttila: Ok, I'll leave it as it was, but as far as I understand the ThenBy approach is the recommended way.
Here's more on that: https://stackoverflow.com/questions/3760001/linq-orderby-versus-thenby

src/NSwag.CodeGeneration.CSharp/Models/CSharpOperationModel.cs
optuuttila: Done.

src/NSwag.CodeGeneration.CSharp/Models/CSharpOperationModel.cs
optuuttila: Done.

src/NSwag.CodeGeneration.CSharp/Models/CSharpOperationModel.cs
optuuttila: Done.

optuuttila:
Yes, I agree. The ASP.NET Core support should be enhanced.
Right now it required quite some tweaking to get the correct controller code generated and it would be very nice if there was (more or less) a single setting that would use the default values and get it mostly right from start.

RSuter:
Is this PR ready now?

optuuttila:
Yes, unless you think it needs something else.

yberstad:
@RSuter, this is just what I'm looking for, do you know when it will be available?

src/NSwag.CodeGeneration.CSharp/Models/CSharpOperationModel.cs
uygary: That's true. However, if we want t keep the existing behavior, `.OrderBy(p => p.Position ?? 0).OrderBy(p => !p.IsRequired)` would need to be replaced by `.OrderBy(p => !p.IsRequired).ThenBy((p => p.Position ?? 0))` or the sorting order would change because the current IsRequired sorting is taking precedence since it's being applied after the Position sorting.

RicoSuter:
Sorry for the delay, see https://github.com/RicoSuter/NSwag/pull/1965#issuecomment-478200142

Will ping here

yberstad:
@RicoSuter, thank you very much!

RicoSuter:

RicoSuter:
Ok, I've updated your PR and moved the logic to a new model class. Please review - then we can merge I think.

PR #1943, opened 2019-02-07
Add support for custom CSS and JavaScript in the Swagger UI

### LINKED ISSUE #1137: How to add custom Swagger UI
As compared to SwashBuckle does Nswag provide a way to inject custom assets like the index.html and the styles.

Sample Swashbuckle Code
------------------------------
c.CustomAsset("index", thisAssembly, "Synapse.RF.Api.SwaggerExtension.index.html");
c.InjectStylesheet(thisAssembly, "Synapse.RF.Api.SwaggerExtension.swaggerExt.css");

Looking for a similar implementation in NSwag. Any help on this issue is really appreciated.

Crevitus:
Allows users to add a URI to load a custom CSS stylesheet and a URI to load a custom JavaScript file into the Swagger index.html.

Resolves #1137

### DIFF
--- a/src/NSwag.AspNetCore/SwaggerUi/index.html
+++ b/src/NSwag.AspNetCore/SwaggerUi/index.html
@@ -11,6 +11,7 @@
   <link href='css/screen.css' media='screen' rel='stylesheet' type='text/css'/>
   <link href='css/reset.css' media='print' rel='stylesheet' type='text/css'/>
   <link href='css/print.css' media='print' rel='stylesheet' type='text/css'/>
+  <link href='{StylesheetUri}' media='screen' rel='stylesheet' type='text/css'/>
 
   <script src='lib/object-assign-pollyfill.js' type='text/javascript'></script>
   <script src='lib/jquery-1.8.0.min.js' type='text/javascript'></script>
@@ -89,6 +90,7 @@
       }
   });
   </script>
+  <script src="{ScriptUri}"> </script>
 </head>
 
 <body class="swagger-section">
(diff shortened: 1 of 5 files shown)


RSuter:
Isnt it a problem that by default it renders an empty src attribute?

Crevitus:
> Isnt it a problem that by default it renders an empty src attribute?

This doesn't present a problem apart from being technically invalid markup, I could add an update to remove the custom tags if unused?

RSuter:
I think it's fine as long as it still works without these urls, have you tested this?

Crevitus:
> I think it's fine as long as it still works without these urls, have you tested this?

Yes I have tested it with and without the urls, it works fine :)

RSuter:
Ok... thanks for the PR, will be in the next version in the next days..

gohalien:
@RicoSuter this is nice, but it would be also nice to load up a custom index.html, that way you can use a modified swagger-ui.js

RicoSuter:
So we'd need another setting (e.g. `string IndexHtml { get; set; }`) which is taken instead of the internal resource if defined here: 
https://github.com/RicoSuter/NSwag/blob/master/src/NSwag.AspNetCore/Middlewares/SwaggerUiIndexMiddleware.cs#L28

RicoSuter:
The question is whether we want this extension point - at some point it just makes more sense to host Swagger UI yourself so that you can do whatever you want...

gohalien:
> The question is whether we want this extension point - at some point it just makes more sense to host Swagger UI yourself so that you can do whatever you want...

Well, for me it could be nice, but of course I really don't know if it is a required feature. My idea was for example adding a property in SwaggerUiSettingsBase, and use it instead of a fixed call to index.html in 
https://github.com/RicoSuter/NSwag/blob/master/src/NSwag.AspNetCore/Extensions/NSwagApplicationBuilderExtensions.cs#L76

PR #1945, opened 2019-02-08
Add support for Axios

thijscrombeen:
Added support for Axios by
- Adding Axios as enum
- Added AxiosClient.liquid
- Modifying existing files to support axios

### DIFF
--- a/src/NSwag.CodeGeneration.TypeScript/Templates/Client.ProcessResponse.HandleStatusCode.liquid
+++ b/src/NSwag.CodeGeneration.TypeScript/Templates/Client.ProcessResponse.HandleStatusCode.liquid
@@ -28,12 +28,8 @@ return response.blob().then(blob => { return { fileName: fileName, data: blob, s
 {%     if response.HasType -%}
 let result{{ response.StatusCode }}: any = null;
 {%         if Framework.IsAxios -%}
-{%              if response.UseDtoClass or response.IsDate -%}
 let resultData{{ response.StatusCode }}  = _responseText;
 {{ response.DataConversionCode }}
-{%              else -%}
-result{{ response.StatusCode }} = _responseText === "" ? null : <{{ response.Type }}>{% if operation.HandleReferences %}jsonParse{% else %}JSON.parse{% endif %}(_responseText, this.jsonParseReviver);
-{%              endif -%}
 {%         else -%}
 {%              if response.UseDtoClass or response.IsDate -%}
 let resultData{{ response.StatusCode }} = _responseText === "" ? null : {% if operation.HandleReferences %}jsonParse{% else %}JSON.parse{% endif %}(_responseText, this.jsonParseReviver);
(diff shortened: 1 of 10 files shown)


DaanMertens:
Great addition, can't wait to see it merged!

src/NSwag.CodeGeneration.TypeScript/Templates/Client.ProcessResponse.HandleStatusCode.liquid
```diff
@@ -27,18 +27,29 @@ return response.blob().then(blob => { return { fileName: fileName, data: blob, s
 {% template Client.ProcessResponse.ReadBodyStart %}
 {%     if response.HasType -%}
 let result{{ response.StatusCode }}: any = null;
-{%         if response.UseDtoClass or response.IsDate -%}
-let resultData{{ response.StatusCode }} = _responseText === "" ? null : {% if operation.HandleReferences %}jsonParse{% else %}JSON.parse{% endif %}(_responseText, this.jsonParseReviver);
+{%         if Framework.IsAxios -%}
```
RSuter: Is this whole if block required - ie. you are inverting the logic for axios...?

thijscrombeen: Hi Rico,

The reason I added this block is because the AxiosResponse' data property is not a string of text.

`{% elseif Framework.IsAxios -%}
{% else -%}`

It is already a json object, hence it does not need
`{% if operation.HandleReferences %}jsonParse{% else %}JSON.parse{% endif %}(_responseText, this.jsonParseReviver);` 

Keeping that line would give me a exception in the form of invalid json at position 0...

RSuter: What about the else block in axios? Should it really use jsonparse etc?

RSuter: And what about reference handling, will this just no be supported?

thijscrombeen: I did not find a case where I would hit the else block. I tried with my own swagger and with the petstore swagger. So if you want I can remove the else block.

As for the reference handling I don't see anything wrong with the TS code when using the petstore swagger.
`let _mappings: { source: any, target: any }[] = [];
        if (status === 200) {
            let result200: any = null;
            result200 = resultData200 ? User.fromJS(resultData200, _mappings) : new User();
            return result200;`

However, when I use my own swagger I sometimes find that the _mappings variable is not defined because it does not hit the following part:
`{% if operation.HandleReferences and operation.RequiresMappings -%}`

The TS generated does include the fromJS(result, _mappings) like so
` result200 = resultData200 ? User.fromJS(resultData200, _mappings) : new User();`

I believe this is because of 
`{{ response.DataConversionCode }}`
I believe the dataconversioncode only checks operation.HandleReferences and does not look at operation.RequiresMappings. But I think that's code of NJsonSchema.

I only have this issue with my own swagger so it could also be that there's something wrong there.

RSuter: I think we can remove the if statement and else block then?

thijscrombeen: Done

RSuter:
Thank you very much for the PR! Will be in v12.0.14 soon, https://github.com/RSuter/NSwag/pull/1958

PR #1946, opened 2019-02-08
Mac compatibility

framinosona:
A simple "/" vs "\" error shows up when using NSwag in a mac Terminal, here's the fix I found.


RSuter:
v12.0.14

Issue #1776, opened 2018-11-27
Invalid generated method names (with brackets)


semso-software:
Hello @RSuter 

I was trying generate a c# client class from a swagger json file.
The tool generates the client nicely but it won't build because of brackets in the method names.

```
System.Threading.Tasks.Task<ApiResponse> OrdersAddedOnOn}Async(System.DateTime added_on,
```

My nswag file:
```
"codeGenerators": {
    "swaggerToCSharpClient": {
      "clientBaseClass": null,
      "configurationClass": null,
      "generateClientClasses": true,
      "generateClientInterfaces": false,
      "generateDtoTypes": true,
      "injectHttpClient": true,
      "disposeHttpClient": true,
      "protectedMethods": [],
      "generateExceptionClasses": true,
      "exceptionClass": "SwaggerException",
      "wrapDtoExceptions": true,
      "useHttpClientCreationMethod": false,
      "httpClientType": "System.Net.Http.HttpClient",
      "useHttpRequestMessageCreationMethod": false,
      "useBaseUrl": true,
      "generateBaseUrlProperty": true,
      "generateSyncMethods": false,
      "exposeJsonSerializerSettings": false,
      "clientClassAccessModifier": "public",
      "typeAccessModifier": "public",
      "generateContractsOutput": false,
      "contractsNamespace": null,
      "contractsOutputFilePath": null,
      "parameterDateTimeFormat": "s",
      "generateUpdateJsonSerializerSettingsMethod": true,
      "serializeTypeInformation": false,
      "queryNullValue": "",
      "className": "{controller}Client",
      "operationGenerationMode": "SingleClientFromPathSegments",
      "additionalNamespaceUsages": [],
      "additionalContractNamespaceUsages": [],
      "generateOptionalParameters": true,
      "generateJsonMethods": true,
      "enforceFlagEnums": false,
      "parameterArrayType": "System.Collections.Generic.IEnumerable",
      "parameterDictionaryType": "System.Collections.Generic.IDictionary",
      "responseArrayType": "System.Collections.ObjectModel.ObservableCollection",
      "responseDictionaryType": "System.Collections.Generic.Dictionary",
      "wrapResponses": false,
      "wrapResponseMethods": [],
      "generateResponseClasses": true,
      "responseClass": "SwaggerResponse",
      "namespace": "MyNamespace",
      "requiredPropertiesMustBeDefined": true,
      "dateType": "System.DateTime",
      "jsonConverters": null,
[...]

PR #1954, opened 2019-02-13
#1776 Omit brackets in operation names

thomasmischke:
ConvertPathToName checked, if a path segment contains "{", but if the
parameters enclosed in those brackets includes underscores, the
resulting operation name included "}", which is an illegal character in
operation names for most languages.
The new introduced regex eliminates parameters from the path before
creating the operation name. A test has been added to check that it is
working.


RSuter:
Thanks for the PR!

RSuter:
v12.0.14

Issue #1955, opened 2019-02-13
C# Web API Controller : Generated code does not handle "required" fields for path adn query param


nikobiker12:
Hello, I use NSwag to generate C# Web API controller code from swagger document. and It looks like required field of path/query param is not "translated" into "System.ComponentModel.DataAnnotations.Required"

For instance: here's apart of my swagger path definition file
```javascript
"/myroute": {
  "get": {
        "parameters": [
          {
             "name": "type",
             "type": "string",
             "in": "query",
             "required": true          // <---  required parameters 
          }
        ]
    }
}
```

I would want the code generated to be
```csharp
[Microsoft.AspNetCore.Mvc.HttpGet, Microsoft.AspNetCore.Mvc.Route("myroute")]
public Task<string> GetMethod([System.ComponentModel.DataAnnotations.Required] string type) // <-- The attribute Required was added
{
    return type;
}
```

What is you opinion ?

I think the only thing needed is to modify the template Controller.liquid by adding
something like this :
```javascript
{% if parameter.IsRequired %}[System.ComponentModel.DataAnnotations.Required]{% endif %}
```

RSuter:
In ASP.NET Core we need to use BindRequiredAttribute and not the data annotation's RequiredAttribute

PR #1965, opened 2019-02-19
#1955 C# Web API Controller : Generator does not handle "required" fi…

nikobiker12:
Can you review the PR please ?
I added two parameters in the settings to sets whether attributes must be generated or not and which type of attribute must be generated.
I refactored the unit tests by using a swagger document "manually" generated instead of using an generated one


src/NSwag.CodeGeneration.CSharp/SwaggerToCSharpControllerGeneratorSettings.cs
```diff
@@ -24,6 +24,8 @@ public SwaggerToCSharpControllerGeneratorSettings()
             UseCancellationToken = false;
             AspNetNamespace = "System.Web.Http";
             RouteNamingStrategy = CSharpControllerRouteNamingStrategy.None;
+            UseModelValidationAttributes = false;
+            RequiredAttributeType = "System.ComponentModel.DataAnnotations.Required";
```
RSuter: ASP.NET Core needs the BindRequiredAttribute to make parameters required not the RequiredAttribute

nikobiker12:
I modified the value of the RequiredAttributeType to BindRequired. Is it Ok ?

RSuter:
I’m wondering if we should just drop & remove web api support and concentrate completely on aspnetcore? I’d say yes...

/cc @rogersillito

optuuttila:
In my view all focus should be put to aspnetcore. 

That's definitely the direction where things are going and allthough there will still be a lot of older webapi users for the time being I think this would be yet another reason for them to just(*) migrate to adpnetcore.

* Been there so I know it's not that straightforward, but hey, that's why we have strangler patterns etc. that come to rescue.

rogersillito:
We need to supply a controller interface for a client that will likely still need Web API support, so I would vote to retain support for that right now.   Is there something specific about this PR that's problematic for supporting both?

src/NSwag.CodeGeneration.CSharp/SwaggerToCSharpControllerGeneratorSettings.cs
rogersillito: I just read a post which may be useful here: [have BindAttribute behave as BindRequiredAttribute in ASP.NET Core MVC](https://www.strathweb.com/2017/12/required-and-bindrequired-in-asp-net-core-mvc) - see _"But I don’t want to mix BindRequiredAttribute and RequiredAttribute"_

RicoSuter:
Ah man, it seems that nswag will support “legacy” framework much longer than the asp team does 🙂

Ok, ill add a new IsAspNetCore setting, remove the namespace setting (as it can be inferred) and then all controller gen PRs have to align with that and check the new flag - they must support at least aspnetcore but must not break web api

RicoSuter:

src/NSwag.CodeGeneration.CSharp/Models/CSharpControllerTemplateModel.cs
```diff
@@ -77,6 +77,12 @@ public class CSharpControllerTemplateModel : CSharpTemplateModelBase
         /// <summary>Gets a value indicating whether to allow adding cancellation token.</summary>
         public bool UseCancellationToken => _settings.UseCancellationToken;
 
+        /// <summary>Gets a value indicating whether to allow adding model validation attributes</summary>
+        public bool UseModelValidationAttributes => _settings.UseModelValidationAttributes;
+
+        /// <summary>Gets the type of the attribte used to specify a parameter as required.</summary>
+        public string RequiredAttributeType => _settings.RequiredAttributeType;
```
RicoSuter: Why do we need a setting for that? Can't we just check the ControllerTarget (AspNet/AspNetCore) and "hardcode" the full attribute name? Don't see a scenario where one would choose another type...

src/NSwag.CodeGeneration.CSharp/SwaggerToCSharpControllerGeneratorSettings.cs
```diff
@@ -53,5 +55,11 @@ public string GetRouteName(SwaggerOperation operation)
 
         /// <summary>Gets or sets the strategy for naming routes (default: CSharpRouteNamingStrategy.None).</summary>
         public CSharpControllerRouteNamingStrategy RouteNamingStrategy { get; set; }
+
+        /// <summary>Gets or sets a value indicating whether to allow adding model validation attributes</summary>
+        public bool UseModelValidationAttributes { get; set; }
```
RicoSuter: Maybe rename to GenerateModelValidationAttributes or GenerateValidationAttributes ?

RicoSuter:
Please update your branch with master branch (fix conflicts)

src/NSwag.CodeGeneration.CSharp/Models/CSharpControllerTemplateModel.cs
RicoSuter: Implement it the same way as latest version of the AspNetNamespace property?

src/NSwag.CodeGeneration.CSharp/SwaggerToCSharpControllerGeneratorSettings.cs
nikobiker12: I ve renamed to GenerateModelValidationAttributes

nikobiker12:
@RicoSuter 
I ve juste took into account your remarks and merge the master branch
Sorry for the delay

src/NSwag.CodeGeneration.CSharp/Models/CSharpControllerTemplateModel.cs
nikobiker12: done

RicoSuter:
v12.2.2

PR #1975, opened 2019-02-22
Generate additional namespace usages from the swagger definitions

mikesingleton:
Adds an option to generate additional namespace usages from the swagger definitions

Background:
My service and client projects both have dependencies on libraries that contain DTOs.
My swagger.json contains the namespaces in the definitions but there's currently no way to utilize those in NSwag (as far as I know).
These changes allow add an option to NSwag to generate namespace usages for DTOs directly from the swagger definitions (given that they're already included in the swagger.json, otherwise the option does nothing).

### DIFF
--- a/src/NSwag.CodeGeneration.CSharp/Models/CSharpFileTemplateModel.cs
+++ b/src/NSwag.CodeGeneration.CSharp/Models/CSharpFileTemplateModel.cs
@@ -51,11 +51,6 @@ namespace NSwag.CodeGeneration.CSharp.Models
         /// <summary>Gets the namespace.</summary>
         public string Namespace => _settings.CSharpGeneratorSettings.Namespace ?? string.Empty;
 
-        /// <summary>Gets the all the namespace usages.</summary>
-        public string[] NamespaceUsages => (_outputType == ClientGeneratorOutputType.Contracts ?
-            _settings.AdditionalContractNamespaceUsages?.Where(n => n != null).ToArray() :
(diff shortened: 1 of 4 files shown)


RicoSuter:
This solves a quite specific problem and also is a leaky abstraction (swagger should be self-describing and the server technology should not not leak into it).

Why do you need these usings/what in the generated code is using it? Do you exclude types or are not all types in the spec?

nbowditch:
Currently, our web api is implemented in C# (AspNetCore) using DTO classes from a separate library of data models, e.g., `My.DtoLib.SomeObject`. We are trying to generate C# clients that will use those same model classes, so we have `generateDtoTypes` set to `false` and the `excludedTypeNames` set to an empty array.

Our thinking was that the "DRY-est" practice would be to have the data model definitions exist in one place. Since we will be calling the web api from our own server logic, we want to be able to use the client with our own types.

However, the generated client classes are referencing the data model classes without the full namespace, e.g. `SomeObject`. We couldn't find any setting for the `nswag.json` file that would add the full namespace, so we came up with this solution. 

Any feedback you can give us would be greatly appreciated - thanks for taking the time to look at this

RicoSuter:
Can't you just add your namespaces using the AdditionalNamespaceUsages setting and import just import your existing types?

nbowditch:
We were hoping for a solution where we wouldn't need to manually add new namespaces/dependencies, especially since those namespaces were already being automatically added to the swagger.json file we are generating. That being said, feel free to close this PR if you guys don't like the design, it does seem like a very specific problem. Thanks again for the feedback!

Issue #1666, opened 2018-10-13
Improving performance


kimbell:
I've been using NSwag for some time now, and have been browsing the generated code on more than one occasion. Today I was made aware of a some work David Fowler is doing to educate the community on best practices around async code. One of his 'bad' examples rang a bell. 

Assuming your code generation hasn't undergone major changes recently, NSwag generates code like this: 
```cs
var responseData_ = response_.Content == null ? null : await response_.Content.ReadAsStringAsync().ConfigureAwait(false);
var result_ = default(AcceptanceReply);
try
{
      result_ = Newtonsoft.Json.JsonConvert.DeserializeObject<AcceptanceReply>(responseData_, _settings.Value);
      return result_;
 }
catch (System.Exception exception_)
{
      throw new SwaggerException("Could not deserialize the response body.", (int)response_.StatusCode, responseData_, headers_, exception_);
}
```
The whole string is read into memory before it gets deserialized into an object. 

Fowler is suggesting this improved approach:
https://github.com/davidfowl/AspNetCoreDiagnosticScenarios/blob/master/Scenarios/Services/PokemonService.cs#L36
```cs
public async Task<PokemonData> GetPokemonAsync()
{
      var response = await _client.GetAsync(_url);
      // This is a hack to work around the fact that this JSON api returns text/plain
      response.Content.Headers.ContentType = new MediaTypeHeaderValue("application/json");
      // Use the built in methods to read the response as JSON
      return await response.Content.ReadAsAsync<PokemonData>();
 }
```

SebastianStehle:
Just until recently I had the same approach of using the stream to parse the json, but recently we had a big integration projection with lots of external services and we decided to read the full body to memory and log it with all request and response headers. And we had almost no performance decrease.

SebastianStehle:
True, should be easy to make a PR for that ;)

Jehoel:
In my hand-written clients I implement two different deserialization methods: one using an intermedia `String` and the other using `JsonTextReader` (or just the built-in object deserialization method in `HttpClient`) wrapped in `#if DEBUG` so the Release code uses the fast code and debug code lets me see the raw request and response text. Is this an approach that NWag could use?

RSuter:
@jehoel, sounds good so long as we dont break anything and dont change the public api

Jehoel:
@RSuter Thank you - I'll prepare a PR sometime this week.

RSuter:
[...]

PR #1976, opened 2019-02-23
Async Stream-based JSON serialization

Jehoel:
As mentioned in this issue thread: https://github.com/RSuter/NSwag/issues/1666

<del>This PR also includes other changes that are orthogonal to the intended change of only enabling stream-based deserialization instead of always using an intermediate `String` instance.</del>

This PR passes all JSON object deserialization work to a new method:

    protected virtual async Task<...> ReadObjectResponseAsync<T>(HttpResponseMessage response, ...)

This method uses the `Stream` from `HttpResponseMessage.ReadAsStreamAsync()` and `JsonTextReader` to deserialize the expected object from the stream without using an intermediate `String` instance. This _should_ result in reduced memory consumption, especially when dealing with large JSON object responses. Because the response is not buffered into a string it means that if an error is encountered the program cannot view the raw response body as a string.

If the `DEBUG` compilation symbol is defined then the method will instead read the response using an intermediate string (Passing the string from `ReadAsStringAsync()` into `JsonConvert.DeserializeObject`) and it will include that string in any raised exceptions.

I wonder if `#if DEBUG` could be replaced with `#if STRING_DESERIALIZATION` and adding `#if DEBUG\r\n#define STRING_DESERIALIZATION\r\n#endif` to the top of the file might be more flexible. I'll see what the project maintainers say.

Unfortunately Json.NET still does not offer an Async `DeserializeObject<T>` method. While `JsonReader`/`JsonWriter` do support async IO, they're only for reading and writing JSON tokens (`JToken`, `JObject`, `JArray`, etc) and not for whole-object deserialization/serialization. I was thinking that I could read the response asynchronously using `JsonReader` into a root `JToken` however it isn't feasible to then deserialize an object from a `JToken` - a quick look inside Json.NET revealed that deserialization uses a very different code-path entirely and doesn't use `JToken` as an intermediary. It doesn't look like this will be fixed in Json.NET given it's a long outstanding request (dating back to 2012) and Microsoft have announced their own forthcoming JSON serialization/deserialization library which they say will have an emphasis on performance.


juanperezADD:
Any new information about when it will be merged?

Thanks in advance

RicoSuter:
I cannot merge this PR as there are too many unrelated changes (e.g. change of full qualified types to using usings which do pollute the namespace with lots of types which might lead to name clashes/compile errors for some users), etc. please create/update the PR with only the stream-based JSON serialization feature for a review...

juanperezADD:
@Jehoel 
Do you have plans on updating the PR with the changes that @RicoSuter suggested?

Jehoel:
@RicoSuter @juanperezADD Sorry for the delay - I've rebased my changes to the latest `master` and restricted my changes to only the addition and use of `ReadObjectResponseAsync<T>`.

SebastianStehle:
Btw: The new json serializers for .net core 3.0 do not use async, because for small payloads it is killing your performance, due to many context switches.

Jehoel:
@SebastianStehle off-topic: but if Microsoft's JSON API will not support async reading from streams, what is their official guidance for deserializing large objects? Surely they could use some form of buffering?

I do wish the Async IO API in .NET included a way for sources to pass metadata like ideal buffer size to their sinks so synchronous parsers can read tokens from an async source, for example.

SebastianStehle:
@Jehoel I think the guidance is to use another serializer ;) ... The System.Text.Json thing is optimized for Web Apis where you typically have small payloads of a few KB to MB

Jehoel:
@RicoSuter are we good to merge? :)

src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.ConvertToString.liquid
```diff
@@ -20,9 +20,9 @@
     else if (value is bool) {
         return System.Convert.ToString(value, cultureInfo).ToLowerInvariant();
     }
-    else if (value is byte[])
```
RicoSuter: Please keep this as is because there are some users which cannot use these C# 7 features...

src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.ProcessResponse.liquid
```diff
@@ -6,45 +6,32 @@ var fileResponse_ = new FileResponse((int)response_.StatusCode, headers_, respon
 client_ = null; response_ = null; // response and client are disposed by FileResponse
 return fileResponse_;
 {%         else -%}
-var responseData_ = response_.Content == null ? null : await response_.Content.ReadAsStringAsync().ConfigureAwait(false); 
-throw new {{ ExceptionClass }}("{{ response.ExceptionDescription }}", (int)response_.StatusCode, responseData_, headers_, null);
+ObjectResponseResult<{{ response.Type }}> objectResponse_ = await this.ReadObjectResponseAsync<{{ response.Type }}>(response_, headers_).ConfigureAwait(false);
```
RicoSuter: Do not use `this.`

src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.liquid
```diff
@@ -58,6 +58,53 @@
...
+            this.ResponseText = responseText;
+        }
+
+        public T      ResponseObject { get; }
+        public string ResponseText   { get; }
+    }
+
+    protected virtual async System.Threading.Tasks.Task<ObjectResponseResult<T>> ReadObjectResponseAsync<T>(System.Net.Http.HttpResponseMessage response, System.Collections.Generic.IReadOnlyDictionary<string,System.Collections.Generic.IEnumerable<string>> responseHeadersAsDictionary)
+    {
+#if DEBUG
```
RicoSuter: Do we really need DEBUG code here?

src/NSwag.CodeGeneration.CSharp/Templates/File.liquid
```diff
@@ -8,10 +8,14 @@
 using {{ usage }};
 {% endfor -%}
 
+#pragma warning disable  108 // Disable "CS0108 '{derivedDto}.ToJson()' hides inherited member '{dtoBase}.ToJson()'. Use the new keyword if hiding was intended."
```
RicoSuter: Can't we just use `#pragma warning disable` as it was before?

src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.liquid
```diff
@@ -58,6 +58,53 @@
...
+            throw new {{ ExceptionClass }}(message: message, (int)response.StatusCode, objectBodyAsString, responseHeadersAsDictionary, jsonEx);
+        }
+#else
+        try
+        {
+            using (System.IO.Stream responseStream = await response.Content.ReadAsStreamAsync().ConfigureAwait(false))
+            using (System.IO.StreamReader streamReader = new System.IO.StreamReader(responseStream))
+            using (Newtonsoft.Json.JsonTextReader jsonTextReader = new Newtonsoft.Json.JsonTextReader(streamReader))
+            {
+                Newtonsoft.Json.JsonSerializer serializer = Newtonsoft.Json.JsonSerializer.Create(this.JsonSerializerSettings);
```
RicoSuter: Maybe we should not recreate the serializer for each request?

src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.ConvertToString.liquid
Jehoel: Done. I note this method has multiple other possible optimizations, but I'll reserve those for another time.

src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.ProcessResponse.liquid
Jehoel: I assume that's to allow the use of `static` implementations?

src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.liquid
Jehoel: That's intentional - to allow developers to inspect the server responses when debugging if there's an incorrect server response.

I've changed it to a different symbol (`NSWAG_STRING_RESPONSES`) to be clearer, so users have to explicitly opt-in to this less efficient code-path by enabling the symbol in their project's source code.

src/NSwag.CodeGeneration.CSharp/Templates/File.liquid
Jehoel: Using `pragma warning disable all` would mask any new warnings that my changes, or other people's changes, might introduce. Personally I would fix the code to address all of the warnings instead of hiding them, but if we're going to hide them then we should be explicit about it.

src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.liquid
Jehoel: This is a bit tricky. While `JsonSerializer` is thread-safe, `JsonTextReader` is not, and the `JsonSerializerSettings` object is also directly exposed via the `JsonSerializerSettings { get { ... } }` property. See here: https://stackoverflow.com/questions/36186276/is-the-json-net-jsonserializer-threadsafe - I note this issue exists in the existing NSwag-generated code too.

I did try to use a single instance of `JsonSerializer`, however the `JsonSerializerSettings` property makes use of a `Lazy<T>` field to store it, and creating two objects from a single `Lazy<T>` is non-trivial.

I don't think there are any good solutions while we use `Lazy<T>` and have an exposed property at the same time.

src/NSwag.CodeGeneration.CSharp/Templates/File.liquid
RicoSuter: So if we remove the ExposeJsonSerializerSettings and never expose or assume it is only changed for initialization we could reuse the serializer?

Jehoel: (I think you meant to post this in reply to a different comment) Yes. But that would be a breaking change. I'll leave that up to you.

RicoSuter: Ok. Fine for me. I hope you catched all warnings in all possible generation paths :-)

src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.liquid
RicoSuter: I personally would define set up the implicit rule that the settings cannot be changed when the first operation method is called only before when setting up the client... create a new client when you need different settings. With this rule we could reuse the serializer... The serialization settings are only exposed to use in external serializations not to modify it - but there is not a read-only version of the settings that's why the mutable settings have to be exposed... (public settings property option)

RicoSuter:
Please review the PR. I've fixed the templates.

Jehoel:
I just fixed a bug I overlooked earlier where `objectBodyAsString` was undefined. LGTM now. Thank you for accepting it! :D

src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.ReadObjectResponse.liquid
```diff
@@ -0,0 +1,33 @@
+﻿protected virtual async System.Threading.Tasks.Task<ObjectResponseResult<T>> ReadObjectResponseAsync<T>(System.Net.Http.HttpResponseMessage response, System.Collections.Generic.IReadOnlyDictionary<string, System.Collections.Generic.IEnumerable<string>> headers)
+{
+#if NSWAG_STRING_RESPONSES // Define this symbol in your project properties to enable inspecting response bodies as strings.
```
RicoSuter: Maybe we should generate a property on the client instead of this compiler switch? This way you can use it in code also if its published eg via nuget...

The problem with this is that you will probably the only one who knows about this/uses this otherwise

RicoSuter:
I need to do more testing before i can merge (with actual running the client. Did anyone do that with a bigger api?

src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.ReadObjectResponse.liquid
```diff
@@ -0,0 +1,33 @@
+﻿protected virtual async System.Threading.Tasks.Task<ObjectResponseResult<T>> ReadObjectResponseAsync<T>(System.Net.Http.HttpResponseMessage response, System.Collections.Generic.IReadOnlyDictionary<string, System.Collections.Generic.IEnumerable<string>> headers)
```
RicoSuter: We should rename this method to ReadJsonObjectResponseAsync as there plans to also support xml deserialization which would require a similar method

RicoSuter:
I’ve added two more comments and i think it would also make sense to change the wording of responseobject vs objectresponse as this is a bit confusing 🙂

src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.ReadObjectResponse.liquid
Jehoel: I like that idea. But there are a few options:

* Single static class `NwagOptions` with a static `Boolean ExposeStringResponses` property.
    * Advantages: Only need to set a single property to enable inspection of all responses regardless of generated client class or when using multiple NSwag-generated libraries.
* Static `Boolean ExposeStringResponses` property on each generated Client class.
     * Advantages: More granularity.
* Instance `Boolean ExposeStringResponses` property on each generated Client class.
     * Advantages: Even greater granularity - but complicates the interface. Mock/testing classes would need to be updated to include this property and alter their behaviour too.
* As a method parameter with a default value, e.g. `GetSomething( String param1, Boolean exposeStringResponses = false, CancellationToken ct = default );`
    * Advantages: Avoids race-conditions.
* Private field set in the `Client` constructor or factory method.
    * Advantages: Doesn't change interface, avoids race-conditions. I'm unsure if consumers would want to change the string-exposure behaviour after the Client object has been created. Presumably it's a setting they'd keep in their `appSettings.json` to help investigate runtime exceptions from unexpected responses.

RicoSuter: I'd go with 

> Instance read/write Boolean ExposeStringResponses property on each generated Client class.

as static properties are bad for isolation between different users in the same process.

The property should only be on the class itself not on the client interface...

We could still make the default of the property dependent on a compile switch, e.g. when NSWAG_STRING_RESPONSES is set then set the property to true otherwise false.

PR #1981, opened 2019-02-25
Generate distinct variables if the names contains special characters

Itori:
In some cases, the generated name of several variables are the same.
In my case, the api provide a parameter $limit and a parameter limit, and both generate a variable limit.

### DIFF
--- a/src/NSwag.CodeGeneration/DefaultParameterNameGenerator.cs
+++ b/src/NSwag.CodeGeneration/DefaultParameterNameGenerator.cs
@@ -15,6 +15,8 @@ namespace NSwag.CodeGeneration
     /// <summary>The default parameter name generator.</summary>
     public class DefaultParameterNameGenerator : IParameterNameGenerator
     {
+        private readonly Dictionary<string, short> _knowParameters = new Dictionary<string, short>();
+
         /// <summary>Generates the parameter name for the given parameter.</summary>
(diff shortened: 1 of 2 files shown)


src/NSwag.CodeGeneration/DefaultParameterNameGenerator.cs
```diff
@@ -15,6 +15,8 @@ namespace NSwag.CodeGeneration
     /// <summary>The default parameter name generator.</summary>
     public class DefaultParameterNameGenerator : IParameterNameGenerator
     {
+        private readonly Dictionary<string, short> _knowParameters = new Dictionary<string, short>();
```
RSuter: Is it really needed to have state in this generator? The problem is that it might be reused and then not resetted...

Itori: As the generator is created on each call, it should work, but you're right, it might be reused and not resetted.

One solution would be to reuse the "allParameters" variable, looking for each previous parameter in the collection to see if there is already one that generate that name, but it would repeat the generate method a lot. So not a good solution.

An other possibility could be to give a hashet<string> to the method, giving the already known names, but i don't really like this idea.

My last idea was to handle it in the CSharpOperationModel, when generating, but it would need to be done for each language. Or we could put it in the OperationModelBase with a state or an input parameter.

What do you think ?

PR #1849, opened 2018-12-21
Improve generation of "consumes" in ASP.NET Core, closes #420

RSuter:
Closes #420 

Tasks: 

- [x] Same for produces
- [x] If consumes/produces is empty then the consumes/produces entry can still be moved to global (?)

PR #2022, opened 2019-03-15
TS: Support for url encoded requests

robinkrokfors:
Allows the TypeScript client to generate `application/x-www-form-urlencode` requests.

Also requires that the ConsumesAttribute is taken into account on controller actions #1849.


RSuter:
Info: Consumes is probably not available in openapi 3 and thus we need to get this info from somewhere else

RSuter:
Maybe add

```
        public bool ConsumesFormUrlEncoded => 
            _operation.ActualConsumes?.Any(c => c == "application/x-www-form-urlencoded") == true ||
            _operation.ActualResponses.Any(r => r.Value.Content.Any(c => c.Key == "application/x-www-form-urlencoded"));
```

for OpenApi3 support

robinkrokfors:
I'll check up on this when I get home.

Why are we using `_operation.ActualResponses` for checking if the operation should do a url encoded request? Instead checking something like `_operation.RequestBody` would make more sense and follow the OpenApi 3 specification, although my brief inspection on `SwaggerOperation` shows that will not be possible at the moment.

RSuter:
Ah shit, thats completely wrong what i posted - you are right :-)

robinkrokfors:
`ConsumesFormUrlEncoded` is now checking the `_operation.RequestBody` for a url encoded media type. Although currently `SwaggerOperation.RequestBody` doesn't seem to be populated correctly when a document is being generated, though that is probably a separate issue.

RSuter:
> Although currently SwaggerOperation.RequestBody doesn't seem to be populated correctly when a document is being generated, though that is probably a separate issue.

If the schema type is OpenApi3 it should be populated as expected, is that not the case?

RSuter:
Is this PR ready?

robinkrokfors:
It is populated if the parameter has a [FromBody] attribute on it, but in OpenApi3 formData is not a valid parameter in-type, therefore these also have to generate in the `SwaggerOperation.RequestBody`, is that not so?

This PR should be good to go.

RSuter:
> It is populated if the parameter has a [FromBody] attribute on it, but in OpenApi3 formData is not a valid parameter in-type, therefore these also have to generate in the SwaggerOperation.RequestBody, is that not so?

Ok, is there already an issue for that?

> This PR should be good to go after the ConsumesAttribute is taken into account.

Just looking into this other PR

robinkrokfors:
I did notice that the other PR #1849 does only take care of the ConsumesAttribute with the `AspNetCoreToSwaggerGenerator` I guess without testing, The `WebApiToSwaggerGenerator` used in for example the tests is not covered by this.

For my testing needs I have added a simple `OperationConsumesProcessor` in the `WebApiToSwaggerGeneratorSettings`, I can include this as well since #1849 seems to be only about the ASP.NET core generation specifically.

RSuter:
So we need the same for web api as in the other PR? That would be the "right" way, correct?

robinkrokfors:
That would be correct, I have added a `OperationConsumesProcessor` that should take care of this with the WebApi generator.

RSuter:
Can you fix the conflicts? But then i think it looks good...

robinkrokfors:
The conflicts are now resolved.

PR #2027, opened 2019-03-19
Don't call UpdateJsonSerializerSettings in csharp template if not present

slang25:
If `generateUpdateJsonSerializerSettingsMethod` is set to `false`, and set `jsonSerializerSettingsTransformationMethod` to something, then the generated code will try to invoke the missing `UpdateJsonSerializerSettings` method.

### DIFF
--- a/src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.liquid
+++ b/src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.liquid
@@ -36,7 +36,9 @@
 {% if SerializeTypeInformation -%}
             settings.TypeNameHandling = Newtonsoft.Json.TypeNameHandling.Auto;
 {% endif -%}
+{% if GenerateUpdateJsonSerializerSettingsMethod -%}
             UpdateJsonSerializerSettings(settings);
+{% endif -%}
             return settings;
         });
     }


RSuter:
Thanks for the PR (and your blog post) - much appreciated!

RSuter:
https://github.com/RSuter/NSwag/pull/2029 v12.0.16

jochenjonc:
This change broke my build because I have a BaseClient where I implemented the UpdateJsonSerializerSettings method. But now it isn't called anymore.

If I activate this new setting, the UpdateJsonSerializerSettings it doesn't get called also because it is a partial method now.

slang25:
Wow, I hadn't considered that scenario. You could change the base class method name and mark is as `protected`, then call it from `UpdateJsonSerializerSettings` in a partial class.

This would be a breaking change for anyone consuming your client however, so you'd need to bump your major version number.

slang25:
I've just noticed there's a `HasBaseClass` in the template, the call to `UpdateJsonSerializerSettings` could be made conditional on that?

jochenjonc:
The HasBaseClass would be a solution.

Using a partial class is not an option, I have multiple clients and I would have to make a partial class per client.

RSuter:
Reverted both PRs, just keep `generateUpdateJsonSerializerSettingsMethod` is set to true and use the `jsonSerializerSettingsTransformationMethod` - the UpdateJsonSerializerSettings method will be optimized away anyway and will not have any impact

RSuter:
@slang25 Why was this an issue in the first place?

slang25:
@RSuter User error on reflection, I had set `generateUpdateJsonSerializerSettingsMethod` to `false` for some reason and convinced myself that it was being called in error, but I can see now that you'd only do this in the case it wasn't provided by a base class.

Apologies for wasting your time here, and thanks again for maintaining this great library 😃

Issue #1386, opened 2018-06-15
Route prefixing does not appear to work for dotnet core


DannyRyman:
We are attempting to use SwaggerToCSharpControllerGenerator to generate a dotnet core compatible web api. Unfortunately we have a slight problem with route prefixing. The generator is generating the code `[routeprefix('v1')]` whereas, I believe that using dotnet core this needs to be `[route('v1)]`

RSuter:
Did you enable the .net core settings?

DannyRyman:
Thanks for your help. Unless I am misunderstanding, I think you misread my question. I am trying to do contract/schema first development using the SwaggerToCSharpControllerGenerator. As I understand it, the IsAspNetCore setting is used when wiring up swagger to your web api on startup? My problem is that the code generated by the SwaggerToCSharpControllerGenerator does not build because it uses routeprefix which is not valid for dotnet core projects, and there does not appear to be a configuration option to change this?

RSuter:
Sorry, the setting does not exist and RoutePrefix is always used: https://github.com/RSuter/NSwag/blob/master/src/NSwag.CodeGeneration.CSharp/Templates/Controller.liquid#L28

cromo:
Our team is running in to this as well, and we're using NSwag.MSBuild to automate the generation of the controller code. We're working around this issue by running a string substitution on the result via powershell, e.g.

```xml
  <Target Name="NSwag" BeforeTargets="Build">
    <Exec Command="$(NSwagExe_Core21) run swagger.nswag" />
    <Exec Command="powershell -Command &quot;(Get-Content LoggingController.cs).replace('RoutePrefix', 'Route') | Set-Content LoggingController.cs&quot;" />
  </Target>
```

That said, there's interest here to make this work out of the box. How involved would that be? The change to the liquid template looks minor, but I haven't looked at how much supporting code would need to be changed nor accounted for testing nor NSwagStudio.

altso:
Another way to make it work is to define `RoutePrefixAttribute`:
```csharp
// ReSharper disable once CheckNamespace
namespace Microsoft.AspNetCore.Mvc
{
    public class RoutePrefixAttribute : RouteAttribute
    {
        public RoutePrefixAttribute(string template)
            : base(template)
        {
        }
    }
}
```

RSuter:
I think we should collect all missing stuff and then decide how to proceed. 

Simplest way is to remove the namespace setting and add isaspnetcore instead and handle this in a single tpl.

[...]

Issue #642, opened 2017-03-01
How to disable generating parameters for headers using SwaggerToCSharpWebApiControllerGenerator


att741:
Swagger document:
```
{
  "swagger": "2.0",
  "paths": {
    "/authorization": {
      "put": {
        "operationId": "Test",
        "parameters": [
          {
            "name": "Authorization",
            "in": "header",
            "required": true,
            "type": "string"
          }
        ],
        "responses": {
          "204": {
            "description": ""
          }
        }
      }
    }
  }
}
```

Generated code:
```
namespace MyNamespace
{
    #pragma warning disable // Disable all warnings

    [System.CodeDom.Compiler.GeneratedCode("NSwag", "9.6.6264.38294")]
    public interface IController
    {
        System.Threading.Tasks.Task TestAsync(string authorization);
    
    }
    
    [System.CodeDom.Compiler.GeneratedCode("NSwag", "9.6.6264.38294")]
    public partial class Controller : System.Web.Http.ApiController 
    {
        private IController _implementation; 
    
        public Controller(IController implementation)
        {
            _implementation = implementation; 
        }
    
        [System.Web.Http.HttpPut, System.Web.Http.Route("authorization")]
        public System.Threading.Tasks.Task Test(string authorization)
        {	
            return _implementation.TestAsync(authorization);
            
        }
    
    }

}
```

rsuter:
This is currently not possible. Do you need this setting on a per operation basis or global?

att741:
By default Web API binds parameters only from the URI or the message body, so Swagger header parameters should not be generated at all by SwaggerToCSharpWebApiControllerGenerator.

rsuter:
Ah sorry, this is for the controller generator... There we should exclude headers or use FromHeader attribute when generating .net core controllers (not supported yet). Can you create a pr with this?

att741:
Currently, only Web API 2 controllers (System.Web.Http.ApiController) are supported and generated by SwaggerToCSharpWebApiControllerGenerator using ControllerTemplate.tt.
To my mind, this template should not generate headers at all.

As for .Net Core MVC (Microsoft.AspNetCore.Mvc.Controller) support, there might be so many differences between Web API 2 and .Net Core MVC to introduce the new SwaggerToCSharpCoreMvcControllerGenerator, which could use new settings and a new template.

Issue #1697, opened 2018-10-30
Additional Namespaces Usages should be applied to the namespaces used by generated C# client anyway


Jehoel:
In the current NSwag C# client generation template all types have fully-qualified namespace names, e.g.

    var content_ = new System.Net.Http.StringContent(Newtonsoft.Json.JsonConvert.SerializeObject(dto, _settings.Value));

In NSwag.Studio I set "Additional namespace usages" to `System.Net.Http,System.Threading.Tasks,Newtonsoft.Json` in the expectation that this would cause the generated C# code to look like this instead;

    var content_ = new StringContent(JsonConvert.SerializeObject(dto, _settings.Value));

...but it doesn't, because [the namespaces are hardcoded in the C# templates](https://github.com/RSuter/NSwag/blob/f7aef821e047196211b72da46cfc0833bbe3ea1d/src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.liquid): 

If the aim is to ensure the types fully resolve then the generated code should use the `global::` keyword like the VS2005 WinForms designer did.

I have other template changes I'd like to make, I'll be happy to submit a PR but because this enters _opinionated programming_ territory I imagine the project owners are not very receptive to changes to those templates, or are they?

The changes I'd like to make include:

- Remove `#pragma warning disable` and generate code that doesn't generate warnings in the first place (and use specific targeted warning-disable pragmas, and/or make it an NSwag configuration option).
- Allow me to specify a base-class for the exception.
- Use type-names instead of `var`.
- Read and deserialize responses with `JsonTextReader` instead of reading into a `String` first, which is faster and avoids extra string allocations (but retain the `String` version for debugging)
- Treat status codes as `int` not `string` (why does this happen?)
- Use `using` instead of trivial `try/finally/dispose` 
- Use a single object deserialization method instead of the excessive amount of code duplication seen in every operation method.

My question:

* Would a PR with these changes be accepted? 
* If not, would a PR that enables these changes as runtime configuration options be accepted?
* Finally, if not that, then a PR that makes the C# code-generation templates as runtime, user-editable templates?

RSuter:
Ref: https://github.com/RSuter/NSwag/issues/1666

aluitink:
I ran into a similar issue using the NSWagCSharp OpenApiReference CodeGenerateor via the csproj file.

I have my own DTOs so generating with /GenerateDtoTypes:false

I could not find an option that allows you to specify additional namespaces to be used in the generated client.

[...]

PR #2033, opened 2019-03-20
Support header parameters for CSharp controller code generation

### LINKED ISSUE #1956: Custom request header params(in: header) are treated as query parameters
Hi,

I use SwaggerToCSharpControllerGenerator to convert my OpenAPI 3.0.0 specification intowebAPI controllers. In this specification, I added a custom request header property as you can see in the example:
```
...
  /pets:
    get:
      operationId: getAllPets
      summary: Get all pets 
      tags: [ 'Pets' ]
      parameters:
        - $ref: "#/components/parameters/requestIdHeader"
...
parameters:
     requestIdHeader:
      in: header
      name: X-requestId
      schema:
        type: string
        format: uuid
      required: false
      description: Unique identifier (GUID) of a specific request.
```

  Result of the code generation:
```
        /// <summary>Get all pets</summary>
        /// <param name="x_requestId">Unique identifier (GUID) of a specific request.</param>
        /// <returns>Successful</returns>
        [System.Web.Http.HttpGet, System.Web.Http.Route("pets")]
        public System.Threading.Tasks.Task<Members> GetAllPets(ystem.Guid? x_requestId)
        {
            return _implementation.GetAllPetsAsync(x_requestId);
        }
```
    

As you can see, the property 'in header' is not taken into account during the code generation because it is added as a query parameter. 
If support for custom headers will not be provided, could it be possible to exclude the custom request header properties(in: header) from the code generation?

Thank you in advance!

rogersillito:
Work in progress - this fixes #1956 (and possibly #642?).   I've not looked at test coverage yet, and have done a limited range of manual testing.   

Essentially I'm creating a `FromHeaderAttribute` that handles binding of header-based parameters.   I've only tested in a framework web project so far, but in theory this should happily co-exist with the one available in ASP.NET Core for now.   [Looking at the discussion around how Core should be supported with the templates](https://github.com/RSuter/NSwag/issues/1386#issuecomment-415066531), I wonder whether this approach might be sufficient until that is resolved; maybe then only including the custom attribute/binding when the _proposed_ `IsAspNetCore` setting is false?


RSuter:
I think it's imporant to first fix the web api vs asp.net core story and then generate which works for each framework... it seems to me that these custom attributes are web api only, right? There should be a check so that they are only generated for web api...

rogersillito:
I was working on the assumption that the `HttpParameterBinding`/`ParameterBindingAttribute` base classes were still available in Core (but having just tested my code in a Core project, _I realise they are not_).   As such _my assumption this would still happily work in Core was wrong_.   

Saying that, if we had a way to tell the template if we're using Core it would be pretty straightforward to switch the behaviour: this works fine using the built-in `FromHeaderAttribute`. (We could do a string comparison against the AspNetNamespace setting in the template, but that wouldn't be nice..!).

rogersillito:
If you did decide to drop support for Web API, I wondered about making use of custom templates as a workaround for my use case...  

I've just been exploring the "Template Directory" setting ([following your comment](https://github.com/RSuter/NSwag/issues/1697#issuecomment-434437774)): providing an alternative version of __Controller.liquid__ works as I would expect (despite the caution to [only override empty/extension templates?](https://github.com/RSuter/NJsonSchema/wiki/Templates)).   The only thing I would specifically need from this PR would be the extra properties on [`CSharpParameterModel`](https://github.com/RSuter/NSwag/pull/2033/files#diff-f14aab342b6f347552b53a9d54b41192R49) and [`ParameterModelBase`](https://github.com/RSuter/NSwag/pull/2033/files#diff-5087220fa49fcb3eb6cc4e9aef5bb94fR165): the built-in template could then be _fixed to work for Core only - and remove the custom attribute_.

RicoSuter:
Sorry for the delay, see https://github.com/RicoSuter/NSwag/pull/1965#issuecomment-478200142

Will ping here when it’s ready

RicoSuter:

rogersillito:
Great - thanks for your work on this @RicoSuter .  I'll hopefully get my PR revised over the next week or so!

src/NSwag.CodeGeneration.CSharp/Templates/Controller.liquid
```diff
@@ -91,8 +91,61 @@ public abstract class {{ Class }}ControllerBase : {% if HasBaseClass %}{{ BaseCl
...
+    public abstract {%  if operation.WrapResponse %}System.Threading.Tasks.Task<System.Net.Http.HttpResponseMessage>{% else %}{{ operation.ResultType }}{% endif %} {{ operation.ActualOperationName }}({% for parameter in operation.Parameters %}{% if parameter.IsHeader %}[FromHeader{% if parameter.IsValidIdentifier == false %}(Name = "{{ parameter.Name }}"){% endif %}] {% endif %}{% if parameter.IsBody %}[{{ AspNetNamespace }}.FromBody] {% endif %}{{ parameter.Type }} {{ parameter.VariableName }}{% if GenerateOptionalParameters and parameter.IsOptional %} = null{% endif %}{% if parameter.IsLast == false or UseCancellationToken %}, {% endif %}{% endfor %}{% if UseCancellationToken %}System.Threading.CancellationToken cancellationToken{% endif %});
 
 {%     endfor -%}
 }
 {% endif -%}
+
+public class FromHeaderBinding : {{ AspNetNamespace }}.Controllers.HttpParameterBinding
```
RicoSuter: This should maybe only be generated in the aspnet and not aspnetcore case?

rogersillito: Yes agreed.   I was thinking for the AspNetCore case, we could then conditionally insert this using statement (so the method declaration could then be the same in each case):

`using FromHeaderAttribute = Microsoft.AspNetCore.Mvc.FromHeaderAttribute;`

?

rogersillito: _I've made the changes as above_ - that does mean it's a little inconsistent compared to how [FromBody] works (which has a fully-qualified namespace) - but it simplifies the template a bit where it's already quite complex.

src/NSwag.CodeGeneration.CSharp/Templates/Controller.liquid
```diff
@@ -1,4 +1,8 @@
-﻿{% if GeneratePartialControllers -%}
+﻿{% if IsAspNetCore -%}
+using FromHeaderAttribute = {{ AspNetNamespace }}.FromHeaderAttribute;
```
RicoSuter: This won't work as the Controller.liquid template is repeated multiple times depending on the client generation mode...

src/NSwag.CodeGeneration.CSharp/Templates/Controller.liquid
RicoSuter: Same problem here: This class will be repeated... we need to add this as an additional code artifact in C# based on the selected option and target framework

Also I we really have to avoid usings as they pollute the current namespace (name clashes) and makes the code coupled to the file template (which is against the code artifact concept) - ideally the code should be self-describing and not depend on some usings somewhere else

rogersillito: Ok no problem, I didn't appreciate that; we can the attributes self-describe.

rogersillito:
@RicoSuter: having figured out the concerns around multiple controller generation I've made a few changes to this PR in line with your comments:

- added a new template that gets run once all controllers have been generated: the custom `FromHeaderAttribute` code for AspNet gets inserted when it's needed  _(i.e. only if there are header parameters)_
- replaced the previous using statements with full namespace prefixing where required
- added tests that cover the above (and use the `GetSwaggerDocument()` setup approached introduced by another contributor)

RicoSuter:
I've updated the code to use code artifacts (used for multiple file output later)

rogersillito:
Having digested your changes that it looks like that approach fits much better with the general approach of the code: _I hadn't fully appreciated the `CodeArtifact` concept_.   The danger of focusing exclusively on the fix at hand without an overall view on the codebase!

RicoSuter:
> I hadn't fully appreciated the CodeArtifact concept. The danger of focusing exclusively on the fix at hand without an overall view on the codebase!

No worries - this whole CodeArtifact was not really incorporated yet - I just took this as an opportunity to add it and change it everywhere...

Everything looks fine now except some additional line breaks etc. will merge soon. Is your feature still working as expected?

rogersillito:
I'll give it another try if you've finished integrated - _at the last try it looked good._

RicoSuter:
Ok. Now everything should be ready... with your go I'll merge

rogersillito:
I've got it building, the tests are passing and my code gen project is running OK - so fine by me:  thanks for you help!

RicoSuter:
v12.3.0

PR #2054, opened 2019-03-27
Corrected behavior for BaseUrl

ArmanHayots:
As possible fix to https://github.com/RicoSuter/NSwag/issues/2050


RicoSuter:
Need to play with the various options and see what makes sense - and how we can support your scenario without breaking existing users.

PR #2057, opened 2019-03-28
Fix generation date parameters using form data

stbychkov:
Hey there!

When we create a `multipart/form-data` request and pass date parameters, they transform into the `string` using `.toString()` function, but you used `.toJSON()` transformation for dates in every part of this repository.

Controllers cannot parse format like this: 
> Thu Mar 28 2019 10:39:19 GMT+0200 (Eastern European Standard Time)

Therefore we must transform dates to ISO format.


RicoSuter:
LGTM. Thanks for the PR!

PR #2061, opened 2019-03-29
Allow detection of application/vnd.*+json as Json response

GeorgDangl:
85c5c2bfe4ede6c216246702e40cd6dd3b6e26a1 updated the detection of binary responses. It only checks for `application/json`, but ignores other valid values, such as `application/vnd.model+json`. This caused the following regression for this controller action:

```csharp
[Produces("application/vnd.com.danglit.ProjectDto.v1+json")]
[ProducesResponseType(typeof(ProjectDto), 200)]
public async Task<IActionResult> ConvertToAvaAsync()
```

Produces the following, shortened Swagger document:

```json
"post": {
  "produces": [
    "application/vnd.com.danglit.ProjectDto.v1+json"
  ],
  "responses": {
    "200": {
      "x-nullable": true,
      "schema": {
        "$ref": "#/definitions/ProjectDto"
      }
    }
  }
}
```

This is now no longer detected as Json and results in the C# Client generating a `FileResponse`.

### DIFF
--- a/src/NSwag.Core.Tests/SwaggerResponseTests.cs
+++ b/src/NSwag.Core.Tests/SwaggerResponseTests.cs
@@ -1,0 +1,28 @@
+﻿using Xunit;
+
+namespace NSwag.Core.Tests
+{
+    public class SwaggerResponseTests
+    {
+        [Theory]
+        [InlineData("application/octet-stream", true)]
+        [InlineData("undefined", true)]
+        [InlineData("text/plain", false)]
+        [InlineData("application/json", false)]
+        [InlineData("application/vnd.model+json", false)]
+        public void When_response_contains_produces_detect_if_binary_response(string contentType, bool expectsBinary)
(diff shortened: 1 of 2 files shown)


RicoSuter:
Thanks for the pr!

RicoSuter:
Maybe we need this check here too?

https://github.com/RicoSuter/NSwag/blob/85c5c2bfe4ede6c216246702e40cd6dd3b6e26a1/src/NSwag.Core/SwaggerParameter.cs#L206

GeorgDangl:
@RicoSuter, I'm not sure what the other lines do, or where they are used:)

With 12.0.20, the generated client as well as the swagger.json on the server are both correct again.

Thank you very much for the extremely fast response & release!

RicoSuter:
Ok then, let’s see if someone reports a problem 🙂. Thanks again for the PR!

PR #2066, opened 2019-04-01
Use controller name from ApiExplorer instead of custom logic for AspNetCoreToSwaggerGenerator

softworkz:
With ApiExplorer, we can do better than this:

            if (controllerName.EndsWith("Controller"))
                controllerName = controllerName.Substring(0, controllerName.Length - 10);

### DIFF
--- a/src/NSwag.SwaggerGeneration.AspNetCore/Processors/OperationTagsProcessor.cs
+++ b/src/NSwag.SwaggerGeneration.AspNetCore/Processors/OperationTagsProcessor.cs
@@ -1,0 +1,30 @@
+﻿//-----------------------------------------------------------------------
+// <copyright file="OperationTagsProcessor.cs" company="NSwag">
+//     Copyright (c) Rico Suter. All rights reserved.
+// </copyright>
+// <license>https://github.com/NSwag/NSwag/blob/master/LICENSE.md</license>
+// <author>Rico Suter, mail@rsuter.com</author>
+//-----------------------------------------------------------------------
+
+using Microsoft.AspNetCore.Mvc.Controllers;
+using NSwag.SwaggerGeneration.Processors;
+using NSwag.SwaggerGeneration.Processors.Contexts;
+
+namespace NSwag.SwaggerGeneration.AspNetCore.Processors
+{
+    /// <summary>Processes the SwaggerTagsAttribute on the operation method.</summary>
+    public class OperationTagsProcessorNetCore : OperationTagsProcessor
+    {
+        protected override void AddControllerNameTag(OperationProcessorContext context)
+        {
+            var aspNetCoreContext = (AspNetCoreOperationProcessorContext)context;
+            if (aspNetCoreContext.ApiDescription.ActionDescriptor is ControllerActionDescriptor descriptor)
+            {
+                aspNetCoreContext.OperationDescription.Operation.Tags.Add(descriptor.ControllerName);
+                return;
+            }
+
+            base.AddControllerNameTag(context);
+        }
+    }
+}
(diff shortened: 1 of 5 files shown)


src/NSwag.SwaggerGeneration/Processors/Collections/OperationProcessorCollection.cs
```diff
@@ -21,5 +21,23 @@ public T TryGet<T>()
...
+        /// <typeparam name="T">The operation processor type to replace.</typeparam>
+        /// <param name="newItem">The replacement item.</param>
+        /// <returns>true, if an item was replaced; otherwise false.</returns>
+        public bool Replace<T>(IOperationProcessor newItem) where T : IOperationProcessor
+        {
+            var item = this.OfType<T>().FirstOrDefault();
+
+            if (item != null)
+            {
+                this.SetItem(this.IndexOf(item), newItem);
```
RicoSuter: Remove this.

src/NSwag.SwaggerGeneration.AspNetCore/Processors/OperationTagsProcessor.cs
```diff
@@ -0,0 +1,30 @@
...
+//-----------------------------------------------------------------------
+
+using Microsoft.AspNetCore.Mvc.Controllers;
+using NSwag.SwaggerGeneration.Processors;
+using NSwag.SwaggerGeneration.Processors.Contexts;
+
+namespace NSwag.SwaggerGeneration.AspNetCore.Processors
+{
+    /// <summary>Processes the SwaggerTagsAttribute on the operation method.</summary>
+    public class OperationTagsProcessorNetCore : OperationTagsProcessor
```
RicoSuter: Rename to AspNetCoreOperationTagsProcessor and move to a new file

How breaking is this change? Usually it should produce the same tags as before, right?

RicoSuter:
Thanks for the PR! Looks good. See my comments.

src/NSwag.SwaggerGeneration/Processors/Collections/OperationProcessorCollection.cs
softworkz: Just to make sure I understand correctly:

You want me to remove the Replace-method from the collection and do the replacement of the OperationTagsProcessor with explicit code inside AspNetCoreToSwaggerGeneratorSettings instead?

src/NSwag.SwaggerGeneration.AspNetCore/Processors/OperationTagsProcessor.cs
softworkz: Will do.

softworkz: Yes, it should produce the same tags only that we leave it up to the runtime to do the controller naming.
In case the controller name is empty, we still fallback to the base class implementation.

src/NSwag.SwaggerGeneration/Processors/Collections/OperationProcessorCollection.cs
RicoSuter: Sorry no, just remove the `this.` prefix :-)

softworkz: LOL... done.

src/NSwag.SwaggerGeneration.AspNetCore/Processors/OperationTagsProcessor.cs
softworkz: I have done the renaming.

PS: I did a hard cast to AspNetCoreOperationProcessorContext because that can be taken for granted. For the ControllerActionDescriptor I didn't because at some time you might decide to add support or documenting page controllers as well and then it could be a PageActionDescriptor instead.

softworkz:
Resolved your comments.

RicoSuter:
Ready?

softworkz:
Yes.

RicoSuter:
Ok.. thanks for the PR.

Issue #1550, opened 2018-08-23
Problem when adding an api controller method with an IFormFile parameter


srollinet:
Hi,

I tried to add a new method in one of my controller. This method takes an `IFormFile` parameter
As soon as I add this method, there is a new `consumes` array in swagger.json which only contains the value `"multipart/form-data"`

Problem is when I generate the ts client, all the methods try to send multipart/form-data

I tried to create a simplified repro project, but I cannot reproduce it :(
Any idea about the cause?

This is the new controller

```
    [ApiVersion("1")]
    public class ConfigurationController : ApiControllerBase
    {
        [HttpPost, DisableRequestSizeLimit]
        public void UploadFile(IFormFile file)
        {
        }
    }
```

swagger.json before I add the controller (there is no consumes array)
Note: in my repro project, there is always a consumes array. Do you know what can prevent it to appear?
```
{
  "x-generator": "NSwag v11.18.6.0 (NJsonSchema v9.10.67.0 (Newtonsoft.Json v11.0.0.0))",
  "swagger": "2.0",
  "info": {
    "title": "My Title",
    "version": "1.0.0"
  },
  "produces": [
    "text/plain",
    "application/json",
    "text/json"
  ],
  ...
}
```

swagger.json when I add the new controller
```
{
  "x-generator": "NSwag v11.18.6.0 (NJsonSchema v9.10.67.0 (Newtonsoft.Json v11.0.0.0))",
  "swagger": "2.0",
  "info": {
    "title": "My Title",
    "version": "1.0.0"
  },
  "consumes": [
    "multipart/form-data"
  ],
  "produces": [
    "text/plain",
    "application/json",
    "text/json"
  ],
  ...
}
```

RSuter:
Are you using UseSwagger*() or UseSwagger*WithApiExplorer()?

srollinet:
I use NSwag.MSBuild / aspNetCoreToSwagger 11.18.7

I managed to reproduce it here.

[nswag_consumes_issue.zip](https://github.com/RSuter/NSwag/files/2321417/nswag_consumes_issue.zip)

I think the problem is not the IFormFile parameter itself, but the absence of the `consumes` array if I remove the `UploadController`

miroga:
Hi,

do you have any update for this issue?
I'd like to upgrade to .net core 2.1 and this is blocking me.

srollinet:
[...]

PR #2067, opened 2019-04-01
Fix overwriting global consumes and produces sections

stbychkov:
Hey, there!

I recently upgraded from v12.0.14 and now I'm getting an error where several requests return **415 HTTP-response**.

Sorry, but I think you have a problem [here](https://github.com/RicoSuter/NSwag/blob/19ecd169938a78fe33afabf43cb1fa2e7c959be1/src/NSwag.SwaggerGeneration.AspNetCore/AspNetCoreToSwaggerGenerator.cs#L197).

For every controller you received from the `IApiDescriptionGroupCollectionProvider` you overwrite **global consumes** and it can be a problem when **operation's consume** attribute equals to the **global consume** value. Look at [here](https://github.com/RicoSuter/NSwag/blob/19ecd169938a78fe33afabf43cb1fa2e7c959be1/src/NSwag.SwaggerGeneration.AspNetCore/AspNetCoreToSwaggerGenerator.cs#L219). If **operation's consume** attributes equals to **global**, you don't write a consume section to the operation at all.

There is an example of this problem:

```
public class Controller1
{
    [Consumes("application/json")]
    public ViewModel GetSomeData() { ... }
}

public class Controller2 
{
    [Consumes("multipart/form-data")]
    public ViewModel SaveSomeData() { ... }
}
```

Controllers consume `application/json` by default. After processing `Controller1`, document will have 3 values in **global consume** section and no one in the operation's section (`GetSomeData`):

- application/json-patch+;
- application/json;
- application/*+json.

But after processing `Controller2` the **document consume section** will be overwritten with `multipart/form-data`. So as we don't have any **consume section** for `GetSomeData` operation and **global consume section** will be equals to `multipart/form-data`, every request to `Controller1` will return 415 HTTP-code. So this logic works only for one document per controller, but we have only one document creation [.](https://github.com/RicoSuter/NSwag/blob/19ecd169938a78fe33afabf43cb1fa2e7c959be1/src/NSwag.SwaggerGeneration.AspNetCore/AspNetCoreToSwaggerGenerator.cs#L64)

These changes are related to the following [commit](https://github.com/RicoSuter/NSwag/commit/556d4b3d6c020305b656ece7662adc8534e9af13)

I hope I gave you enough details to explain this case. 

Thanks.

### DIFF
--- a/src/NSwag.SwaggerGeneration.AspNetCore/AspNetCoreToSwaggerGenerator.cs
+++ b/src/NSwag.SwaggerGeneration.AspNetCore/AspNetCoreToSwaggerGenerator.cs
@@ -69,6 +69,20 @@ namespace NSwag.SwaggerGeneration.AspNetCore
                 .GroupBy(item => item.Item2.ControllerTypeInfo.AsType())
                 .ToArray();
 
+            document.Consumes = apiGroups
+                .SelectMany(s => s.SupportedRequestFormats)
+                .Select(s => s.MediaType)
+                .Where(m => apiGroups.All(a => !a.SupportedRequestFormats.Any() || a.SupportedRequestFormats.Contains(m)))
+                .Distinct()
+                .ToList();
+            
+            document.Produces = apiGroups
+                .SelectMany(c => c.SupportedResponseTypes)
+                .SelectMany(s => s.ApiResponseFormats.Select(f => f.MediaType))
+                .Where(c => apiGroups.All(o => !o.SupportedResponseTypes.Any() || o.SupportedResponseTypes.Contains(c)))
+                .Distinct()
+                .ToList();
+            
             var usedControllerTypes = new List<Type>();
             foreach (var controllerApiDescriptionGroup in apiGroups)
             {
@@ -187,40 +201,6 @@ namespace NSwag.SwaggerGeneration.AspNetCore
             List<Tuple<SwaggerOperationDescription, ApiDescription, MethodInfo, IEnumerable<string>, IEnumerable<string>>> operations,
             SwaggerGenerator swaggerGenerator, SwaggerSchemaResolver schemaResolver)
         {
-            var globalConsumes = operations
-                .SelectMany(o => o.Item4)
-                .Where(c => operations.All(o => o.Item4.Count() == 0 || o.Item4.Contains(c)))
-                .Distinct();
-
-            if (globalConsumes.Any())
-            {
-                if (document.Consumes == null)
-                {
-                    document.Consumes = globalConsumes.ToList();
-                }
-                else
-                {
-                    document.Consumes.AddRange(globalConsumes.Where(c => !document.Consumes.Contains(c)));
-                }
-            }
-
-            var globalProduces = operations
-                .SelectMany(o => o.Item5)
-                .Where(c => operations.All(o => o.Item5.Count() == 0 || o.Item5.Contains(c)))
-                .Distinct();
-
-            if (globalProduces.Any())
-            {
-                if (document.Produces == null)
-                {
-                    document.Produces = globalProduces.ToList();
-                }
-                else
-                {
-                    document.Produces.AddRange(globalProduces.Where(c => !document.Produces.Contains(c)));
-                }
-            }
-
             var addedOperations = 0;
             var allOperation = operations.Select(t => t.Item1).ToList();
             foreach (var tuple in operations)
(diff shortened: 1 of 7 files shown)


src/NSwag.SwaggerGeneration.AspNetCore/AspNetCoreToSwaggerGenerator.cs
```diff
@@ -187,40 +201,6 @@ public static JsonSerializerSettings GetJsonSerializerSettings(IServiceProvider
             List<Tuple<SwaggerOperationDescription, ApiDescription, MethodInfo, IEnumerable<string>, IEnumerable<string>>> operations,
```
RicoSuter: I think we can remove the last two parameters of the tuple, right?

src/NSwag.SwaggerGeneration.AspNetCore/AspNetCoreToSwaggerGenerator.cs
```diff
@@ -69,6 +69,20 @@ public async Task<SwaggerDocument> GenerateAsync(ApiDescriptionGroupCollection a
                 .GroupBy(item => item.Item2.ControllerTypeInfo.AsType())
                 .ToArray();
 
+            document.Consumes = apiGroups
+                .SelectMany(s => s.SupportedRequestFormats)
+                .Select(s => s.MediaType)
+                .Where(m => apiGroups.All(a => !a.SupportedRequestFormats.Any() || a.SupportedRequestFormats.Contains(m)))
```
RicoSuter: Why is it ok when empty (!any)? Is this correct?

src/NSwag.SwaggerGeneration.AspNetCore/AspNetCoreToSwaggerGenerator.cs
```diff
@@ -69,6 +69,20 @@ public async Task<SwaggerDocument> GenerateAsync(ApiDescriptionGroupCollection a
...
+                .Distinct()
+                .ToList();
+            
+            document.Produces = apiGroups
+                .SelectMany(c => c.SupportedResponseTypes)
+                .SelectMany(s => s.ApiResponseFormats.Select(f => f.MediaType))
+                .Where(c => apiGroups.All(o => !o.SupportedResponseTypes.Any() || o.SupportedResponseTypes.Contains(c)))
```
RicoSuter: Same here with !any. Is this actually solving the problem?

RicoSuter:
Is this a fix for https://github.com/RicoSuter/NSwag/issues/1550 ?

src/NSwag.SwaggerGeneration.AspNetCore/AspNetCoreToSwaggerGenerator.cs
stbychkov: Well, I think we should leave them because we need consume/produce values of the operation to check if they exists in a global section. And if they exist, we haven't to create the same sections in the operation scope.

RicoSuter: Ah ok... reviewing on my phone and not in an ide 🙂

RicoSuter:
Thanks for the pr! Maybe we should add some more unit test which cover the problems you’re solving

src/NSwag.SwaggerGeneration.AspNetCore/AspNetCoreToSwaggerGenerator.cs
stbychkov: You're right. I think this condition is redundant

src/NSwag.SwaggerGeneration.AspNetCore/AspNetCoreToSwaggerGenerator.cs
stbychkov: Fixed

stbychkov:
> Is this a fix for #1550 ?

I think it's related to this issue, but I'm not sure.

stbychkov:
> Thanks for the pr! Maybe we should add some more unit test which cover the problems you’re solving - this will help us avoiding regressions in the future

I've added tests to check this problem, but I can't build this project right now, because I don't have Mono environment at my laptop. Could you please check it out?

RicoSuter:
I will check... but try to build NSwag.Min.sln - it requires less skds (wix etc) - but on linux/mac you probably need mono for the netdx targets

RicoSuter:
Is the PR ready otherwise?

stbychkov:
> I will check... but try to build NSwag.Min.sln - it requires less skds (wix etc) - but on linux/mac you probably need mono for the netfx targets

I see, but I'm using mac

stbychkov:
> Is the PR ready otherwise?

It will be better, if you check it one more time :)

RicoSuter:
Sure, will run the integration tests and do some checks... 

How did you compile & test your changes then?

RicoSuter:
@srollinet will this solve your problem? See unit tests

stbychkov:
> How did you compile & test your changes then?

I have different environments and I tested there, but at the moment I don't have access to them.

srollinet:
@RicoSuter not 100% sure but yes it will probably fix it

RicoSuter:
Just fixed the compile errors but one test fails now...

But there’s a bigger problem: We are now generating the global collections before the operations have been filtered (eg with operation processors) so they are potentially wrong...

Will fix that soon

RicoSuter:
Please retest with version on master and report problems or create a new PR... but to me it looks fine now - tests are also passing

stbychkov:
> Please retest with version on master and report problems or create a new PR... but to me it looks fine now - tests are also passing

It works fine. Thank you!

PR #2081, opened 2019-04-03
Allow mime types that contain parameters to be identified as non binary

### LINKED ISSUE #2080: Incorrectly identifies mime types with optional parameters as binary
After this commit:
https://github.com/RicoSuter/NSwag/commit/85c5c2bfe4ede6c216246702e40cd6dd3b6e26a1

All mime types are treated as binary if they don't use the exact wording "application/json". If the mime type contains optional parameters for example "application/json;charset=UTF-8" the request body gets incorrectly identified as binary that yields unexpected results in generated client code.

Example swagger definition:

```javascript
    "paths": {
        "some/path": {
            "post": {
                "operationId": "someId",
                "consumes": [
                    "application/json;charset=UTF-8"
                ],
                "produces": [
                    "application/json"
                ]
            }
        }
    }
```

DanielGunnars:
fixes #2080

### DIFF
--- a/src/NSwag.Core/SwaggerParameter.cs
+++ b/src/NSwag.Core/SwaggerParameter.cs
@@ -203,6 +203,6 @@ namespace NSwag
         public bool IsBinaryBodyParameter => !IsXmlBodyParameter &&
                                              Kind == SwaggerParameterKind.Body &&
                                              ((SwaggerOperation)Parent).ActualConsumes?.Any() == true &&
-                                             ((SwaggerOperation)Parent).ActualConsumes?.Contains("application/json") != true;
+                                             ((SwaggerOperation)Parent).ActualConsumes?.Any(p => p.Contains("application/json")) == false;
     }
 }


src/NSwag.Core/SwaggerParameter.cs
```diff
@@ -203,6 +203,6 @@ public override bool IsNullable(SchemaType schemaType)
         public bool IsBinaryBodyParameter => !IsXmlBodyParameter &&
                                              Kind == SwaggerParameterKind.Body &&
                                              ((SwaggerOperation)Parent).ActualConsumes?.Any() == true &&
-                                             ((SwaggerOperation)Parent).ActualConsumes?.Contains("application/json") != true;
+                                             ((SwaggerOperation)Parent).ActualConsumes?.Any(p => p.Contains("application/json")) == false;
```
RicoSuter: Is there a reason why you changed it from != true to == false?

RicoSuter: Ok doesnt matter because of the previous any check

PR #2088, opened 2019-04-08
Fix generation of produces and consumes lists

softworkz:
There were several bugs with the current implementation:

- Document.Consumes and Document.Produces was overwritten each time when GenerateForControllerAsync was executed
- The conditions for adding a consumes or produces list to an operation were incorrect. These did a check whether an operation supports a content type that is not included in the document's consumes or produces lists. But actually, those lists need to be explicitly set at the operation level unless there's an exact match to the document level list

I'll create a separate PR for a third bug in this context


RicoSuter:
Isnt this already fixed with this PR?
https://github.com/RicoSuter/NSwag/pull/2067

PR #2089, opened 2019-04-08
Fix content (mime) types in SwaggerResponse

softworkz:
Previously, this was hardcoded to application/json or application/octet-stream. 
This resulted in missing mime types in OpenApi documents. 

E.g. it was impossible to have xml responses in OpenApi 3 documents.


mynameisD:
if I were read it correctly, OpenApi3 would only generate "application/json" or "application/octet-stream"? In my Application, I tried to add "text/plain" but not very successful. Will this PR solve the problem?

PR #2090, opened 2019-04-08
Add AspNetCoreOperationSecurityScopeProcessor

softworkz:
This doesn't compile by default!
It requires aspnetcore2.2

I still thought it might be worth sharing this as it solves the following problems:

- fixes bug with existing OperationSecurityScopeProcessor which marks all methods as requiring authorization
- Uses EndpointMetadata from ActionDescriptor instead of custom code for detecting attributes. This makes sure that we're getting the exact same results as the aspnetcore runtime
- Checks for presence of the AllowAnonymous attribute
- The current code in OperationSecurityScopeProcessor is incorrectly concatenating AuthorizeAttributes of the method with AuthorizeAttributes of the declaring type and then doing 'SelectMany'. This is incorrect. Attributes of the declaring type should only be used when there's no such attribute on the method. That problem doesn't exist with the code in this PR

### DIFF
--- a/src/NSwag.SwaggerGeneration.AspNetCore/Processors/AspNetCoreOperationSecurityScopeProcessor.cs
+++ b/src/NSwag.SwaggerGeneration.AspNetCore/Processors/AspNetCoreOperationSecurityScopeProcessor.cs
@@ -1,0 +1,86 @@
+﻿//-----------------------------------------------------------------------
+// <copyright file="AspNetCoreOperationSecurityScopeProcessor.cs" company="NSwag">
+//     Copyright (c) Rico Suter. All rights reserved.
+// </copyright>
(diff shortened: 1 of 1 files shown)


RicoSuter:
Ill merge and only compile it into the 2.2+ assemblies... this way it should work. Is that ok?

softworkz:
Yes, that's great. It's the code we're having in production...

RicoSuter:
Updated the wiki: https://github.com/RicoSuter/NSwag/wiki/AspNetCore-Middleware#enable-authorization-in-generator-and-swagger-ui

RicoSuter:
Added some reflection to make this work in .net standard 2.0...

Thanks for the PR.

PR #2104, opened 2019-04-11
Improve IsXmlBodyParameter, IsBinaryBodyParameter and SwaggerRespons.IsBinary()

### LINKED ISSUE #2076: NSwagStudio v12.1.0.0 mis-generates `FileResponse` for regular response model
@RicoSuter - NSwagStudio v12.1.0.0 mis-generates `FileResponse` for regular response model definition which points to a C# custom class type.

I can send you swagger.json file to reproduce the bug - ping me at me@dmitry-pavlov.com

RicoSuter:
Closes #2076

### DIFF
--- a/src/NSwag.Core.Tests/SwaggerResponseTests.cs
+++ b/src/NSwag.Core.Tests/SwaggerResponseTests.cs
@@ -10,6 +10,7 @@ namespace NSwag.Core.Tests
         [InlineData("text/plain", false)]
         [InlineData("application/json", false)]
         [InlineData("application/vnd.model+json", false)]
+        [InlineData("*/*", false)]
         public void When_response_contains_produces_detect_if_binary_response(string contentType, bool expectsBinary)
         {
             // Arrange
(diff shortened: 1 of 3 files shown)


dmitry-pavlov:
Looks good to me @RicoSuter

ljvanschie:
This could be tested by adding the following to https://github.com/RicoSuter/NSwag/blob/master/src/NSwag.Core.Tests/SwaggerResponseTests.cs#L13, right?

```cs
[InlineData("*/*", false)]
[InlineData("application/json;charset=UTF-8", false)]
```

If I'm seeing it right, it didn't fix it yet.

RicoSuter:
*/* doesnt make much sense to me TBH - how can you support everything?

RicoSuter:
But ill add this special case

RicoSuter:
Ok now?

ljvanschie:
Thanks! I agree it's a bit odd, but some libraries (like Springfox) default to `*/*`, and I think in that case it's more safe to assume the result isn't binary.

Looks good now 👍 

The other case (`[InlineData("application/json;charset=UTF-8", false)]`) still fails, but I'm not sure if it's a valid `produces` value.

RicoSuter:
Ok, I hope it's good enough now :-)

ljvanschie:
Great, thanks! 🙌

ljvanschie:
`produces` with value `*/*` is treated as binary again. I see the test has also been updated for this. Can this be reverted so that `*/*` is treated as a json response?

PR #2130, opened 2019-04-22
Feature/adds ability to host on different urls

### LINKED ISSUE #2129: NSwag should be able to work when the Swagger UI is exposed to more than one different URLs
The issue relates to the comment https://github.com/RicoSuter/NSwag/issues/1892#issuecomment-484969539

When an application is exposed to several URLs, _SwaggerDocumentMiddleware_ should cache __swaggerJson_ separately for every host/port/basePath.

StasPerekrestov:
Closes #2129 

When an application is exposed publicly to several URLs, we need to store generated swagger JSON documents separately and lookup the appropriate document using the corresponding values of **basePath** and **host** (I just skipped port number, but it's easy to add it).

What I don't really like in this PR is that I reuse PostProcess callback to extract the mentioned values to build a lookup key. It requires the creation of additional SwaggerDocument instance just for the sake of the two values transfer.


src/NSwag.AspNetCore/Middlewares/SwaggerDocumentKey.cs
```diff
@@ -0,0 +1,35 @@
+namespace NSwag.AspNetCore.Middlewares
+{
+    internal class SwaggerDocumentKey
```
RicoSuter: Can't we just use `Tuple<string, string, int>` (basePath, host, port) to avoid this code?

src/NSwag.AspNetCore/Middlewares/SwaggerDocumentMiddleware.cs
```diff
@@ -76,29 +79,51 @@ protected virtual async Task<string> GetDocumentAsync(HttpContext context)
             }
 
             var apiDescriptionGroups = _apiDescriptionGroupCollectionProvider.ApiDescriptionGroups;
-            if (apiDescriptionGroups.Version == Volatile.Read(ref _version) && _swaggerJson != null)
+
+            // The bellow code uses the swagger document Post Processing logic to get an actual
+            // host and basePath values. It looks like it's better to use a more lightweight DTO in PostProcess callback.
+            var doc = new SwaggerDocument();
+            _settings.PostProcess?.Invoke(doc, context.Request);
+            var key = new SwaggerDocumentKey(doc.BasePath ?? context.Request.PathBase, doc.Host ?? context.Request.Host.Host);
```
RicoSuter: Can't we just cache with host and port to avoid calling PostProcess? From what is the base path derived from? What are the scenarios?

src/NSwag.AspNetCore/Middlewares/SwaggerDocumentKey.cs
StasPerekrestov: What about named tuples? Of course, it'll require adding 
`<PackageReference Include="System.ValueTuple" Version="4.5.0" />`
but having something like
`(string basePath, string host, int port)` makes the intention behind tuple's arguments explicit.

The other thing is that we store the tuples in a dictionary and tuples are value types, so it'll add some boxing overhead, however, I don't think the swagger API is in the hot path, so I shouldn't be an issue.

src/NSwag.AspNetCore/Middlewares/SwaggerDocumentMiddleware.cs
StasPerekrestov: In one of my demo environments, we expose our API on a subpath (e.g. https://mydomain.com/my-demo-application/), so I have to rely on the subpath value.

I completely agree the initial changes look like a huge hack and we have to rethink it towards the more appropriate solution. Speaking of the machinery behind the code, let me put some words here:

When a user performs a request, the middleware doesn't know which version of a JSON document it should return. So, it should look up the right JSON document from the dictionary taking into account the source of the HTTP request(internal network access or it's a request behind the reverse proxy). To allow the middleware to make the right lookup we provide a callback. In the current PR, I rely on the logic of the PostProcess callback.

So, the code of the callback I use looks like:
```
config.PostProcess = (document, request) =>
					{
						document.Host = ExtractHost(request) + ":" + ExtractPort(request);
						document.BasePath = ExtractBasePath(request);
						document.Schemes.Clear();
						SwaggerSchema httpScheme = ExtractProto(request).EndsWith("s", StringComparison.OrdinalIgnoreCase)
							                           ? SwaggerSchema.Https
							                           : SwaggerSchema.Http;
						document.Schemes.Add(httpScheme);
					};
```
ExtractHost/ExtractPort/ExtractProto implementations are straightforward and look like
```
private string ExtractProto(HttpRequest request) => FirstOrDefault(request.Headers["X-Forwarded-Proto"]) ?? request.Scheme;
```

Perhaps, to avoid passing an empty document we just need to add another callback that allows getting rid of calling _PostProcess_ to build the lookup key.
e.g.
instead of
```
var doc = new SwaggerDocument();
_settings.PostProcess?.Invoke(doc, context.Request);
var key = new SwaggerDocumentKey(doc.BasePath ?? context.Request.PathBase, doc.Host...
```
use something like
`Func<(string basePath, string host, int port)> Request;`
So the user code looks like
```
.UseSwagger(config => {
    config.Request  = () => (ExtractBasePath(request), ExtractHost(request), ExtractPort(request));
                        // or
    config.Request  = () => /* build an url that indicates the application locatiion, e.g. https://mydomain.com/my-demo-application/ and use it instead of tuple */;
});
```

RicoSuter: Maybe something like `Func<HttpRequest, string> CreateDocumentCacheKey { get; set; }` on the settings class with a default implementation which just returns `string.Empty` (always the same key = current behavior) - this way you can customize it completely...

StasPerekrestov: LGTM :)

RicoSuter:
Is this ready now?

StasPerekrestov:
Well, it should be. Actually, I haven't tested it locally, it's not easy to build a single NuGet package for the aspnetcore project. However, the code looks reasonable at least to me.

The key creation logic also looks pretty safe. https://github.com/RicoSuter/NSwag/blob/e661b4b25e8117390421914cbeb6f8016d3de6e6/src/NSwag.AspNetCore/Middlewares/SwaggerDocumentMiddleware.cs#L83 

Also, if you find it required, I may perform testing locally and provide the results in this PR.

StasPerekrestov:
It works for both URLs (5000/5001 ports):
http://localhost:5000/swagger/index.html
https://localhost:5001/swagger/index.html
```
.UseSwagger(config =>
				{
					config.CreateDocumentCacheKey = request => ExtractHost(request) + ExtractPort(request);
...
}
```

checking for reverse-proxy scenario

StasPerekrestov:
Confirmed. The PR is ready

RicoSuter:
I've added exception caching - not so nice but works and is even less code, ok for you?

StasPerekrestov:
Just added a slight change that removes one dict lookup.

Thanks!

PR #2159, opened 2019-05-08
Fix #2158 by not initializing Security collection in Operation

### LINKED ISSUE #2158: Operation level ActualSecurity fails to resolve from document level Security
I think there is a bug here:

SwaggerOperation.cs
```
public ICollection<SwaggerSecurityRequirement> ActualSecurity => Security ?? Parent.Parent.Security;
```

Symptom is ```ActualSecurity``` will never be picked up from document level.

Fix:
~~Code need now to check if ```Security``` is an empty array or not, instead of that null check.~~

Update:
Turned out that we have this clause in OpenAPI specifications talking about operation level security requirement:
```
To remove a top-level security declaration, an empty array can be used.
```
Which means an empty array in operation should override an array in document level. Then I think we need to stop initializing ```Security``` array here (keeping it as null):

SwaggerOperation.cs
```
public ICollection<SwaggerSecurityRequirement> Security { get; set; } = new Collection<SwaggerSecurityRequirement>();
```

harouny:
Fix for issue: #2158

### DIFF
--- a/src/NSwag.Core/SwaggerOperation.cs
+++ b/src/NSwag.Core/SwaggerOperation.cs
@@ -138,7 +138,7 @@ namespace NSwag
 
         /// <summary>Gets or sets a security description.</summary>
         [JsonProperty(PropertyName = "security", Order = 14, DefaultValueHandling = DefaultValueHandling.Ignore)]
-        public ICollection<SwaggerSecurityRequirement> Security { get; set; } = new Collection<SwaggerSecurityRequirement>();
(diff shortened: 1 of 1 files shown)


RicoSuter:
Thanks for the pr, makes sense

harouny:
Thanks @RicoSuter Please let me know when a package with that fix deployed to Nuget

RicoSuter:
v12.3.0

Issue #2165, opened 2019-05-13
IOperationNameGenerator based on tag name and operation id


Place1:
Many of the OpenAPI Generators and Swagger Generators use the first tag+operation id to define the Service.method in the generated server or client side code.

In a similar manner to #1823 it'd be nice to have a `"operationGenerationMode": "MultipleClientsFromFirstTagAndOperationId"`.

I'd be happy to submit a PR for this if you'd accept the feature :D @RicoSuter

RicoSuter:
Isnt this what you are looking for:

https://github.com/RicoSuter/NSwag/blob/master/src/NSwag.CodeGeneration/OperationNameGenerators/MultipleClientsFromFirstTagAndPathSegmentsOperationNameGenerator.cs

PR: https://github.com/RicoSuter/NSwag/pull/1973

Place1:
@RicoSuter almost, except that uses `{tag}.{path}` to generate the controller/action where as I need `{tag}.{operationId}`.

From a quick glance it looks like i need [this line](https://github.com/RicoSuter/NSwag/blob/master/src/NSwag.CodeGeneration/OperationNameGenerators/MultipleClientsFromFirstTagAndPathSegmentsOperationNameGenerator.cs#L39) to use the operation id rather than the path. Then the resulting client and/or server side code would match the structure that OpenAPI generator or Swagger Generator produce.

PR #2169, opened 2019-05-15
added multiple clients from first tag and operation id generator

Place1:
this is a PR to implement the feature described in #2165


RicoSuter:
v12.3.0

PR #2174, opened 2019-05-16
Fix 'UseSwagger()' generating swagger failed when using 'swaggerToCSh…

FuturistiCoder:
…arpController' with controller style: Abstract

### Use Case:
- Framework: net462
- using swaggerToCSharpController generator to generate abstract class ***ControllerBase
- using swagger middleware `UseSwagger().UseSwaggerUi3()`
then NSwag throws an Unhandled Exception when generating http://localhost:****/swagger/v1/swagger.json 

Here is a bug from Microsoft:
https://github.com/dotnet/coreclr/issues/6600
https://stackoverflow.com/questions/38700721/reflection-with-generic-syntax-fails-on-a-return-parameter-of-an-overridden-meth
which throws an Unhandled Exception: System.IndexOutOfRangeException

This is a workaround for this bug.


RicoSuter:
v12.3.0

PR #2183, opened 2019-05-21
types2jsonschema command to generate json scema from C# files

dzmitry-lahoda:
Support for https://github.com/RicoSuter/NJsonSchema/issues/972 .

I like that swagger code already works well (all this hosting, type resolution, etc). So also strictly it is not swagger, but implementation wide it is. 

I have not done reuse or clean up. 

Would you accept pull with no copy paste reduction?
Would you accept pull without moving command from `swagger` namespace?

I have replaced `C:\Users\dzmitry\.nuget\packages\nswag.msbuild\12.3.0\tools\Win` with custom build and it works for me. 
```
  <Target Name="NSwag" AfterTargets="Build">
      <Exec Command="$(NSwagExe) types2jsonschema /assembly:Common.dll  /classnames:Foo.Bar /output:Configuration.schema.json" />
  </Target>
```
I failed to build nuget package and I do not have Visual Studio.  Still I want all my team to get new package seamlessly.


dzmitry-lahoda:
I have generate definition, but I also need somehow specify kind of root class after all definitions.

````
   "allOf": [
        {"$ref": "#/definitions/MyRootClass"},
    ]
```

How would I tell generator to generate that?

dzmitry-lahoda:
I could do 
```
        document.OneOf.Add(document.Definitions.First().Value);
            return document.ToJson();
```
With first class passed as being root. Would that be fine?

dzmitry-lahoda:
I have used first passed object as root object of schema by default

dzmitry-lahoda:
for this generation next parameters would be fruitfull https://docs.mongodb.com/manual/reference/operator/query/jsonSchema/#omissions

RicoSuter:
Missing $ref is bad...

src/NSwag.Commands/Commands/SwaggerGeneration/TypesToJsonSchemaCommand.cs
```diff
@@ -0,0 +1,153 @@
+﻿using System.Collections.Generic;
```
RicoSuter: Move this to the /Generation directory

src/NSwag.Commands/Commands/SwaggerGeneration/TypesToJsonSchemaCommand.cs
```diff
@@ -0,0 +1,153 @@
...
+            {
+                Reference = document.Definitions.First().Value
+            });
+
+            SetAllowAdditionalProperties(document);
+
+            return document.ToJson();
+        }
+
+        private void SetAllowAdditionalProperties(JsonSchema4 document)
```
RicoSuter: This should be an option in the JSON Schema generator itself => create a PR in NJsonSchema

dzmitry-lahoda:
> 
> 
> Missing $ref is bad...

Is there parameter to totally flat out schema to avoid refs? In NJsonSchema? Or this also should be implemented?

src/NSwag.Commands/Commands/SwaggerGeneration/TypesToJsonSchemaCommand.cs
dzmitry-lahoda: did pr, this pull after that pull

src/NSwag.Commands/Commands/SwaggerGeneration/TypesToJsonSchemaCommand.cs
dzmitry-lahoda: will tackle in single pass as soon as https://github.com/RicoSuter/NJsonSchema/pull/981 and may be other pull (for flat ref) considered. thanks

RicoSuter:
> Is there parameter to totally flat out schema to avoid refs(kind on inlining)? In NJsonSchema? Or this also should be implemented?

There is no such option yet, mainly because it would end in an endless recursion if there are cyclic references :-) 

We'd need to add a special setting for that too and throw exceptions if its not possible

src/NSwag.Commands/Commands/SwaggerGeneration/TypesToJsonSchemaCommand.cs
```diff
@@ -0,0 +1,153 @@
...
+                .SelectMany(n => PathUtilities.FindWildcardMatches(n, allExportedClassNames, '.'))
+                .Distinct();
+
+            foreach (var className in matchedClassNames)
+            {
+                var type = assemblies.Select(a => a.GetType(className)).FirstOrDefault(t => t != null);
+                await generator.GenerateAsync(type, schemaResolver).ConfigureAwait(false);
+            }
+
+            document.AllOf.Add(new JsonSchema4
```
RicoSuter: Why allOf? You should be able to either define the root type or generate all types into “definitions”

RicoSuter:
> I have generate definition, but I also need somehow specify kind of root class after all definitions.

You'd first call schemaGenerator.Generate(schema, type, ...) to generate the given type as the root into the schema and then loop over all types as it is done now.

RicoSuter:
I've moved the class and merged master into your branch so that it compiles again.

src/NSwag.Commands/Commands/SwaggerGeneration/TypesToJsonSchemaCommand.cs
dzmitry-lahoda: > `define the root` 
Does it mean that root definition properties-type from current `allOf` should inlin? I.e. there should be no definition for root object? What if we have 2 roots? `oneOf`? So what should I do? May be example? I use online json validator to check sanity of generated stuff.

> generate all types into “definitions”
Do not get it. I need at least one root type inlined or `allOf` or `oneOf` to get validation. What is example of this?

dzmitry-lahoda:
> I've moved the class and merged master into your branch so that it compiles again.

So I rebase on it? Right?

> There is no such option yet, mainly because it would end in an endless recursion if there are cyclic references :-)
> We'd need to add a special setting for that too and throw exceptions if its not possible

Should it be `NJsonSchema`  ? I would try that here either than. 

Than return into `NSwag` to allow generation scenarios in my case (complex game config with hierarchical relations and polymorphism, validate in vscode and altova and runtime) and ability to cross validate and store that in `mongodb`. I.e. end to end cycle from editing json to validating to storing in DB.

RicoSuter:
Because these are JSON Schema only features they should be in NJsonSchema and only the CLI support is here (the command class with the settings).

RicoSuter:
> So I rebase on it? Right?

No, just continue working on this branch

> Should it be NJsonSchema ? I would try that here either than.

For this we'd need a seting for the JsonSchemaGenerator so that the GenerateWithRefernce methods do not generate a ref but inline or throw if reused... something like this.

RicoSuter:
Add

```


        [Argument(Name = "AnyType", IsRequired = false, Description = "The any .NET type (default: 'object').")]
        public string AnyType
        {
            get { return Settings.CSharpGeneratorSettings.AnyType; }
            set { Settings.CSharpGeneratorSettings.AnyType = value; }
        }
```

RicoSuter:
We also need to add InlineNamedAny to the command and other InlineNamed* settings

dzmitry-lahoda:
I will try to inline in NJsonSchema, thinking within next 2 weeks, and will return to this pull than.

RicoSuter:
You can use switcher.json with
https://github.com/RicoSuter/DNT/blob/master/README.md#switch-to-projects

dzmitry-lahoda:
sorry for delay, we will have release soon and i had to switch into devops. so may be after release i may return to proper schema for our game mechanics config.

PR #2195, opened 2019-05-30
Roll forward on console tools

ryanbrandenburg:
This allows the nswag tool to roll forward on to higher major versions, making it usable out of the box for 3.0 preview builds.

### DIFF
--- a/src/NSwag.ConsoleCore/runtimeconfig.template.json
+++ b/src/NSwag.ConsoleCore/runtimeconfig.template.json
@@ -1,3 +1,4 @@
 {
-  "applyPatches": true
+  "applyPatches": true,
+  "rollForwardOnNoCandidateFx": 2
 }


RicoSuter:
So with this the binary will always run on the latest installed version? Will this be enough as the used asp.net core version will still be v2.1 or v2.2?

ryanbrandenburg:
> So with this the binary will always run on the latest installed version?

Kind of. More specifically it will roll forward onto the next higher version. So if you're requesting 2.1.0 and available frameworks are `["2.2.2", "2.2.4", "3.0.0-preview4"]` it will select "2.2.4", but if you only have `["3.0.0-preview4", "3.0.0-preview6"]` it will select "3.0.0-preview6". The docs defining this behavior [are here](https://github.com/dotnet/core-setup/blob/master/Documentation/design-docs/roll-forward-on-no-candidate-fx.md).

For your second question I'm not entirely sure I follow, but if you're asking if this change will allow a `netcoreapp2.1` project to run if only 3.0.0-previewX is installed the answer is yes, I tested it locally.

RicoSuter:
> For your second question I'm not entirely sure I follow, but if you're asking if this change will allow a netcoreapp2.1 project to run if only 3.0.0-previewX is installed the answer is yes, I tested it locally.

The .net core 2.1 binaries are compiled against the latest ASP.NET Core 2.1 version... if the runtime is changed to eg .net core 3.0 it will still run against ASP.NET Core 2.1... 

Hope this does not break users, this whole assembly loading stuff is extremely brittle with all the possible version combinations, frameworks, runtimes, etc...

We could just add a .NET Core 3.0 binary - the only reason why it's not added yet is because it is in preview and it would mean that people who clone the repo also need this preview to build the solution...

ryanbrandenburg:
Good points! @dougbu is taking a stab cross-compiling this tool for netcoreapp3.0 since he's already making other changes, so I'll close this out.

dougbu:
Works fine 😺 for code generation scenarios ❕

For example, in a 3.0-only environment and using a 3.0 solution:
```
  GenerateNSwagCSharp:
    dotnet --roll-forward-on-no-candidate-fx 2 ...\.nuget\packages\nswag.msbuild\12.5.10\build\../tools/NetCore21//dotnet-nswag.dll swagger2csclient /className:WebApplication3_0_FredClient /namespace:ConsoleClient /input:...\WebApplication3.0\WebApplication3.0\obj\WebApplication3.0_Fred.json /output:obj\WebApplication3.0_FredClient.cs
  NSwag command line tool for .NET Core NetCore21, toolchain v12.5.10.0 (NJsonSchema v10.0.5.0 (Newtonsoft.Json v11.0.0.0))
  Visit http://NSwag.org for more information.
  NSwag bin directory: ...\.nuget\packages\nswag.msbuild\12.5.10\tools\NetCore21
  Code has been successfully generated but not written to file (no change detected).

  Duration: 00:00:01.1127311
```

@RicoSuter 🆗 w/ you if I include this command addition in a small update to the Microsoft.ApiDescription.Client project?

(I doubt commands that need an inside man and / or load versions of ASP.NET Core packages will work as well.)

RicoSuter:
@dougbu how would you do that? Wasn't this PR exactly that? Isn't there a way to only roll forward if the requested runtime is not available?

dougbu:
I'll add the `--roll-forward-on-no-candidate-fx 2` argument to the command executed in NSwag.ApiDescription.Client.targets. This won't make your command line work everywhere but will help in that specific scenario.

RicoSuter:
@dougbu fine for me if this is enough.. also this cannot break existing users and will work until we directly support core 3 as target framework directly

PR #2196, opened 2019-05-30
Automatically handle proxy routes, paths and headers

RicoSuter:
Closes https://github.com/RicoSuter/NSwag/issues/2194

Adds automatic support for

- X-Forwarded-Proto (HTTP scheme, http/https)
- X-Forwarded-Host
- X-Forwarded-Prefix (base path, Traefik)
- PathBase

When

- Generating the document's server URL (in the spec’s JSON)
- Generating the redirect and document URLs for the UIs

Btw: I’m setting OpenAPI 3’s server url which is automatically mapped to Swagger 2 properties


RicoSuter:
I've pushed this preview to NuGet.org so that you can easily try it out in your environment: 

Package: NSwag.AspNetCore
Version: 13.0.0-20190531.5 (Prerelease)

gameboyLV:
it works well in iis and nginx,thank you.

OculiViridi:
I'm currently updating my APIs project to use the new NSwag package v13 (from last v12).

Can you please tell me if I still need this piece of configuration about proxy settings or how it should be changed to be compliant with the new v13 release?

```csharp
public static void UseSwaggerWithConfiguration(this IApplicationBuilder app)
{
    app.UseOpenApi(config =>
    {
        config.PostProcess = (document, request) =>
        {
            if (request.Headers.ContainsKey("X-Forwarded-Host"))
            {
                document.Host = request.Headers["X-Forwarded-Host"].First();
            }
        };
    });

    app.UseSwaggerUi3(config =>
    {
        // HACK: Temporary solution for working when behind reverse proxy
        config.TransformToExternalPath = (internalUiRoute, request) =>
        {
            // HACK: Redirect from \swagger to \swagger\index.html
            if (request.PathBase + request.Path == internalUiRoute)
                return internalUiRoute;

            // HACK: same as before for redirect
            var externalPath = request.Headers.ContainsKey("X-Forwarded-Path") ? request.Headers["X-Forwarded-Path"].First() : request.PathBase.Value;

            return externalPath + internalUiRoute;
        };

        config.SwaggerRoutes.Add(new SwaggerUi3Route("v1.0", "/swagger/v1.0/swagger.json"));
    });
}
```

dooferorg:
I've found these threads searching for reasons why NSwag (13.1.3) is not handling the URI of an nginx reverse proxy properly. Am I to understand that it's meant to handle it automatically? Is there some option to make it realize it'll also be accessed via a reverse proxy?

OculiViridi:
> I've found these threads searching for reasons why NSwag (13.1.3) is not handling the URI of an nginx reverse proxy properly. Am I to understand that it's meant to handle it automatically? Is there some option to make it realize it'll also be accessed via a reverse proxy?

@dooferorg Yes, NSwag is now able to handle it correctly and automatically. There's no need for specific configuration or workarounds, like the one I've used above with v12.

PR #2204, opened 2019-06-04
TS injection and baseUrl fixes

alexdresko:
* Allows ConfigurationClass to be injected.

    If a configuration class is specified, it has to be added to the `inject` list. Otherwise, Aurelia's DI will pass an empty string for the configuration class instance. For that matter, I'm really not sure how/why you'd want to inject a string with DI. 
* Handles empty baseUrl

   The string that is injected into the client is always empty... so the check I added here allows the base class to configure the base URL.

### DIFF
--- a/src/NSwag.CodeGeneration.TypeScript/Templates/FetchClient.liquid
+++ b/src/NSwag.CodeGeneration.TypeScript/Templates/FetchClient.liquid
@@ -22,7 +22,7 @@
 {%     endif -%}
         this.http = http ? http : <any>window;
 {%     if UseGetBaseUrlMethod -%}
-        this.baseUrl = baseUrl  && baseUrl !== "" ? baseUrl : this.getBaseUrl("{{ BaseUrl }}");
+        this.baseUrl = this.getBaseUrl("{{ BaseUrl }}", baseUrl);
 {%     else -%}
         this.baseUrl = baseUrl ? baseUrl : "{{ BaseUrl }}";
 {%     endif -%}
--- a/src/NSwag.CodeGeneration.TypeScript/Templates/FetchClient.liquid
+++ b/src/NSwag.CodeGeneration.TypeScript/Templates/FetchClient.liquid
@@ -8,7 +8,7 @@
 {% endif -%}
 
 {% if UseAureliaHttpInjection -%}
-@inject(String, HttpClient)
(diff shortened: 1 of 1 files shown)


src/NSwag.CodeGeneration.TypeScript/Templates/FetchClient.liquid
```diff
@@ -22,7 +22,7 @@
 {%     endif -%}
         this.http = http ? http : <any>window;
 {%     if UseGetBaseUrlMethod -%}
-        this.baseUrl = baseUrl ? baseUrl : this.getBaseUrl("{{ BaseUrl }}");
+        this.baseUrl = baseUrl  && baseUrl != "" ? baseUrl : this.getBaseUrl("{{ BaseUrl }}");
```
RicoSuter: !==?

alexdresko: Done.

RicoSuter:
Thanks for the PR. Is the PR ready now?

alexdresko:
Happy to help. PR seems ready. I'm using the code here and it works great.

src/NSwag.CodeGeneration.TypeScript/Templates/FetchClient.liquid
```diff
@@ -22,7 +22,7 @@
 {%     endif -%}
         this.http = http ? http : <any>window;
 {%     if UseGetBaseUrlMethod -%}
-        this.baseUrl = baseUrl ? baseUrl : this.getBaseUrl("{{ BaseUrl }}");
+        this.baseUrl = baseUrl  && baseUrl !== "" ? baseUrl : this.getBaseUrl("{{ BaseUrl }}");
```
RicoSuter: I'm a bit worried about this someone might want to provide "" as the value to use and not call getBaseUrl() in this case...

Maybe we should replace the whole line with

```
this.baseUrl = this.getBaseUrl("{{ BaseUrl }}", baseUrl);
```

So that the getBaseUrl() has the complete control over what baseUrl to use? 

This is a breaking change but gives much more flexibility and gives you a build error which can easily be fixed...

What do you think?

RicoSuter: Ok, I've changed the template.

PR #2214, opened 2019-06-06
 Execute `dotnet-nswag.dll` with `--roll-forward-on-no-candidate-fx 2`

dougbu:
- #2195

nits:
- use new name of `swagger2csclient` command
- ignore _all_ .nupkg files

### DIFF
--- a/src/NSwag.ApiDescription.Client/NSwag.ApiDescription.Client.targets
+++ b/src/NSwag.ApiDescription.Client/NSwag.ApiDescription.Client.targets
@@ -2,7 +2,8 @@
 <Project>
   <PropertyGroup>
     <_NSwagCommand>$(NSwagExe)</_NSwagCommand>
-    <_NSwagCommand Condition="'$(MSBuildRuntimeType)' == 'Core'">$(NSwagExe_Core21)</_NSwagCommand>
+    <_NSwagCommand
+        Condition="'$(MSBuildRuntimeType)' == 'Core'">dotnet --roll-forward-on-no-candidate-fx 2 $(NSwagDir_Core21)/dotnet-nswag.dll</_NSwagCommand>
   </PropertyGroup>
 
   <!-- OpenApiReference support for C# -->
--- a/src/NSwag.ApiDescription.Client/NSwag.ApiDescription.Client.targets
+++ b/src/NSwag.ApiDescription.Client/NSwag.ApiDescription.Client.targets
@@ -12,7 +12,7 @@
     <ItemGroup>
       <!-- @(CurrentOpenApiReference) item group will never contain more than one item. -->
       <CurrentOpenApiReference>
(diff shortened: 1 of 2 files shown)


dougbu:
@RicoSuter I realize #2195 is closed but this at least addresses part of the 3.0 compat issue

src/NSwag.ApiDescription.Client/NSwag.ApiDescription.Client.targets
```diff
@@ -2,7 +2,8 @@
 <Project>
   <PropertyGroup>
     <_NSwagCommand>$(NSwagExe)</_NSwagCommand>
-    <_NSwagCommand Condition="'$(MSBuildRuntimeType)' == 'Core'">$(NSwagExe_Core21)</_NSwagCommand>
+    <_NSwagCommand
+        Condition="'$(MSBuildRuntimeType)' == 'Core'">dotnet --roll-forward-on-no-candidate-fx 2 $(NSwagDir_Core21)/dotnet-nswag.dll</_NSwagCommand>
```
RicoSuter: Why arent we checking other fw versions like net core 2.2?

RicoSuter:
So this should be enough for now?

src/NSwag.ApiDescription.Client/NSwag.ApiDescription.Client.targets
dougbu: This code uses the MSBuild runtime environment to guess whether _any_ version of `dotnet` is installed. Figuring out the version when the tool command of interest works with 2.1 or later installed is extra work for minimal benefit.

dougbu:
> So this should be enough for now?

Yes, it works in a 3.0-only environment just fine.

RicoSuter:
Ok then - v13 will be released in the next days... need to update the wiki, do more tests etc. 
Is there documentation about how to use this service reference stuff, the supported scenarios, etc.? I'd like to verify whether all important things are covered, if something is missing etc...

RicoSuter:
Thanks for the PR!

dougbu:
> Is there documentation about how to use this service reference stuff, the supported scenarios, etc.? I'd like to verify whether all important things are covered, if something is missing etc...

Not yet though it's definitely due. This lack exists in part because preview 6 isn't released yet and those bits contain the Web project side of things. The following is available:
- https://github.com/aspnet/AspNetCore/blob/master/src/Mvc/Extensions.ApiDescription.Client/src/build/Microsoft.Extensions.ApiDescription.Client.props contains some docs for the user-visible properties and items usable in "client" projects
- https://github.com/aspnet/AspNetCore/blob/master/src/Mvc/Extensions.ApiDescription.Server/src/build/Microsoft.Extensions.ApiDescription.Server.props contains some docs for the user-visible properties usable in "server" (Web) projects
  - These features will work end-to-end after we release 3.0 preview 6 and you publish your next release after my next PR (coming in the next few days)

PR #2215, opened 2019-06-07
Change default of DefaultResponseReferenceTypeNullHandling to NotNull

RicoSuter:
This PR:

- Changes the default of `DefaultResponseReferenceTypeNullHandling` from `Null` to `NotNull`.
- Adds support for the "nullable" xml docs attribute

Details: 

- Currently there is no way to specify nullability of the response with `ProducesResponseTypeAttribute`. This could be solved with a new "nullable" attribute in the XML docs (which always wins if specified):

```xml
/// <summary>
/// Gets an order.
/// </summary>
/// <response code="200" nullable="true">Order created.</response>
...
```

- Most of the time (I assume?) people want a not nullable response and use 404 in the not found case and do not return null
- I think Swagger UI 3 still cannot handle nullable responses (spec in the first example)
- In the Swagger 2 the difference is only in the `x-nullable` property and thus the impact is lower.

Sample operation: 

```csharp
[HttpGet("{petId}", Name = "FindPetById")]
[ProducesResponseType(typeof(Pet), StatusCodes.Status200OK)]
[ProducesResponseType(typeof(SerializableError), StatusCodes.Status400BadRequest)]
[ProducesResponseType(StatusCodes.Status404NotFound)]
public async Task<ActionResult<Pet>> FindById(int petId)
```

With `DefaultResponseReferenceTypeNullHandling = Null` (current default, OpenAPI 3)

```
    "/pet/{petId}": {
      "get": {
        "operationId": "Pet_FindById",
        "parameters": [
          {
          }
        ],
        "responses": {
          "200": {
            "description": "",
            "content": {
              "application/json": {
                "schema": {
                  "nullable": true,
                  "oneOf": [
                    {
                      "$ref": "#/components/schemas/Pet"
                    }
                  ]
                }
              }
            }
          },
```

With `DefaultResponseReferenceTypeNullHandling = NotNull` (new default, OpenAPI 3)

```json
    "/pet/{petId}": {
      "get": {
        "operationId": "Pet_FindById",
        "parameters": [
          {
          }
        ],
        "responses": {
          "200": {
            "description": "",
            "content": {
              "application/json": {
                "schema": {
                  "$ref": "#/components/schemas/Pet"
                }
              }
            }
          },
```

Open questions:

- **Should we change the default?**
    - Pro: Most devs probably already design their APIs so that they don't return null and thus expect the default to be NotNull
    - Con: NotNull does not really reflect what the actual API is allowed to return (i.e. it is a missmatch between the spec and the runtime)

/cc @pranavkm @dougbu What's your take on this?

### DIFF
--- a/samples/WithoutMiddleware/Sample.AspNetCore20/nswag_project.nswag
+++ b/samples/WithoutMiddleware/Sample.AspNetCore20/nswag_project.nswag
@@ -15,7 +15,7 @@
       "apiGroupNames": null,
       "defaultPropertyNameHandling": "Default",
       "defaultReferenceTypeNullHandling": "Null",
-      "defaultResponseReferenceTypeNullHandling": "Null",
+      "defaultResponseReferenceTypeNullHandling": "NotNull",
       "defaultEnumHandling": "Integer",
       "flattenInheritanceHierarchy": false,
       "generateKnownTypes": true,
--- a/samples/WithoutMiddleware/Sample.AspNetCore20/nswag_project_swagger.json
+++ b/samples/WithoutMiddleware/Sample.AspNetCore20/nswag_project_swagger.json
@@ -19,7 +19,7 @@
         ],
         "responses": {
           "200": {
-            "x-nullable": true,
+            "x-nullable": false,
             "description": "",
             "schema": {
               "type": "array",
@@ -82,7 +82,7 @@
         ],
         "responses": {
           "200": {
-            "x-nullable": true,
+            "x-nullable": false,
             "description": "",
             "schema": {
               "type": "string"
--- a/samples/WithoutMiddleware/Sample.AspNetCore21/nswag_assembly.nswag
+++ b/samples/WithoutMiddleware/Sample.AspNetCore21/nswag_assembly.nswag
@@ -15,7 +15,7 @@
       "apiGroupNames": null,
       "defaultPropertyNameHandling": "Default",
       "defaultReferenceTypeNullHandling": "Null",
-      "defaultResponseReferenceTypeNullHandling": "Null",
+      "defaultResponseReferenceTypeNullHandling": "NotNull",
       "defaultEnumHandling": "Integer",
       "flattenInheritanceHierarchy": false,
       "generateKnownTypes": true,
--- a/samples/WithoutMiddleware/Sample.AspNetCore21/nswag_assembly_swagger.json
+++ b/samples/WithoutMiddleware/Sample.AspNetCore21/nswag_assembly_swagger.json
@@ -19,7 +19,7 @@
         ],
         "responses": {
           "200": {
-            "x-nullable": true,
+            "x-nullable": false,
             "description": "",
             "schema": {
               "type": "array",
@@ -82,7 +82,7 @@
         ],
         "responses": {
           "200": {
-            "x-nullable": true,
+            "x-nullable": false,
             "description": "",
             "schema": {
               "type": "string"
--- a/samples/WithoutMiddleware/Sample.AspNetCore21/nswag_project.nswag
+++ b/samples/WithoutMiddleware/Sample.AspNetCore21/nswag_project.nswag
@@ -15,7 +15,7 @@
       "apiGroupNames": null,
       "defaultPropertyNameHandling": "Default",
       "defaultReferenceTypeNullHandling": "Null",
-      "defaultResponseReferenceTypeNullHandling": "Null",
+      "defaultResponseReferenceTypeNullHandling": "NotNull",
       "defaultEnumHandling": "Integer",
       "flattenInheritanceHierarchy": false,
       "generateKnownTypes": true,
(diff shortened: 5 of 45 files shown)


RicoSuter:
> How does that impact the generated C#/TS code?

- CSharp does not change because reference types cannot be made non nullable
- TypeScript will change from `Person | null` as response to `Person` 

You can still change this default back to how it is now

RicoSuter:
I see another problem in the future: Even with Nullable Reference Types you cannot describe nullability with the ProduceResponseTypeAttribute, because this `[ProducesResponseType(typeof(Pet?), StatusCodes.Status200OK)]` is not allowed (because it's just a Type object with no context and there is no place to generate the NullableAttribute). 

Maybe it would be best to also add also an `IsNullable` attribute to the `ProducesResponseTypeAttribute`? /cc @rynowak @dougbu @pranavkm

RicoSuter:
@StasPerekrestov good to know... what’s your opinion on the default?

dougbu:
> Most of the time (I assume?) people want a not nullable response and use 404 in the not found case and do not return null

I don't know the whole history here but agree with this sentiment. Provide a well-documented override of this default and all would be well from my perspective.

> you cannot describe nullability with the `ProduceResponseTypeAttribute`

Your point is valid but leave nullable reference types out of it: Knowing the compiler will allow a `null` assignment is very different from knowing the user _will_ assign `null` in some cases.

Agree a `ReturnsNull` property or similar may be needed if we find users return `null` in a few but not all actions.

RicoSuter:
> Your point is valid but leave nullable reference types out of it: Knowing the compiler will allow a null assignment is very different from knowing the user will assign null in some cases.

The problem here is the inconsistency from a user perspective: As a user in an NRT world, I'd expect that `typeof(Pet)` is automatically treated as not nullable and `typeof(Pet?)` is nullable (but this probably does not compile, right?), but from a tooling perspective we cannot assume not nullable because `typeof(Pet?)` is not allowed. As a user I'd like to express nullability in the same way as I do it everywhere else (e.g. parameters) - not  with an additional property on the attribute...

StasPerekrestov:
I guess, you are looking for something like Optional<T>/Maybe to avoid
using attributes, however usage of these patterns isn't common in c#
world...

RicoSuter:
> I guess, you are looking for something like Optional<T>/Maybe to avoid
> using attributes, however usage of these patterns isn't common in c#
> world...

Or `Nullable<T>` but that's only for value types - however as a user I just expect that `Pet` vs `Pet?` just works everywhere and thus I'm a bit worried that people will be quite confused.

StasPerekrestov:
On one hand we have the issue described in the task,
On the other hand there is an issue related to reference types handling,
so, basically, we have two related issues.
In c# world we don't have to many options to go with. We need either to
wrap instances, e.g. Maybe monad, but c# doesn't have appropriate sugar
syntax to deal with it, or use Attributes to express the disired behavior.

RicoSuter:
> In c# world we don't have to many options to go with. We need either to
> wrap instances, e.g. Maybe monad, but c# doesn't have appropriate sugar
> syntax to deal with it, or use Attributes to express the disired behavior.

My point is that this should have been solved with the C# 8 NRT feature somehow, but `typeof()` cannot "store" this nullability info...

However this PR is not really about NRT or C# 8 but about changing the default of DefaultResponseReferenceTypeNullHandling to NotNull and adding the "nullable" attribute to the xml docs (this will solve the problems for now). **My big question is: What is the best default for DefaultResponseReferenceTypeNullHandling?**

StasPerekrestov:
What concerns me, is that XML docs now change the behavior of XML schema generation. This is something, not many developers are aware of and also may be a source of confusion.

From my point of view, I'd rather look for an appropriate .net attribute to change the required behavior than for an attribute of XML doc.

StasPerekrestov:
Regarding the new defaults, I'm OK with that. However, the information that describes the behavior and the way of changing it should be more easy to find.

RicoSuter:
> What concerns me, is that XML docs now change the behavior of XML schema generation. This is something, not many developers are aware of and also may be a source of confusion.

Yes, somehow you are right, but the alternative is even more verbose: 

```
/// <summary>
/// Gets an order.
/// </summary>
/// <response code="200">The order.</response>
[HttpGet("{orderId}", Name = "GetOrder")]
[ProducesResponseType(typeof(Pet), StatusCodes.Status200OK)]
[ProducesResponseTypeNullable(true, StatusCodes.Status200OK)]
public async Task<ActionResult<Order>> GetOrder(int orderId)
```

I think it's the still the nicest way to do as xml docs attribute as it is the only way to avoid having specifying a single response code three times...

Ideally there's just a `IsNullable` (or similar) property on the `ProducesResponseTypeAttribute` but for backward compat. we'd still keep the "nullable" xml docs attribute.

RicoSuter:
Docs: https://github.com/RicoSuter/NSwag/wiki/AspNetCoreOpenApiDocumentGenerator#response-nullability

PR #2221, opened 2019-06-08
"PreservePropertyNameCasing" parameter

john-larson:
"PreservePropertyNameCasing" parameter added for giving the option to not automatically UpperCamelCasing for C# and camelCasing for TypeScript models in code generation.


john-larson:
@RicoSuter Could you review the pull request please?

RicoSuter:
The problem with this PR is that there is already an extension point (the CodeGeneratorSettings.PropertyNameGenerator) and we are wrapping it in the same abstraction level...

Maybe we should not implement this in CSharpGeneratorBaseSettings.cs but in the command with a PropertyNameGeneratorType property which actually instantiates the given full type name. Later it is planned that you can also load your custom DLLs and load generators: 
https://github.com/RicoSuter/NSwag/issues/924

anderejd:
This looks great from my perspective, I'd like a simple setting, not a plugin system.

Later perhaps multiple built-in modes, something like:
 - PreserveNames
 - camelCase
 - PascalCase
 - snake_case

PR #2226, opened 2019-06-10
Use new Microsoft.Extensions.ApiDescription.Server package

dougbu:
- provides document download on build and aligns with Microsoft.Extensions.ApiDescription.Client p2p references
  - this feature is on by default
  - can be disabled using `<OpenApiRetrieveDocuments>false</OpenApiRetrieveDocuments>`
  - timing can be changed using `<OpenApiRetrieveDocumentsAfterBuild>false</OpenApiRetrieveDocumentsAfterBuild>`
    - then, depend on the `RetrieveOpenApiDocuments` target from somewhere else
- add `GetDocumentNames()` to `IDocumentProvider` and `AspNetCoreOpenApiDocumentGenerator`
- rename `IDocumentProvider` namespace Microsoft.Extensions.ApiDescription -> Microsoft.Extensions.ApiDescriptions
  - avoid conflict with `Microsoft.AspNetCore.Mvc.ApiExplorer.ApiDescription` type
- depend on Microsoft.Extensions.ApiDescription.Server package from NSwag.AspNetCore package
  - restore and update the previously-unused (then deleted) NSwag.AspNetCore.nuspec file

nits:
- remove and sort `using`s
- take VS suggestions in `NSwagServiceCollectionExtensions`
- add missing `NSwagServiceCollectionExtensions` copyright header
- correct spelling in `AspNetCoreOpenApiDocumentGenerator`


src/NSwag.AspNetCore/NSwag.AspNetCore.nuspec
```diff
@@ -0,0 +1,61 @@
+<?xml version="1.0"?>
+<package>
+    <metadata>
+        <id>$id$</id>
+        <version>$version$</version>
+        <authors>$authors$</authors>
+        <description>$description$</description>
+        <tags>Swagger Documentation WebApi AspNet TypeScript CodeGen</tags>
+        <projectUrl>https://github.com/NSwag/NSwag</projectUrl>
+        <licenseUrl>https://github.com/NSwag/NSwag/blob/master/LICENSE.md</licenseUrl>
```
dougbu: @RicoSuter does it matter (much) that I'm using old GitHub locations for this repo i.e. should `projectUrl` be https://github.com/RicoSuter/NSwag ?

dougbu:
@RicoSuter please let me know if this needs to change at all.

FYI our 3.0 Preview 6 release should be soon.

RicoSuter:
@dougbu does this have to be part of the v13.0.0 release? We can push this later, right?

dougbu:
@RicoSuter this will only be WIP for a few more days (I think) and I'd really like to see it in your next release.

RicoSuter:
@dougbu is it ok to do a patch release (v13.0.1) with this in a few days? I really want to get this version out :-)

dougbu:
Sure❕ I wasn't directing, just asking 😺

RicoSuter:
Do we really need to use a nuspec file? I dont like this as it’s quite bit harder to maintain deps etc..

dougbu:
@RicoSuter it's not possible to add a dev dependency to a shipped package without the .nuspec file. W.r.t. maintaining dependencies, I tried to minimize the issues by keeping all of the versions in the .csproj file.

RicoSuter:
> provides document download on build and aligns with Microsoft.Extensions.ApiDescription.Client p2p references

Does this mean that the spec is generated on build by default when the NSwag.AspNetCore package is installed. 
This might be a bit dangerous to enable by default because the generation might take quite some time for bigger apis.

> rename IDocumentProvider namespace Microsoft.Extensions.ApiDescription -> Microsoft.Extensions.ApiDescriptions

Does this also work when multiple IDocumentProviders are registered, e.g. also from Swashbuckle or another document generator like GRPC or AsyncApi?

> depend on Microsoft.Extensions.ApiDescription.Server package from NSwag.AspNetCore package

Isn't this a breaking change, e.g. for "older" target frameworks?

src/NSwag.AspNetCore/NSwag.AspNetCore.nuspec
RicoSuter: We should use the correct URL, why using the old one?

dougbu: This matches what I see in your released packages. I'll update.

dougbu:
> Does this mean that the spec is generated on build by default when the NSwag.AspNetCore package is installed.

Yes, it's on by default.

> This might be a bit dangerous to enable by default because the generation might take quite some time for bigger apis.

Can be disabled using `<OpenApiGenerateDocuments>false</OpenApiGenerateDocuments>`.

> Does this also work when multiple IDocumentProviders are registered, e.g. also from Swashbuckle or another document generator like GRPC or AsyncApi?

`IDocumentProvider`s are limited to OpenAPI document providers i.e. will only available in the NSwag.AspNetCore and Swashbuckle.AspNetCore packages. If a project uses both of those packages, the `dotnet-getdocument` tool in Microsoft.Extensions.ApiDescription.Server will choose the first assembly it finds in the application containing an `IDocumentProvider` definition. This is silent because it seems to be a test-only scenario.

> Isn't this a breaking change, e.g. for "older" target frameworks?

Not really. The `dotnet-getdocument` tool supports all supported .NET Core versions (2.1 and greater). If users choose to use the latest NSwag.AspNetCore in older projects, they also can use `<OpenApiGenerateDocuments>false</OpenApiGenerateDocuments>`. I'll auto-disable this feature for downlevel releases in our next 3.0 preview to save them the minor trouble. Will also document this property for exactly this use case in the meantime.

dougbu:
3.0 Preview 6 packages are now public. This PR can be safely merged, perhaps with a release note mentioning `<OpenApiRetrieveDocuments>false</OpenApiRetrieveDocuments>` for .NET Core 2.0 and earlier web projects using the latest NSwag.AspNetCore

RicoSuter:
If I only install nswag.aspnetcore, will it still auto gen the specs even if they are not used anywhere? Or what is this feature for?

dougbu:
Yes, it will get the documents and save them in obj/ by default. The feature is for both API compatibility checks (when default doc location is overridden and user chooses to check the files in) and consumption in client projects using `@(OpenApiProjectReference)` items.

dougbu:
:up::date: to rebase on 'master' and change all GitHub URLs to avoid the extra redirects

RicoSuter:
> Yes, it will get the documents and save them in obj/ by default

I don't like this as it means that I still pay the performance impact even if I'm only interested in the HTTP middlewares (specs + web UI) and do not store the documents in the filesystem or use any `@(OpenApiProjectReference)`... Can't we disable it by default and only enable it if the output path is set (i.e. it should end up in the filesystem) or if a project uses `@(OpenApiProjectReference)` (probably harder to find out)? 

> The feature is for both API compatibility checks (when default doc location is overridden and user chooses to check the files in)

Compatibility checks = doing a diff in the PR?

dougbu:
> Can't we disable it by default and only enable it if the output path is set (i.e. it should end up in the filesystem) or if a project uses @(OpenApiProjectReference) (probably harder to find out)?

That's not viable because `@(OpenApiProjectReference)` will silently fail if the feature is opt-in. I could add an error if the feature is disabled when the client project calls the `OpenApiGetDocuments` target but that won't happen until preview 7 is released and feels user-violent.

Users concerned about performance or just don't want the docs can use `<OpenApiGenerateDocuments>false</OpenApiGenerateDocuments>`. They can also use `<OpenApiGenerateDocumentsOnBuild>false</OpenApiGenerateDocumentsOnBuild>` if they only want to update the documents explicitly or as part of some other target.

> Compatibility checks = doing a diff in the PR?

Yes

RicoSuter:
> ... and feels user-violent.

For me it feels user-violent that when you install a package like NSwag.AspNetCore the build immediately takes much longer (I have big APIs which take > 20s) and you are not even using a feature of it. I think 80% of the users only want the Swagger UI + spec server (only generated on demand when accessing /swagger) and will not use spec generation to file or service project references... With this you immediately get the penalty (because no user will know of this new option) and the feature is not even usable...

Wouldnt it be better if we create a new package, e.g. NSwag.AspNetCore.Build (or whatever?) which has this enabled etc. and which references the original NSwag.AspNetCore package?

dougbu:
The idea is to enable the document generation by default exactly so that code generation works end-to-end. There’s no way to know if there’s another project depending on this one until `OpenApiGetDocuments` is invoked.


  *   the feature is not even usable
What do you mean? With the changes in this PR, everything works end-to-end.

RicoSuter:
NSwag.AspNetCore will then depend on a preview package, isn't this a problem?

dougbu:
> NSwag.AspNetCore will then depend on a preview package, isn't this a problem?

No more than NSwag.ApiDescription.Client depending on a preview package -- just leads to another warning in your build that can be ignored. In any case, the feature is code compete at this point though it has a couple of rough edges we'll fix in 3.0 preview 7.

@glennc and I'll discuss moving from 0.3.0 to 3.0.0 for these packages in preview 7 (which might be closer to release candidate).

RicoSuter:
I just want to ensure that this does not break existing users - not everyone is using the latest versions of .NET Core 3...

dougbu:
No, it won't break existing users, especially because we're switching the feature to be opt-in as you requested. Will update this PR to that end soon.

RicoSuter:
> especially because we're switching the feature to be opt-in as you requested.

This all is preview at the moment anyway - so enabling it by default should be discussed later I think...

RicoSuter:
I’ve changed the versions to v13.0.2 and as soon as you updated the pr i’ll merge and release it.

dougbu:
Thanks @RicoSuter! I've been laid up since yesterday and might not get to this 'til early next week.

dougbu:
:up::date: to rebase on 'master' (removing conflicts) and change doc generation to opt-in

RicoSuter:
This pr might interfere with https://github.com/RicoSuter/NSwag/pull/2261

dougbu:
@RicoSuter yes it will conflict. If you and @mkArtakMSFT like, I could add `<IncludeOpenAPIAnalyzers>true</IncludeOpenAPIAnalyzers>` in this PR. The alternative is to pare #2261 down to just that addition.

mkArtakMSFT:
We can absolutely simplify this by including the flag in this PR.
@RicoSuter it's up to you!
I vote for simplicity.

dougbu:
@RicoSuter I added the property setting to this PR. I'll remove this last commit if you prefer to separate the concerns.

dougbu:
:up::date: again to split concerns more clearly into 3 commits

RicoSuter:
@dougbu Can you point me to the docs with the available msbuild tags, e.g. OpenApiRetrieveDocuments or OpenApiDocumentsDirectory - is there a ways to also specify the file name, e.g. rename it to openapi.json?

dougbu:
Document filenames are intentionally not controllable. They following the format `{project name}_{logical document name}.json` though `_{logical document name}` is left out for the common name "v1". This makes the filenames unique enough to avoid extra machinations in projects referencing the documents. It also maintains the `Startup` configuration as the source of truth for names.

Of course, this is all MSBuild. Users could for example leave the documents in `obj/` (a default that may be overridden) but use the `Copy` task to move them elsewhere and rename them for other purposes.

More generally, what is the main impetus for changing the document filenames (especially for files that land in `obj/` by default)?

----
Documentation is currently minimal. Most of the information is in [Microsoft.Extensions.ApiDescription.Server.props](https://github.com/aspnet/AspNetCore/blob/master/src/Mvc/Extensions.ApiDescription.Server/src/build/Microsoft.Extensions.ApiDescription.Server.props).

RicoSuter:
So there's no easy way to generate an openapi.json for a given document name (e.g. "v1") in the web project dir?

dougbu:
In a target that runs after `GenerateOpenApiDocuments`:
``` xml
<Copy SourceFiles="$(OpenApiDocumentsDirectory)/$(MSBuildProjectName).json"
    DestinationFiles="$(MSBuildProjectDirectory)openapi.json" />
```

PR #2241, opened 2019-06-14
Make Binary Body type configurable

plk:
Currently binary request bodies in Controller generation are hard-coded to "System.IO.Stream" which isn't that useful these days as most modern frameworks need `IFormFile` collections. This PR makes this type configurable.


RicoSuter:
What about instead of introducing a new setting we just use

```
            if (parameter.IsBinaryBodyParameter)
            {
                if (_settings is CSharpControllerGeneratorSettings controllerSettings)
                {
                    return controllerSettings.ControllerTarget == CSharpControllerTarget.AspNetCore ?
                        "Microsoft.AspNetCore.Http.IFormFile" :
                        "System.Web.HttpPostedFileBase";
                }
                else
                {
                    return "System.IO.Stream";
                }
            }
```

plk:
Could do but I would recommend a customisation here - if you look around, there are lots of questions with people asking how to do file upload in the various ASP flavours and there are many suggestions using all sorts of types to do it. I wouldn't restrict it to anything hard-coded and just let people choose but keep the current default for backwards compat?

However, could be improved by having a conditional to check if the param is multi to therefore allow e.g.: `ICollection<IFormFile>` or just 'IFormFile` but I couldn't see how to detect this.

RicoSuter:
We'd need a special CSharpTypeResolver for controllers here: 

https://github.com/RicoSuter/NJsonSchema/blob/fd6cc55f2c6ccccf66e2157397f50500c50866f2/src/NJsonSchema.CodeGeneration.CSharp/CSharpTypeResolver.cs#L185

Changing it there to IFormFile would also produce a collection of IFormFile if the param is of type array binary I think

plk:
Do you mean instead of your suggestion or in addition to it? I am a bit unsure what to do here as I can't see a way to make a resolver specially for controllers in NJsonSchema as "controller" isn't really a concept in that project is it? Could we just do it exactly the same as for `FileParameter` in `CSharpOperationMOdel.cs` where the multi is handled in the same place as opposed to in NJsonSchema?

plk:
Another thought - why don't we just use your suggestion but use a collection by default since can't hurt and it covers any number of files in the body?

```
            if (parameter.IsBinaryBodyParameter)
            {
                if (_settings is CSharpControllerGeneratorSettings controllerSettings)
                {
                    return controllerSettings.ControllerTarget == CSharpControllerTarget.AspNetCore ?
                        "System.Collections.Generic.ICollection<Microsoft.AspNetCore.Http.IFormFile>" :
                        "System.Collections.Generic.ICollection<System.Web.HttpPostedFileBase>";
                }
                else
                {
                    return "System.IO.Stream";
                }
            }
```

The files are usually access by filename in the collection anyway and so it's hardly worth having a seperate return for a single file.

RicoSuter:
I dont like to always generate a collection, but we could have a check for `parameter.Type == Array && parameter.Item.IsBinary` and generate the collection and then use `if (parameter.IsBinaryBodyParameter)` to just generate a single IFormFile

plk:
That's fine - I can't see however to detect the multi condition with these specifications - I tried the following but it always generates the non-multi type:

```
            if (parameter.IsBinaryBodyParameter)
            {
                if (_settings is CSharpControllerGeneratorSettings controllerSettings)
                {
                    
                    if (parameter.Type == JsonObjectType.Array && parameter.Item.IsBinary)
                    {
                        return controllerSettings.ControllerTarget == CSharpControllerTarget.AspNetCore ?
                        "System.Collections.Generic.ICollection<Microsoft.AspNetCore.Http.IFormFile>" :
                        "System.Collections.Generic.ICollection<System.Web.HttpPostedFileBase>";
                    }
                    else
                    {
                        return controllerSettings.ControllerTarget == CSharpControllerTarget.AspNetCore ?
                        "Microsoft.AspNetCore.Http.IFormFile" :
                        "System.Web.HttpPostedFileBase";
                    }
                }
                else
                {
                    return "System.IO.Stream";
                }
            }

```

RicoSuter:
I think IsBinaryBodyParameter only applies for single and the if array and isbinary must be outside

RicoSuter:
Also i think you need to use parameter.ActualSchema.Type to be safe

plk:
I see the issue now - the schema was not quite right due to the extra level of indirection. This works:

```
            if (parameter.IsBinaryBodyParameter)
            {
                if (_settings is CSharpControllerGeneratorSettings controllerSettings)
                {
                    
                    if (schema.Type == JsonObjectType.Array && schema.Item.IsBinary)
                    {
                        return controllerSettings.ControllerTarget == CSharpControllerTarget.AspNetCore ?
                        "System.Collections.Generic.ICollection<Microsoft.AspNetCore.Http.IFormFile>" :
                        "System.Collections.Generic.ICollection<System.Web.HttpPostedFileBase>";
                    }
                    else
                    {
                        return controllerSettings.ControllerTarget == CSharpControllerTarget.AspNetCore ?
                        "Microsoft.AspNetCore.Http.IFormFile" :
                        "System.Web.HttpPostedFileBase";
                    }
                }
                else
                {
                    return "System.IO.Stream";
                }
            }
```

with both:

```
      requestBody:
        content:
          multipart/form-data:
            schema:
                type: string
                format: binary
```

and 

```
      requestBody:
        content:
          multipart/form-data:
            schema:
              type: array
              items:
                type: string
                format: binary
```

I have also added a couple of tests for these schemata in `BinaryTests,cs`.

Shall I make a pull request?

RicoSuter:
Looks good. You can just update this branch - you dont need to create a new pr

plk:
I will see what happens - I may have removed the original clone so I don't know what will happen - may need a new PR ...

plk:
Apologies, I had deleted the previous fork and so couldnt' update - I opened PR  #2253 and will close this.

PR #2248, opened 2019-06-17
Deserialize XML body in CSharp client

### LINKED ISSUE #2082: Generate a Rest Client with XML Deserialization and not JSON
I am interfacing with an "external web API server" that returns only xml content-type. I have generated a rest client with NswagStudio but an exception is always thrown because of the response type which is XML and the deserialisation type which is JSON.

Is there any way to generate a client with XML Deserialisation?

RicoSuter:
Closes #2082

### DIFF
--- a/src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.ReadObjectResponse.liquid
+++ b/src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.ReadObjectResponse.liquid
@@ -2,37 +2,83 @@
 
 protected virtual async System.Threading.Tasks.Task<ObjectResponseResult<T>> ReadObjectResponseAsync<T>(System.Net.Http.HttpResponseMessage response, System.Collections.Generic.IReadOnlyDictionary<string, System.Collections.Generic.IEnumerable<string>> headers)
 {
+    if (response.Content == null)
+    {
+        return new ObjectResponseResult<T>(default(T), string.Empty);
+    }
+
     if (ReadResponseAsString)
     {
         var responseText = await response.Content.ReadAsStringAsync().ConfigureAwait(false);
-        try
+
+        if (response.Content.Headers.ContentType != null && response.Content.Headers.ContentType.MediaType.Contains("xml"))
         {
-            var typedBody = Newtonsoft.Json.JsonConvert.DeserializeObject<T>(responseText, JsonSerializerSettings);
-            return new ObjectResponseResult<T>(typedBody, responseText);
+            try
+            {
+                using (var reader = new System.IO.StringReader(responseText))
+                {
+                    var serializer = new System.Xml.Serialization.XmlSerializer(typeof(T));
+                    var typedBody = (T)serializer.Deserialize(reader);
+                    return new ObjectResponseResult<T>(typedBody, string.Empty);
+                }
+            }
+            catch (Exception exception) 
+            {
+                var message = "Could not deserialize the response body stream as " + typeof(T).FullName + ".";
+                throw new {{ ExceptionClass }}(message, (int)response.StatusCode, null, headers, exception);
+            }
         }
-        catch (Newtonsoft.Json.JsonException exception)
+        else
         {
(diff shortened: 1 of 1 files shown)


src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.ReadObjectResponse.liquid
```diff
@@ -2,37 +2,83 @@
...
+    {
+        return new ObjectResponseResult<T>(default(T), string.Empty);
+    }
+
     if (ReadResponseAsString)
     {
         var responseText = await response.Content.ReadAsStringAsync().ConfigureAwait(false);
-        try
+
+        if (response.Content.Headers.ContentType != null && response.Content.Headers.ContentType.MediaType.Contains("xml"))
```
RicoSuter: What should we check here?

emilw: I think it looks ok, the common is probably application/xml, but this covers text/xml also.
The idea here is only to support the basic 1-1 with C# class and XML?

emilw:
If we want to support the full flow, i realised that we do not need to include any external dependencies to the XML attributes etc. We could just generate the class based on the Swagger spec and replace these names.

On server side for C# it would look as:
``
public class Supplier
{
[XmlElement(ElementName = "SupplierIdentifier")]
public string SupplierId {get; set;}
}
``

Instead of recreating it with XMLElement as above we could just generate the class itself as:
``
public class Supplier
{
public string SupplierIdentifier {get; set;}
}
``

This would help us avoid referencing the XMLSerialization stuff from the generated class.

src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.ReadObjectResponse.liquid
Jehoel: If the `response.Content == null`, shouldn't the `ResponseText` be `null` too instead of `String.Empty`?

Jehoel: `response.Content.Headers.ContentType.MediaType.Contains("xml"))` - this is probably incorrect. The OOXML formats are XML files stored in a PKZIP container but have MIME types like `application/vnd.openxmlformats-officedocument.wordprocessingml.document`.

I think opting for a whitelist that users can override is better:

    public virtual bool IsXml( string mimeContentType ) {
        switch( mimeContentType ) {
        case "application/xml":
        case "application/xhtml+xml":
        case "text/xml":
            return true;
        default:
            //return false;
            return
                mimeContentType.EndsWith( "+xml", StringComparison.Ordinal )  ||
                mimeContentType.EndsWith( "/xml", StringComparison.Ordinal ); // risky?
        }
    }

src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.ReadObjectResponse.liquid
```diff
@@ -2,37 +2,83 @@
...
+                }
+            }
+            catch (Exception exception) 
+            {
+                var message = "Could not deserialize the response body stream as " + typeof(T).FullName + ".";
+                throw new {{ ExceptionClass }}(message, (int)response.StatusCode, null, headers, exception);
+            }
         }
-        catch (Newtonsoft.Json.JsonException exception)
+        else
```
Jehoel: If the code will check the content-type for XML responses, shouldn't it do the same for JSON responses?

Is there a way for a Swagger file to assert the response will be a JSON serialized object even if it's returned with an incorrect MIME type? (i.e. JSON served as `text/plain` or `text/html`?)

ruza-kavanka:
Hi guys, I face the same issue (cannot deserialize XML responses), when do you plan to resolve the conflicts and release please? I would appreciate this feature very much. Thanks a lot! RK.

Issue #2075, opened 2019-04-03
Generating ASP.NET Core Controllers, FileParameter vs. IFormFile


DrDeterminant:
In my OpenAPI 2.0 specification, I have the following portion of code for having a file upload available:
       consumes:
        - multipart/form-data
      parameters:
        - in: formData
          name: someFile
          type: file
          required: true
          description: Some file

The generated ASP.NET Core code for the controller contains the definition of a class "FileParameter" that is also passed in to the controller method as an argument. There two problems:
1) The given definition of the FileParameter class does not contain a parameterless constructor, which causes the model binding of ASP.NET to fail. If I add such a constructor:
2) The model binding does not seem to function properly in this case, the method receives a FileParameter object. However, its fields are initialized with null.

Changing the FileParameter class manually to IFormFile fixes these issues.
Is there some other way to make this work? Or is it unsupported currently?

RicoSuter:
Probably not supported correctly

SamuelMcAravey:
I can submit a pull request for this if that's ok. I've just run into the same issue here, and I've got the fix sitting in my fork.

RicoSuter:
Please create a pr and ill review it

bpieslak:
Good morning everyone.
I encountered this issue today and came across this outstanding PR.

I figure everyone is extremely busy, but wanted to check in and see if the PR that Samuel created might be reviewed or merged anytime soon.

Thanks!
-Brian

KoalaBear84:
We maybe are having the same issue. Any idea when this will be merged and released? Thanks in advance.

MarcinGadomski94:
Any updates on reviewing that PR? Met with the same problem.

ForgeColeColon:
As a temporary fix until this is merged, i modified the Controller.liquid class to replace FileParameter with IFormFile and works like a charm.  File is transferred fine.  Here is my change to line 113:
[...]

PR #2266, opened 2019-06-25
Users/samuelmcaravey/iformfile fix

SamuelMcAravey:
Updates for using IFormFile in AspNetCore controller generation.

Related: #2075

### DIFF
--- a/src/NSwag.CodeGeneration.CSharp.Tests/ParameterTests.cs
+++ b/src/NSwag.CodeGeneration.CSharp.Tests/ParameterTests.cs
@@ -309,5 +309,43 @@ namespace NSwag.CodeGeneration.CSharp.Tests
             //// Assert
             Assert.Contains("System.Threading.Tasks.Task Foo(Microsoft.AspNetCore.Http.IFormFile foo)", code);
         }
+
+        [Theory]
+        [InlineData(Models.CSharpControllerStyle.Abstract)]
+        [InlineData(Models.CSharpControllerStyle.Partial)]
+        public void When_multiple_file_parameter_with_AspNetCore_controller_use_IFormFile(Models.CSharpControllerStyle controllerStyle)
+        {
+            //// Arrange
+            var document = new OpenApiDocument();
+            document.Paths["foo"] = new OpenApiPathItem
+            {
+                {
+                    OpenApiOperationMethod.Get, new OpenApiOperation
+                    {
+                        Parameters =
+                        {
+                            new OpenApiParameter
+                            {
+                                Kind = OpenApiParameterKind.FormData,
+                                Name = "foo",
+                                Type = JsonObjectType.File,
+                                CollectionFormat = OpenApiParameterCollectionFormat.Multi
+                            }
+                        }
+                    }
+                }
+            };
+
+            //// Act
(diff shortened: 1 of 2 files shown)


src/NSwag.CodeGeneration.CSharp.Tests/ParameterTests.cs
```diff
@@ -272,5 +272,80 @@ public void Deep_object_properties_are_correctly_named()
...
+                    OpenApiOperationMethod.Get, new OpenApiOperation
+                    {
+                        Parameters =
+                        {
+                            new OpenApiParameter
+                            {
+                                Kind = OpenApiParameterKind.FormData,
+                                Name = "foo",
+                                Type = JsonObjectType.File,
+                                CollectionFormat = OpenApiParameterCollectionFormat.Multi
```
SamuelMcAravey: Not 100% sure if this is correct, but setting this was the only way I could find to trigger the branch of code that generated the collection of files.

src/NSwag.CodeGeneration.CSharp/Models/CSharpOperationModel.cs
```diff
@@ -248,11 +248,30 @@ protected override string ResolveParameterType(OpenApiParameter parameter)
...
+                }
+
                 if (parameter.CollectionFormat == OpenApiParameterCollectionFormat.Multi && !schema.Type.HasFlag(JsonObjectType.Array))
                 {
+                    if (useFormFile)
+                    {
+                        return "System.Collections.Generic.IEnumerable<Microsoft.AspNetCore.Http.IFormFile>";
+                    }
+
                     return "System.Collections.Generic.IEnumerable<FileParameter>";
```
RicoSuter: Sorry for my late reply. 

I think one of the reasons to use FileParameter (also for controllers) is that the generated service class is independent of the technology (ASP.NET Core) and thus should use FileParameter but then this has to be transformed to a IFormFile when calling the generated methods from the generated controller. Not sure whether it's worth the effort though (and I'm personally not using controller generation). 

What do you think?

src/NSwag.CodeGeneration.CSharp.Tests/ParameterTests.cs
RicoSuter: Maybe this is something which is not yet supported with NSwag and OAS3.0...

src/NSwag.CodeGeneration.CSharp/Models/CSharpOperationModel.cs
thantzinmyint: I don't think that it's worth the effort to map FileParameter to IFormFile in this case, given that we're able to confirm that we're targeting ASP.Net core and it is the accepted way to handle multi-part form submissions with attached files. Is this something that we could keep moving forward?

Evangelinexx:
@RicoSuter Are we able to get this PR merged?

MarcinGadomski94 (approved):

ForgeColeColon:
Are there still plans to merge?  I am facing the same problem, and getting IFormFile will fix it.

thantzinmyint (approved):

kvde:
@RicoSuter 
Is there an update for this issue?

PR #2267, opened 2019-06-26
added security schema info into the operation generation model

### LINKED ISSUE #2055: SwaggerToCSharpControllerGenerator does not support security definitions
When I generate my controllers there are no "Authorize" attributes being added.

I'm using a swagger 2.0 spec with a security definition that looks like this (globally):

```yaml
securityDefinitions:
  bearerAuth:
    type: apiKey
    in: header
    name: Authorization
    description: my description
security:
  - bearerAuth: []
```

I've had to use my own liquid template to fix the routeprefix issue (#1386), and I was hoping I could solve this security definitions issue myself as well but unfortunately the [CSharpOperationModel](https://github.com/RicoSuter/NSwag/blob/master/src/NSwag.CodeGeneration.CSharp/Models/CSharpOperationModel.cs#L19) doesn't provide any of the required data to the template.

Perhaps the [CSharpOperationModel](https://github.com/RicoSuter/NSwag/blob/master/src/NSwag.CodeGeneration.CSharp/Models/CSharpOperationModel.cs#L19) should provide a `public bool RequiresAuthentication { get; set; }` property that can be used in-order to add the `[Authorize]` attribute.

Place1:
this should fix #2055

i've only added the data to the code generation model to avoid a breaking change to the default template.

### DIFF
--- a/src/NSwag.CodeGeneration.CSharp/Models/CSharpOperationModel.cs
+++ b/src/NSwag.CodeGeneration.CSharp/Models/CSharpOperationModel.cs
@@ -207,10 +207,10 @@ namespace NSwag.CodeGeneration.CSharp.Models
         }
 
         /// <summary>True if the operation has any security schemes</summary>
-        public bool RequiresAuthentication => _operation.ActualSecurity.Count() != 0;
+        public bool RequiresAuthentication => (_operation.ActualSecurity?.Count() ?? 0) != 0;
 
         /// <summary>Gets the security schemas that apply to this operation</summary>
-        public IEnumerable<OpenApiSecurityRequirement> ActualSecurity => _operation.ActualSecurity;
+        public IEnumerable<OpenApiSecurityRequirement> Security => _operation.ActualSecurity;
 
         /// <summary>Gets the name of the parameter variable.</summary>
         /// <param name="parameter">The parameter.</param>
--- a/src/NSwag.CodeGeneration.CSharp/Models/CSharpOperationModel.cs
+++ b/src/NSwag.CodeGeneration.CSharp/Models/CSharpOperationModel.cs
@@ -206,6 +206,12 @@ namespace NSwag.CodeGeneration.CSharp.Models
             }
         }
 
+        /// <summary>True if the operation has any security schemes</summary>
+        public bool RequiresAuthentication => _operation.ActualSecurity.Count() != 0;
+
+        /// <summary>Gets the security schemas that apply to this operation</summary>
+        public IEnumerable<OpenApiSecurityRequirement> ActualSecurity => _operation.ActualSecurity;
+
         /// <summary>Gets the name of the parameter variable.</summary>
         /// <param name="parameter">The parameter.</param>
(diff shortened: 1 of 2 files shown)


src/NSwag.CodeGeneration.CSharp/Models/CSharpOperationModel.cs
```diff
@@ -206,6 +206,12 @@ public string RouteName
             }
         }
 
+        /// <summary>True if the operation has any security schemes</summary>
+        public bool RequiresAuthentication => _operation.ActualSecurity.Count() != 0;
+
+        /// <summary>Gets the security schemas that apply to this operation</summary>
+        public IEnumerable<OpenApiSecurityRequirement> ActualSecurity => _operation.ActualSecurity;
```
RicoSuter: Let's name this property just Security

src/NSwag.CodeGeneration.CSharp/Models/CSharpOperationModel.cs
```diff
@@ -206,6 +206,12 @@ public string RouteName
             }
         }
 
+        /// <summary>True if the operation has any security schemes</summary>
+        public bool RequiresAuthentication => _operation.ActualSecurity.Count() != 0;
```
RicoSuter: Is ActualSecurity never null?

Place1: nice catch!

src/NSwag.CodeGeneration.CSharp/Models/CSharpOperationModel.cs
Place1: sure thing

uniquelau:
@Place1 could you share an example of the liquid template you are using?

I've had a little play around, and I can see, the `Operation.ActualSecurity` is being set, but I can't figure out how to loop over the values.

RicoSuter:
The property name in the template is "Security".

uniquelau:
@RicoSuter thanks - would I be right in thinking that I need to extend the class in some way to be able to get access to the required data? I've had a little play inside `Controller.Method.Annotations.liquid`.

```
{% if operation.RequiresAuthentication -%}
    {% for o in operation.Security -%}
        [Authorize({{ o }})]
    {% endfor -%}
{% endif %}
```

```
Results in:

[Authorize([bearerAuth, System.Collections.Generic.List`1[System.Object]])]
```

RicoSuter:
o is also a complex object

RicoSuter:
Not sure whether its a good idea to directly expose the openapi classes - maybe we should map them to template models.

RicoSuter:
Sorry o is a dictionary:

https://apiverse.io/organizations/nuget-org/nuget-feeds/public/packages/NSwag.Core/versions/latest?type=NSwag.OpenApiSecurityRequirement&namespace=All&assembly=netstandard-2.0

Its hard to work with dicts in the tpls

PR #2282, opened 2019-07-01
Add repository metadata to NuGet packages

pascalberger:
Add repository metadata to NuGet packages


RicoSuter:
The correct URL is actually https://github.com/RicoSuter/NSwag.git

What benefits do we have with this additional information?

pascalberger:
> The correct URL is actually https://github.com/RicoSuter/NSwag.git

I'll update this

> What benefits do we have with this additional information?

This will show a link to the repository on nuget.org. See [here](https://devblogs.microsoft.com/nuget/introducing-source-code-link-for-nuget-packages/) for details.
This is the same as the project link as long as nswag.org redirects to GitHub, but makes it more explicit.

pascalberger:
> The correct URL is actually https://github.com/RicoSuter/NSwag.git

I've updated the URLs

RicoSuter:
Thanks for the PR!

PR #2283, opened 2019-07-01
Use embedded license for NuGet packages

pascalberger:
Use license expression for NuGet packages instead of deprecated license URL


RicoSuter:
Thanks for the PR!

PR #2284, opened 2019-07-02
Ensure axios returns request headers, use responseType: blob for files

mirata:
Axios request headers have been fixed. Currently, an empty request header object is returned because the object type is different to fetch.

Also, file requests have been fixed. The 'blob' requestType is used instead of 'arraybuffer' to ensure the content-type is preserved (previously a blob was constructed from an arraybuffer which can lose the content-type and display a file as garbage text)

These modifications should make the the client responses more consistent with the fetch template

### DIFF
--- a/src/NSwag.CodeGeneration.TypeScript/Templates/AxiosClient.liquid
+++ b/src/NSwag.CodeGeneration.TypeScript/Templates/AxiosClient.liquid
@@ -44,7 +44,7 @@
             data: content_,
 {%     endif -%}
 {%     if operation.IsFile -%}
-            responseType: "arraybuffer",
+            responseType: "blob",
 {%     endif -%}
             method: "{{ operation.HttpMethodUpper | upcase }}",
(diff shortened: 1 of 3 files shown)


thijscrombeen:
@mirata The reason I used ArrayBuffer is because of comments in the following https://github.com/axios/axios/issues/1392 where it was observed that blob doesn't work very well with large files. Did you have any issues with it?

mirata:
@thijscrombeen Can't say I had that issue. The trouble is that in the original code, the blob is constructed without a content type. So to be safe, I can rip out the content type from the header - that way I can still construct the blob the same way you had.

thijscrombeen:
@RicoSuter As for the code, it looks fine so in my opinion it's OK to merge.
@mirata Fair enough, if any issues arrive we can handle them then

RicoSuter:
Let's merge and see. I personally don't use axios so I need your feedback on this.

Issue #2294, opened 2019-07-04
TransformToExternalPath returning "//swagger/v1/swagger.json" when X-Forward-Prefix equals "/"


Choc13:
Hi, 

We are trying to upgrade to NSwag 13.0.3 and OpenApi3, however we run into issues with the swagger UI not being able to find the default swagger document when running behind a reverse proxy on Kube. 

In particular we see the following error in the UI:

![image](https://user-images.githubusercontent.com/2675431/60669458-413ed380-9e66-11e9-8933-71e194cefc3c.png)

We have the default setup in our `Startup` class, such as:

```
public void Configure(IApplicationBuilder app)
{
    app
        .UseForwardedHeaders()
        .UsePathBase("/") // This value comes from config in reality
        .UseOpenApi()
        .UserSwaggerUi3();
}

public void ConfigureServices(IServiceCollection services)
{
    services.AddOpenApiDocument();
}
```

Having traced this through your source, it appears that this is happening because the default implementation of `TransformToExternalPath` is appending `internalUiRoute` to `request.GetPathBase` and `GetPathBase` returns `"/"` because `X-Forwarded-Prefix` has a value of `"/"`, which results in the external path for the document being set to `"//swagger/v1/swagger.json"`. 

Essentially the trimming of `"/"`'s is inconsistent in [`HttpRequestExtension.GetPathBase`](https://github.com/RicoSuter/NSwag/blob/NSwag-Build-1067/src/NSwag.AspNetCore/HttpRequestExtension.cs#L58). Line 58 will return `"/"` as a minimum, whereas line 71 will return empty string if there is no path. The fix here would seem to be to make the behaviour at least consistent in both degenerate cases and then add or remove the `"/"` as necessary in `TransformToExternalPath`.

I would be happy to submit a PR to fix this, but I'm not entirely sure what the default behaviour should be here. On the one hand, it seems like the minimal base path should be `"/"` as it should be an absolute path, but on the other hand the default implementation of `TransformToExternalPath` seems to be expecting there to be no trailing slash on the result of `GetPathBase`. 

On a related note, why does `GetPathBase` try and look for a path section of the `X-Forwarded-Host` header? According to the Mozilla documentation about this header it seems like it should only ever contain a host section of a URI. Therefore it seems like the implementation of `GetPathBase` could be simplified somewhat to:

```
public static string GetBasePath(this HttpRequest request)
{
[...]

PR #2296, opened 2019-07-05
Fix default implementation of TransformToExternalPath

Choc13:
Addresses #2294 
* Ensures only a single separator is added between the base path and the internal ui path.

### DIFF
--- a/src/NSwag.AspNetCore/SwaggerUiSettingsBase.cs
+++ b/src/NSwag.AspNetCore/SwaggerUiSettingsBase.cs
@@ -35,7 +35,7 @@ namespace NSwag.AspNetCore
         public SwaggerUiSettingsBase()
         {
             TransformToExternalPath = (internalUiRoute, request) =>
-                request.GetBasePath().TrimEnd('/') + '/' + internalUiRoute.TrimStart('/');
+                request.GetBasePath()?.TrimEnd('/') + '/' + internalUiRoute?.TrimStart('/');
         }
 
         /// <summary>Gets or sets the internal swagger UI route (must start with '/').</summary>
--- a/src/NSwag.AspNetCore/SwaggerUiSettingsBase.cs
+++ b/src/NSwag.AspNetCore/SwaggerUiSettingsBase.cs
@@ -35,7 +35,7 @@ namespace NSwag.AspNetCore
         public SwaggerUiSettingsBase()
         {
             TransformToExternalPath = (internalUiRoute, request) =>
-                request.GetBasePath()?.TrimEnd('/') + '/' + internalUiRoute?.TrimStart('/');
+                '/' + (request.GetBasePath()?.TrimEnd('/') + '/' + internalUiRoute?.TrimStart('/')).Trim('/');
         }
 
         /// <summary>Gets or sets the internal swagger UI route (must start with '/').</summary>
--- a/src/NSwag.AspNetCore/SwaggerUiSettingsBase.cs
+++ b/src/NSwag.AspNetCore/SwaggerUiSettingsBase.cs
@@ -35,9 +35,7 @@ namespace NSwag.AspNetCore
         public SwaggerUiSettingsBase()
         {
             TransformToExternalPath = (internalUiRoute, request) =>
-            {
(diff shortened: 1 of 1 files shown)


src/NSwag.AspNetCore/SwaggerUiSettingsBase.cs
```diff
@@ -35,9 +35,7 @@ public abstract class SwaggerUiSettingsBase : SwaggerSettings
         public SwaggerUiSettingsBase()
         {
             TransformToExternalPath = (internalUiRoute, request) =>
-            {
-                return request.GetBasePath() + internalUiRoute;
-            };
+                request.GetBasePath().TrimEnd('/') + '/' + internalUiRoute.TrimStart('/');
```
RicoSuter: Do we need null checks?

Choc13: Good idea. I'll add that.

RicoSuter: I'm usually defensive here... checking the docs/source code might be another option :-)

Choc13: Yeah, agree with the defensive approach. I've added the null handling, so if both sides are null then this will output `"/"`. I don't see a situation where the internal route would be `null` and I'm pretty sure `GetBasePath` at least returns empty string or "/" anyway depending on which case it falls into.

src/NSwag.AspNetCore/SwaggerUiSettingsBase.cs
```diff
@@ -35,9 +35,7 @@ public abstract class SwaggerUiSettingsBase : SwaggerSettings
         public SwaggerUiSettingsBase()
         {
             TransformToExternalPath = (internalUiRoute, request) =>
-            {
-                return request.GetBasePath() + internalUiRoute;
-            };
+                request.GetBasePath()?.TrimEnd('/') + '/' + internalUiRoute?.TrimStart('/');
```
RicoSuter: Shouldnt the produced path start with a /?

Choc13: This will providing `GetBasePath` returns a path that starts with `/` or returns `null` or `string.Empty`. I guess another question would be should `GetBasePath` always start with `/`?

Choc13: If you want to be really safe, I can ensure there is always a `/` added to `GetBasePath` here too. I think what would be good would be some kind of function internal to this library for building up URLs. Similar to how Flurl does it for example.

RicoSuter: Ok, looks good to me, assuming that GetBasePath() will return either /, null or ""..

src/NSwag.AspNetCore/SwaggerUiSettingsBase.cs
```diff
@@ -35,9 +35,7 @@ public abstract class SwaggerUiSettingsBase : SwaggerSettings
         public SwaggerUiSettingsBase()
         {
             TransformToExternalPath = (internalUiRoute, request) =>
-            {
-                return request.GetBasePath() + internalUiRoute;
-            };
+                '/' + (request.GetBasePath()?.TrimEnd('/') + '/' + internalUiRoute?.TrimStart('/')).Trim('/');
```
RicoSuter: I think the returned route must start with /, no?

`request.GetBasePath()?.TrimEnd('/') + '/' + internalUiRoute?.TrimStart('/')` should be enough

RicoSuter:
LGTM now... merge?

> I think what would be good would be some kind of function internal to this library for building up URLs. Similar to how Flurl does it for example.

Would make sense if we are doing more of this stuff in the future... but to me it's fine for now.

Choc13:
Yep go for it if you're happy. What's your release schedule usually like? Will this go out as a patch?

RicoSuter:
Yes, i can release a patch

PR #2310, opened 2019-07-16
Fixed processing of Summary and Description fields of OpenApiOperation

korneyev:
Currently content of a DescriptionAttribute on a controller method (operation) is being parsed into a OpenAPI summary field instead of a description field (see this issue: https://github.com/RicoSuter/NSwag/issues/1930).

Unfortunately, this behavior can't be changed by simply parsing DescriptionAttribute into description field since it breaks backward compatibility. Also we need some source for summary field - and it looks like there is no appropriate attribute in System.ComponentModel namespace.

Proposed solution is to extend OpenApiOperationAttribute class with Summary and Description properties and use them in OperationSummaryAndDescriptionProcessor. 

If controller method has OpenApiOperationAttribute, then we'll get summary and description from it. If there is no such attribute, then usual logic will be applied - get summary from DescriptionAttribute (or xml docs summary) and get description from xml docs remarks.

### DIFF
--- a/src/NSwag.Annotations/OpenApiOperationAttribute.cs
+++ b/src/NSwag.Annotations/OpenApiOperationAttribute.cs
@@ -38,6 +38,12 @@ namespace NSwag.Annotations
             Summary = summary;
             Description = description;
         }
+
+        /// <summary>Gets or sets the operation summary.</summary>
+        public string Summary { get; private set; }
+
+        /// <summary>Gets or sets the operation description.</summary>
+        public string Description { get; private set; }
     }
 
     /// <summary>Specifies the operation id.</summary>
@@ -54,11 +60,5 @@ namespace NSwag.Annotations
 
         /// <summary>Gets or sets the operation ID.</summary>
         public string OperationId { get; private set; }
-
-        /// <summary>Gets or sets the operation summary.</summary>
-        public string Summary { get; protected set; }
-
-        /// <summary>Gets or sets the operation description.</summary>
-        public string Description { get; protected set; }
     }
 }
--- a/src/NSwag.Annotations/OpenApiOperationAttribute.cs
+++ b/src/NSwag.Annotations/OpenApiOperationAttribute.cs
@@ -10,14 +10,33 @@ using System;
 
 namespace NSwag.Annotations
 {
-    /// <summary>Specifies the operation id.</summary>
+    /// <summary>Specifies the operation id, summary and description</summary>
     [AttributeUsage(AttributeTargets.Method)]
     public class OpenApiOperationAttribute : SwaggerOperationAttribute
     {
-        /// <summary>Initializes a new instance of the <see cref="SwaggerOperationAttribute"/> class.</summary>
+        /// <summary>Initializes a new instance of the <see cref="OpenApiOperationAttribute"/> class.</summary>
         /// <param name="operationId">The operation ID.</param>
         public OpenApiOperationAttribute(string operationId) : base(operationId)
         {
+        }
+
+        /// <summary>Initializes a new instance of the <see cref="OpenApiOperationAttribute"/> class.</summary>
+        /// <param name="summary">The operation summary.</param>
+        /// /// <param name="description">The operation description.</param>
+        public OpenApiOperationAttribute(string summary, string description) : base(null)
+        {
+            Summary = summary;
+            Description = description;
+        }
+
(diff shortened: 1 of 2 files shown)


src/NSwag.Generation/Processors/OperationSummaryAndDescriptionProcessor.cs
```diff
@@ -22,29 +23,56 @@ public class OperationSummaryAndDescriptionProcessor : IOperationProcessor
...
+        private void ProcessDescription(OperationProcessorContext context, List<Attribute> attributes)
+        {
+            dynamic openApiOperationAttribute = attributes
+                .SingleOrDefault(a => a.GetType().Name == "OpenApiOperationAttribute");
+
+            string description = openApiOperationAttribute?.Description;
+
+            if (string.IsNullOrEmpty(description))
+            {
+                description = context.MethodInfo.GetXmlDocsRemarks();
```
RicoSuter: Isnt this already done somewhere else?

korneyev: Hmmm.... Not quite sure if I understood your question correctly. Can you "expand" you idea - what is possibly wrong in the code snippet above?

src/NSwag.Annotations/OpenApiOperationAttribute.cs
```diff
@@ -35,5 +54,11 @@ public SwaggerOperationAttribute(string operationId)
 
         /// <summary>Gets or sets the operation ID.</summary>
         public string OperationId { get; private set; }
+
+        /// <summary>Gets or sets the operation summary.</summary>
+        public string Summary { get; protected set; }
```
RicoSuter: Why is this protected?

src/NSwag.Generation/Processors/OperationSummaryAndDescriptionProcessor.cs
RicoSuter: I think there is already a place where GetXmlDocsRemarks is used and set, or is this wrong? Isnt this somehow duplicated code?

korneyev: As far as I can see - OperationSummaryAndDescriptionProcessor class is the only place where GetXmlDocsRemarks is used. Source code before my suggested change has this call at line 41 in Process method, but i've split it to separate ProcessSummary and ProcessDescription methods.

src/NSwag.Annotations/OpenApiOperationAttribute.cs
korneyev: SwaggerOperationAttribute is marked as Obsolete and descendant OpenApiOperationAttribute is recommended to use. However, (I suppose this is to preserve backward compatibility), SwaggerOperationAttribute still used to obtain OperationID property, and OpenApiOperationAttribute just "wraps" it.

In order to keep all properties "together", I've placed Summary and Description in SwaggerOperationAttribute, not in OpenApiOperationAttribute - but in order to force usage of last one, I've created constructors with Summary and Description in OpenApiOperationAttribute  and not in SwaggerOperationAttribute. This was the reason of having protected setter for those properties.

Probably it is a kind of "over-engeneering", and thosу properties should just be in OpenApiOperationAttribute.

src/NSwag.Generation/Processors/OperationSummaryAndDescriptionProcessor.cs
RicoSuter: Ok then it's fine I think.

src/NSwag.Annotations/OpenApiOperationAttribute.cs
RicoSuter: > I suppose this is to preserve backward compatibility

Exactly 

> Probably it is a kind of "over-engeneering", and thosу properties should just be in OpenApiOperationAttribute.

New functionality should be added only in OpenApiOperationAttribute i think, so fine for me. Can we use private instead of protected then?

korneyev: Yes, I think those properties should be in OpenApiOperationAttribute.
I moved them and made private setters - see new commit.

RicoSuter:
Is the PR ready?

korneyev:
> 
> 
> Is the PR ready?

Yes, it is ready.

PR #2315, opened 2019-07-19
Fix throwException formatting in generated Typescript clients

ljvanschie:
Fixes the formatting of generated Typescript clients (adds a space in the if-statement).

```ts
function throwException(message: string, status: number, response: string, headers: { [key: string]: any; }, result?: any): any {
    if(result !== null && result !== undefined)
        throw result;
    else
        throw new SwaggerException(message, status, response, headers, null);
}
```

This makes formatting consistent with other if-statements within the generated clients. It's something that autoformatters will often change otherwise.


RicoSuter:
Thanks for the pr

ljvanschie:
Thanks for merging, that was fast! 😄

PR #2321, opened 2019-07-22
Use ExceptionDescription in TypeScript client

altso:
Use response description as a custom message in error class (`exceptionClass`).

### DIFF
--- a/src/NSwag.CodeGeneration.TypeScript/Templates/Client.ProcessResponse.HandleStatusCode.liquid
+++ b/src/NSwag.CodeGeneration.TypeScript/Templates/Client.ProcessResponse.HandleStatusCode.liquid
@@ -65,7 +65,7 @@ return result{{ response.StatusCode }};
 {%                 endif -%}
 {%             endif -%}
 {%         else -%}
-return throwException({% if Framework.IsAngularJS %}this.q, {% endif %}"A server error occurred.", status, _responseText, _headers, result{{ response.StatusCode }});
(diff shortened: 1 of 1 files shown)


src/NSwag.CodeGeneration.TypeScript/Templates/Client.ProcessResponse.HandleStatusCode.liquid
```diff
@@ -90,7 +90,7 @@ return{% if operation.HasResultType %} null{% endif %};
 {%             endif -%}
 {%         endif -%}
 {%     else -%}
-return throwException({% if Framework.IsAngularJS %}this.q, {% endif %}"A server error occurred.", status, _responseText, _headers);
+return throwException({% if Framework.IsAngularJS %}this.q, {% endif %}"{{ response.ExceptionDescription }}", status, _responseText, _headers);
```
RicoSuter: Is `ExceptionDescription` an existing property on the repsonse model?

RicoSuter: Maybe we should fall back to the current message if the description is null or empty?

altso: Apologies, I should have put more info in the summary. `ExceptionDescription` is an existing property and never returns `null` or empty string.

https://github.com/RicoSuter/NSwag/blob/85ae862fd6d68173a201a79e0ad06e0be2ec5de1/src/NSwag.CodeGeneration/Models/ResponseModelBase.cs#L94-L96

altso:
Thank you!

Issue #2319, opened 2019-07-22
Parameter defined with reference missing in generated C# code


molnarm:
Example YAML:

```yaml
swagger: '2.0'
info:
  title: Reference parameters
  version: v0.0.1
paths:
  '/orders/{orderId}/order-items/{orderItemId}':
    parameters:
      - $ref: '#/parameters/OrderId'
      - $ref: '#/parameters/OrderItemId'
    delete:
      summary: Delete an order item
      description: >-
        This method allows to remove an order item from an order, by specifying
        their ids.
      responses:
        "204":
          description: No Content.
        default:
          description: Default response
          schema:
            $ref: '#/definitions/Error'
definitions:
  Error:
    type: object
    properties:
      message:
          type: string
parameters:
  OrderId:
    name: orderId
    in: path
    description: Identifier of the order.
    required: true
    type: string
    format: uuid
  OrderItemId:
    name: orderItemId
    in: path
    description: Identifier of the order item.
    required: true
    type: string
    format: uuid
```

Generated code (only the interface without comments for briefness, the issue is visible here too:

```csharp
public partial interface IClient
    {
        /// <summary>Delete an order item</summary>
        System.Threading.Tasks.Task OrdersOrderidOrderItemsOrderitemidDeleteAsync(System.Guid orderId);
    
        /// <summary>Delete an order item</summary>
        void OrdersOrderidOrderItemsOrderitemidDelete(System.Guid orderId);
    
        /// <summary>Delete an order item</summary>
        System.Threading.Tasks.Task OrdersOrderidOrderItemsOrderitemidDeleteAsync(System.Guid orderId, System.Threading.CancellationToken cancellationToken);
    
    }
```

As you can see, the DELETE operation has two parameters (orderId and orderItemId), but only the first one is present in the generated code. The swagger is valid according to the online editor.

molnarm:
I looked into this a bit, and I think the problem is [here](https://github.com/RicoSuter/NSwag/blob/85ae862fd6d68173a201a79e0ad06e0be2ec5de1/src/NSwag.Core/OpenApiOperation.cs#L117): when concatenating the parameters of the parent, the projection to ActualParameter is missing. As a result, all parameters from the parent have `null` as name, so only the first one is retained.

I would create a pull request, but I'm not that familiar yet with the internals of your library to be sure about this (and I couldn't find the test for referenced parameters either).

RicoSuter:
Can you create a PR so i can check the fix?

PR #2334, opened 2019-07-29
#2319 Parent parameter references should be resolved before consolidating duplicates in operation parameters

molnarm:


RicoSuter:
Looks good to me, thanks.

Issue #2337, opened 2019-07-31
Square brackets in the query parameter names are not escaped


mirroj:
I'm facing issue that the square brackets in the query parameter names are not escaped. We have API that looks like this:
`GET test?page[number]=1&page[size]=1`

In order to correctly call the API, the parameter names should be escaped like this:
`GET test?page%5Bnumber%5D=1&page%5Bsize%5D=1`

The following is extracted from section 3.2.2 of RFC 3986 :

> A host identified by an Internet Protocol literal address, version 6
> [RFC3513] or later, is distinguished by enclosing the IP literal
> within square brackets ("[" and "]"). This is the only place where
> square bracket characters are allowed in the URI syntax.

Following OpenAPI spec produces the output:
```
{
  "swagger" : "2.0",
  "info" : {
    "version" : "1.0.2",
    "title" : "Test API"
  },
  "host" : "localhost:8080",
  "basePath" : "/",
  "tags" : [ {
    "name" : "api"
  } ],
  "schemes" : [ "http" ],
  "paths" : {
     "/test" : {
      "get" : {
        "parameters" : [ {
          "name" : "page[number]",
          "in" : "query",
          "type" : "integer"
        }],
        "responses" : {
          "default" : {
            "description" : "successful operation"
          }
        }
      }
    }
  }
}

```

PR #2338, opened 2019-07-31
Added System.Uri.EscapeDataString for parameter names in query string

mirroj:
Fix for Issue #2337 
The following is extracted from section 3.2.2 of RFC 3986 :

> A host identified by an Internet Protocol literal address, version 6
> [RFC3513] or later, is distinguished by enclosing the IP literal
> within square brackets ("[" and "]"). This is the only place where
> square bracket characters are allowed in the URI syntax.

The square brackets that appear in query string should be escaped. For example:
`GET test?page[number]=1&page[size]=1`

Should be escaped to:
`GET test?page%5Bnumber%5D=1&page%5Bsize%5D=1`


RicoSuter:
Thanks for the PR!

PR #2356, opened 2019-08-16
Added global alias for System in C# generation

knstntn:
I run into namespace conflict if using NSwag code generation for C#

## Context
We have namespace `Microsoft.Omex.System` (unfortunate name, and we work on renaming this https://github.com/microsoft/Omex/issues/101). I made a new namespace `Microsoft.Omex.Metrics.ApiClient` and used this namespace for ApiClient generated using nswag

Result is shown on this picture
image

## Current Behavior
Compilation failed because `System` resolved to `Microsoft.Omex.System`

## Expected Behavior
Compilation is fine.

## Possible Solution
Add global alias `System = global::System` in generated file
image
or
prefix all usages of `System` with `global::` alias.

Given that first fix is way smaller, I decided to go for it. Let me know if second option is preferred. 

PS: I did not make a test - I failed to build code on my local machine, so was not even sure how to run tests.

### DIFF
--- a/src/NSwag.CodeGeneration.CSharp.Tests/FileTests.cs
+++ b/src/NSwag.CodeGeneration.CSharp.Tests/FileTests.cs
@@ -1,0 +1,39 @@
+﻿using System;
+using System.Net.Http;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Mvc;
+using NSwag.Generation.WebApi;
+using Xunit;
+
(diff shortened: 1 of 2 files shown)


RicoSuter:
> I failed to build code on my local machine, so was not even sure how to run tests.

Try with NSwag.Min.sln

But i think this should be fine.

knstntn:
> Try with NSwag.Min.sln


Thanks, it worked. Added simple test to check that corresponding content is there.

PR #2363, opened 2019-08-20
[WIP] Add OpenApiCallbackAttribute

piaste:
Opening a pull request for CI testing and feedback. Some XML comments still missing etc.

This attribute allows one to define OpenAPI callbacks associated a method. Very important for webhook APIs.

Since an attribute constructor isn't enough to fully define every property of an `OpenApiPathItem` object, I tried to strike a balance between flexibility and ergonomics in the constructor overloads:

 - Callback URL: required
 - Callback name: optional, defaults to "OperationID_Callback"
 - Operation: defaults to POST
 - MIME type: same logic as the default processor, if unspecified checks `.IsBinary` for octet-stream or json
 - Types: zero means empty request body, multiples uses `oneOf` to encode multiple options. You can also add multiple callbacks, of course.


RicoSuter:
Looks good so far, just ping me when the PR is ready for a review.

src/NSwag.Annotations/OpenApiCallbackAttribute.cs
```diff
@@ -0,0 +1,104 @@
...
+            this.Name = name;
+            this.Types = types;
+            this.CallbackUrl = callbackUrl;
+            this.Method = method;
+            this.MimeType = mimeType;
+        }
+        /// <summary>
+        /// The name of the callback
+        /// </summary>
+        public string Name { get; }
```
RicoSuter: I think in the other attributes we also have setters to change properties (not only in the ctor). Maybe we should do it the same way in this attribute...

RicoSuter:
Any update on this pr?

laerciolima:
which annotation should i use to callback?

RicoSuter:
?

laerciolima:
hello, I would like to know if this implementation is callback (https://swagger.io/docs/specification/callbacks/) and what annotation do I use in C#?

piaste:
Apologies for completely forgetting about this PR. I'll finish it this weekend.

piaste:
I've addressed some issues, including your comment, and I've added an extension point to customize expected responses to each callback.

I think it's ready but I have one issue with passing tests which may indicate a deeper problem.

When the attribute contains the `typeof<SomeComplexType>`, in the tests it generates an empty schema instead of building the object's schema.

E.g. given:

```csharp
public class SwaggerCallbackController : Controller
    {
        public class MyPayload
        {
            public int MyProperty { get; set; }
        }
        [OpenApiCallback("url4", "multiple", types: new[] { typeof(string), typeof(MyPayload) })]
        public void MyAction(string foo)
        // snip
    }
```

the generated schema for the callback to `url4` becomes:

```json
{"oneOf":[{"type":"string"},{}]}
```

The code that generates the schemas for the individual types in the attribute is the same I'm currently using in production:

```csharp
var payloadSchemas = types.Select(t =>
                    context.SchemaGenerator.GenerateWithReferenceAndNullability<JsonSchema>(
                        t.ToContextualType(),
                        context.SchemaResolver
                    )).ToArray();
```

I'm guessing the NJsonSchema generator has some problems 'seeing' the `MyPayload` class? Could it have something to do with the fact that I placed the operation processor in the `NSwag.Generation` library instead of `NSwag.Generator.AspNetCore` and/or `NSwag.Generation.WebApi`?

piaste:
Ah, I noticed that you asked for a ping. @RicoSuter

pekunicki:
Hi, are there any plans on merging this feature to the master?

aboyaniv:
will this go to master ? is not , how can i do a webhook documentation ?

RicoSuter:
I think this can be completely done with NSwag extension points and does not need to be part of NSwag itself (ie own pkg)

PR #2376, opened 2019-08-28
Return FileResult type for binary response if an AspNetCore controller is generated

Alamaster99:


Alamaster99:
Hi @RicoSuter , will this PR be merged ever? Thanks.

NorekZ:
Thanks @Alamaster99 for fix, but I need to get in into master. @RicoSuter can you merge it pls?

Thank you very much

RicoSuter:
Here is the problem: The method in the IController interface should return a FileResponse (as it does now). This ensures that the interface is not coupled to aspnet core and is independent of any fw (avoid leaky abstraction). But then this FileResponse should be transformed into a FileResult in the controller. This is quite some more work then just change the type. Do you agree that this would be better? 

However because i dont use controller generator i probably should just merge this..

NorekZ:
Thank you for your response @RicoSuter

I understand the problem but is there any other way to solve it without need to implement custom IActionResultExecutor<FileResponse>, that would populate HttpResponse out of FileResponse (like it works for FileResult)? Tried also adding FileResult as a base class of FileResponse in partial class, but with no success (FileResult's constructor would have to call FileResult's constructor with 2 parameters)

I can't think of any other way to transform FileResponse to FileResult...

I suggest you to just merge it as it is and possibly plan some major rework in the future.

Alamaster99:
Hi, thanks for the replies. I see merging this version as the simplest and least problematic solution. To if you do not plan to make any larger changes to implement this, I would be glad if you would merge this.

RicoSuter:
The transformation would also be generated. But for now lets just merge this

Alamaster99:
> The transformation would also be generated. But for now lets just merge this

Yes. Would you please merge it? Thanks.

FlorianArnould:
> Nuget package: NSwag.MSBuild 13.7.0

I get an error from this change
```c#
[Microsoft.AspNetCore.Mvc.HttpGet, Microsoft.AspNetCore.Mvc.Route("file/{name}")]
public abstract System.Threading.Tasks.Task<FileResult> DownloadFile(string name, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken));
```
It breaks the file as there is no import in it. We miss the full namespace for `FileResult`, should be `Microsoft.AspNetCore.Mvc.FileResult`.

Worked around using: `/AdditionalNamespaceUsages:Microsoft.AspNetCore.Mvc` but should be cleaner to fix the generator

RicoSuter:
We'd need to fix this and use a fully qualified type name (with namespace) - I wonder why this is not done in the first place? Maybe because it's different between aspnetcore and aspnet? Can you create a PR for that?

PR #2381, opened 2019-08-29
Use OpenApiTagAttribute of class to set tags for those methods where such attribute is not set explicitly

korneyev:
Currently  OpenApiTagAttribute (and OpenApiTagsAttribute as well) has targets of both AttributeTargets.Class and AttributeTargets.Method. 

But setting it as class attribute has no effect because OperationTagsProcessor uses only attributes from method and if no such attributes are present - processor uses controller class name.

Suggested change adds following OperationTagsProcessor logic: if no OpenApiTag atrtributes present at method, then try to get OpenApiTag (and OpenApiTags as well) from class, and if they are not present - then use controller class name as usual.

This can be helpful when controller contains a lot of methods and all of them should have the same tag(s).


RicoSuter:
Please also add a test

korneyev:
> 
> 
> Please also add a test

I've added OperationTagsProcessorTests. Since OperationTagsProcessor is in NSwag.Generation assembly I have to add new test project NSwag.Generation.Tests

korneyev:
> 
> 
> Please also add a test

@RicoSuter  ping ;)

PR #2457, opened 2019-10-10
Use 3.0 GA packages from Microsoft

dougbu:
- adjust opting into document generation slightly (might need a release note?)
- bump npm version in lock file


RicoSuter:
> might need a release note?

What would you write?

dougbu:
> What would you write?

A quick note that `$(OpenApiGenerateDocuments)` should now be used to (completely) _disable_ document generation e.g. for projects targeting .NET Framework and using older versions of ASP.NET Core (a much harder to detect case than using older TFMs) while `$(OpenApiGenerateDocumentsOnBuild)` should be used to opt into the feature on every build.

The change isn't breaking per-se but those who've opted into doc gen using `$(OpenApiGenerateDocuments)` may wonder why docs aren't being generated after upgrading w/o a note.

PR #2458, opened 2019-10-10
Added support for custom name generators

mirroj:
Calling the InitializeCustomTypes method to load custom type/property/enum name generators.


RicoSuter:
And this just works? There's a reason why it's not enabled yet :-)

BastiaanLouis:
> My question: What custom types can you use like this? How can you load external assemblies?

Wondering what is needed to finish this PR. @RicoSuter: What do you mean with this question? More specific, what do you mean with 'custom types'? It lookes like any custom type (with a default constructor) can be initialized like this.

RicoSuter:
I think we also need a config with custom assembly paths to load the specified types from etc

PR #2460, opened 2019-10-11
Use http method names to distinguish duplicate operations

mirroj:
When generating C# client from yml document for a typical REST api that contains paths:
```
GET person
POST person
DELETE person
```
the generated client contains methods 
```
Person1Async()
Person2Async()
Person3Async()
```
I think this is pretty common case and the http mehod name could be used to distinguish the duplicates

### DIFF
--- a/src/NSwag.Core.Tests/OperationIdTests.cs
+++ b/src/NSwag.Core.Tests/OperationIdTests.cs
@@ -28,5 +28,40 @@ namespace NSwag.Core.Tests
             //// Assert
             Assert.True(document.Operations.GroupBy(o => o.Operation.OperationId).All(g => g.Count() == 1));
         }
+
+        [Fact]
+        public void When_operation_ids_are_not_unique_then_method_names_are_used()
+        {
+            //// Arrange
+            var document = new OpenApiDocument();
+            document.Paths["path1"] = new OpenApiPathItem
+            {
+                {
+                    OpenApiOperationMethod.Get,
+                    new OpenApiOperation { OperationId = "foo" }
+                },
+                {
+                    OpenApiOperationMethod.Post,
+                    new OpenApiOperation { OperationId = "foo" }
+                }
+            };
+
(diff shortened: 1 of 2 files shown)


RicoSuter:
There are two points to note: 

- A document with missing operation IDs is invalid - so this operation id generator is in theory not even needed/"allowed"
- This is potentially a breaking change for some users (this is my biggest problem)

RicoSuter:
Maybe we should first exactly define how we generate operation IDs - for example NSwag generates operation IDs in ASP.NET Core as "controller_method" (e.g. "BusinessProcess_Search") so I'd prefer a similar form here.

Gakk:
Initially I agreed with @mirroj, also wanting my generated client method names to start with the HTTP method.

The response from @RicoSuter about operation ID lead me to read some more about OpenAPI, as I knew nothing about operation ID.

At my company we are using Swashbuckle to generate Swagger documentation from server, and then NSwag to read the documentation and generate client classes.

When generating a Swagger document with Swashbuckle it does not out of the box add operation IDs. An I can not see that this violates the standard either. The field for **operationId** is **not required** in the specifications: http://spec.openapis.org/oas/v3.0.2#fixed-fields-7

Nevertheless when having a document with well formed operation IDs, I manage without using HTTP method names when generating the client - as I already can have the HTTP method names in the operation ID.

But often I am not the owner of the Swagger document, and then I might have to generate client code without operation IDs, and then the issue gets important again.

Maybe add method names when there are no operation IDs present?

RicoSuter:
Cant you just use a path based operation name generator setting?

mirroj:
@RicoSuter you mean the settings `"operationGenerationMode": "SingleClientFromPathSegments"`, right. I haven't know it exists when I created this issue. It did the job for our purposes. We have a long path, so the generated method name is longer but it does have the HTTP method appended. It's definitely better than `Method1`, `Method2`

PR #2462, opened 2019-10-12
Fix #2453

### LINKED ISSUE #2453: aspnetcore2swagger cli command throws FileNotFoundException when using relative assembly path
I am trying to create a swagger file for my api using the `NSwag.ConsoleCore` package a .NET Core CLI tool (13.1.2). Sadly I can't get it to work.
```
dotnet nswag aspnetcore2openapi /assembly:./bin/Debug/NSwagTest.dll /output:swagger.json
```
produces the error
```
System.IO.FileNotFoundException: Could not load file or assembly 'C:\Users\Adrian\source\repos\NSwagTest\NSwagTest\bin\Debug\bin\Debug\NSwagTest.dll
```
Note the double `bin\Debug\bin\Debug` inside the path. On the other hand
```
dotnet nswag aspnetcore2swagger /assembly:./bin/Debug/NSwagTest.dll /output:swagger.json
```
produces a similiar error
```
System.IO.FileNotFoundException: Could not load file or assembly 'C:\Users\Adrian\source\repos\NSwagTest\NSwagTest\NSwagTest.dll'.
```
But no more `bin\Debug`. I feel like there is a bug building the absolute path here. Using the full absolute path for `/assembly` works, but IMO isn't really feasible in an CI environment, where you you only can work within your own project file structure.

## Steps to reproduce
1. Create new ASP.NET Core Web API project
1. Add `<DotNetCliToolReference Include="NSwag.ConsoleCore" Version="13.1.2" />` to `.csproj` file
1. Run `dotnet nswag aspnetcore2openapi` command with `/assembly` option from within the project folder

kuper-adrian:
#2453

Tried to fix the problem myself. Looks like the problem originated from `AspNetCoreToOpenApiCommand` where different base directories were used to load the assemblies (during web host creation) and generate the documents.

My own basic tests were successful, but since I'm far from fully understanding the code base, I'm not sure about the implications of my changes. Couldn't find any unit tests for `AspNetCoreToOpenApiCommand` or related classes either. So any feedback would be nice.

### DIFF
--- a/src/NSwag.Commands/Commands/Generation/AspNetCore/AspNetCoreToOpenApiCommand.cs
+++ b/src/NSwag.Commands/Commands/Generation/AspNetCore/AspNetCoreToOpenApiCommand.cs
@@ -306,10 +306,9 @@ namespace NSwag.Commands.Generation.AspNetCore
 
         protected override async Task<string> RunIsolatedAsync(AssemblyLoader.AssemblyLoader assemblyLoader)
         {
-            var currentWorkingDirectory = ChangeWorkingDirectoryAndSetAspNetCoreEnvironment();
             using (var webHost = await CreateWebHostAsync(assemblyLoader))
             {
-                var document = await GenerateDocumentAsync(assemblyLoader, webHost.TryGetPropertyValue<IServiceProvider>("Services"), currentWorkingDirectory);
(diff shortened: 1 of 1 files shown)


RicoSuter:
The problem is that this code is there for a reason and just removing it probably introduces regressions...

RicoSuter:
Probably it's better to fix your problem somehow in ChangeWorkingDirectoryAndSetAspNetCoreEnvironment so that it applies for every caller

PR #2472, opened 2019-10-18
Add an extension method for the OperationId and httpPaths path templates in the WebApiOpenApiDocumentGeneratorSettings class

mccj:
Add an extension method for the OperationId and httpPaths path templates in the WebApiOpenApiDocumentGeneratorSettings class


RicoSuter:
Can't you just implement an operation processor which updates the operation id based on the context (instead of adding more extension points)?

Issue #2375, opened 2019-08-28
Support for Authorization code flow with PKCE ?


svenerp:
I am trying to get authorization code flow with PKCE to work. 
When clicking authorize I get the following error: 
[invalid_request]: code challenge required. 

Am I doing something wrong, or is the flow not supported (yet)?
Client Credential flow and implicit flow both work fine.

jamesthurley:
It looks like this is now supported in Swagger UI as of 8 October:

https://github.com/swagger-api/swagger-ui/pull/5361#issuecomment-539259962

espenrl:
I wanted to do a pull request on this. But I'm unsure about upgrading the bundled swagger-ui files. I've done an analysis of what needs to be done.

1. Upgrade bundled swagger-ui (Swagger UI 3 only) 
2. Add configuration property `UsePkceWithAuthorizationCodeGrant`
3. Template the property into index.html

#### 2

https://github.com/RicoSuter/NSwag/blob/1a6c959bdb41b41f2f534d3e61a311f1cf0de6f7/src/NSwag.AspNetCore/OAuth2ClientSettings.cs#L18

``` CSharp
/// <summary>Proof Key for Code Exchange. Only applies to `accessCode` flow. Supported in SwaggerUI 3.</summary>
public bool UsePkceWithAuthorizationCodeGrant { get; set; }
```

#### 3

https://github.com/RicoSuter/NSwag/blob/1a6c959bdb41b41f2f534d3e61a311f1cf0de6f7/src/NSwag.AspNetCore/SwaggerUi3/index.html#L87

``` CSharp
usePkceWithAuthorizationCodeGrant: {UsePkceWithAuthorizationCodeGrant}
```

NOTE: `SwaggerUi3Settings.TransformHtml` code will do the actual job of inserting the property into `index.html`.

RicoSuter:
Can you create a PR? I’ll update swagger ui 3

espenrl:
Sure, I have actually figured out how to update the swagger-ui library. I will do some testing and make a PR.

colinheppener:
Good day, 
First of all I'd like to thank you all for working on such a great project.
I was wondering if there is any idea when this might be included in a future release?
I was hoping to use it for a project I'm working on right now and this feature was merged after the most recent NuGet package release. Thanks!

RicoSuter:
I’ll publish a new version next week

colinheppener:
That would be perfect. Thanks!

[...]

PR #2480, opened 2019-10-22
NSwag.AspNetCore support PKCE

espenrl:
Tested and verified against IdentityServer.

swagger-ui updated from `swagger-ui-dist-3.24.0.tgz`


espenrl:
#2375

PR #2484, opened 2019-10-23
Add Default Parameters For C# Controllers

basvandrongelen:
The default parameters should be consistently generated regardless what options are chosen.

This fix adds the default params for the abstract controller and in case the wrapper option is chosen.


src/NSwag.CodeGeneration.CSharp/Templates/Controller.liquid
```diff
@@ -62,7 +62,7 @@ public partial class {{ Class }}Controller : {% if HasBaseClass %}{{ BaseClass }
...
     {
 {%         if operation.WrapResponse -%}
-        var result = await _implementation.{{ operation.ActualOperationName }}Async({% for parameter in operation.Parameters %}{{ parameter.VariableName }}{% if parameter.IsLast == false or UseCancellationToken %}, {% endif %}{% endfor %}{% if UseCancellationToken %}cancellationToken{% endif %}).ConfigureAwait(false);
+        var result = await _implementation.{{ operation.ActualOperationName }}Async({% for parameter in operation.Parameters %}{{ parameter.VariableName }}{% if parameter.HasDefault %} ?? {{parameter.Default}}{% endif %}{% if parameter.IsLast == false or UseCancellationToken %}, {% endif %}{% endfor %}{% if UseCancellationToken %}cancellationToken{% endif %}).ConfigureAwait(false);
```
RicoSuter: Please dont use “new” c# features as some users might not be able to use them - use `... ? ... :` here. Do you agree?

basvandrongelen: Hi Rico,
I've copied it from the same file (line 75). I can change it but that means we also need to cast the nullable types. For example an integer x with a default value of 100.

`int? x = null;`
`x is null ? 100 :  x  //This will fail...`
`x is null ? 100 : (int)x  //This will work...`

What do you prefer?

RicoSuter: `x != null ? x : 100`?

basvandrongelen: Same: CS1503	Argument 1: cannot convert from 'int?' to 'int'

RicoSuter: Ok, it's fine for me to keep as is. Can we merge then?

RicoSuter:
Can we merge if we'd keep the `??`?

Issue #2084, opened 2019-04-04
parameterDateTimeFormat is not part of the JsonSerializerSettings


danilsonStan:
For C# clients, it is possible to set the dateTime format string via "JsonSerializerSettings". 
It is applied single values. However, when there are dto objects, that contain dateTimes, the serialization uses the settings of the JsonSerializerSettings, but by default, there is not dateTime format specified. So the dateTime format is not always applied. 

I believe either Newton serializer settings should have the dateTime setting set in the constructor of the client or the dateTime format should be set via attribute on the property. Or am I doing something wrong?

RicoSuter:
The ParameterDateTimeFormat setting is only used for deserializing query parameters (Parameter* prefix) and in this context it would be wrong if it is also used in the JsonSerializerSettings… 

Isn't there a special json converter generated which handles the serialization/deserialization of date/time properties in DTOs? What exactly is the problem? Can you post a sample Swagger spec (or part of it) and the actual/expected output?

danilsonStan:
Thanks for your answer. 
Ahh i see, then it makes sense. A misunderstanding from my side. 
Unfortunately, we have a specific format for the dateTimes, but i believe/hope we might be able to solve it via the JsonSerializerSettings and the partial UpdateJsonSerializerSettings method.  (#2085 would also help in this)

Im going to some more tests on our side, that might take me a while though.

knoepdan:
I did some more investigation, regarding to the serialization of dateTime in the project we use nswag for. 

In the Open Api definition file, the concerned properties are defined as normal dateTime. However, the actual format that they are sent to is non standard and thus needs a custom serialization. Unfortunately, this custom format cannot be changed as it is outside of our control. Since I believe i misunderstood the actual meaning of ParameterDateTimeFormat and the fact that we can solve the issue by changing the serializer settings, i think this issue can be closed. 
(however, it would help us to if issue #2085 was done)

knoepdan:
Unfortunately, i have come accross an issue here, that i don't know to solve correctly yet. 

We have the following open api definition: 
```
"paths": {
    "/v1.0/somePath}/{fromDate}/{to}": {
      "get": {
        "tags": [ "example"  ],
        "parameters": [
          {
            "name": "fromDate",
            "in": "path",
            "schema": {
              "type": "string",
              "format": "date"
            },
          {
            "name": "to",
            "in": "path",
            "schema": {
              "type": "string",
              "format": "date-time"
            },
          }
```

[...]

Issue #1211, opened 2018-03-02
NSwag generates incorrect serialization for date parameters


williamyaoh-ah:
According to the Swagger 2.0 spec, ["type": "string", "format": "date" should](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#data-types) be formatted according to `full-date` [here](https://xml2rfc.tools.ietf.org/public/rfc/html/rfc3339.html#anchor14). (E.g. "2018-01-01") However, even when a request parameter is specified this way, NSwag serializes the date with a timestamp appended, like "2018-01-03T00:00:00".

Here's a minimal Swagger specification that exhibits this behavior:

```
{
    "swagger": "2.0",
    "info": {
        "version": "0.0.1.0",
        "title": "API"
    },
    "paths": {
        "/api/endpoint": {
            "post": {
                "consumes": [ "application/json" ],
                "parameters": [
                    { "name": "date",
                      "in": "body",
                      "schema": {
                          "type": "string",
                          "format": "date"
                      }
                    }
                ],
                "responses": {
                    "200": {
                    }
                }
            }
        }
    }
}
```

After generating a client from this specification using NSwag -> C# and running it against a local server, the parameter has a timestamp attached to the date.

RSuter:
Good point, we should probably generate a custom json converter and add it to the date properties.

https://stackoverflow.com/a/18639829/876814

Do you know if there is a way to correctly serialize these dates without having to generate the custom json converter?

williamyaoh-ah:
> Do you know if there is a way to correctly serialize these dates without having to generate the custom json converter?

Not that I know of. With the built-in date conversion base class in Newtonsoft, it doesn't seem like that much extra generated code to add though.

I remember having issues dealing with nested DateTimes when trying to customize Newtonsoft's JSON serialization in the past though; let me see if I can repro that, cuz getting *all* usages of a `date` formatted string in Swagger might be more complicated than at first glance.

williamyaoh-ah:
Yup, if the DateTimes are inside another data structure, things get ugly. Here's some example C# where the dates are in a List; compiles fine, errors at runtime:

```csharp
using System;
using System.Collections.Generic;

using Newtonsoft.Json;
using Newtonsoft.Json.Converters;

namespace Program {
  public class CustomDateTimeConverter : IsoDateTimeConverter {
    public CustomDateTimeConverter() {
      base.DateTimeFormat = "yyyy-MM-dd";
    }
  }

[...]

PR #2507, opened 2019-11-05
ParameterDateFormat support

knoepdan:
Currently, Serialization of C# DateTime objects when used as Parameters can be configured by the Parameter ParameterDateTimeFormat. However, sometimes the DateTime object should be treated as Date and not DateTime. 

Therefore, I added the parameter ParameterDateFormat that defines the Format when only the date is required. 

Example: 
```
"paths": {
    "/v1.0/somePath}/{fromDate}/{to}": {
      "get": {
        "tags": [ "example"  ],
        "parameters": [
          {
            "name": "fromDate",
            "in": "path",
            "schema": {
              "type": "string",
              "format": "date"
            },
          {
            "name": "to",
            "in": "path",
            "schema": {
              "type": "string",
              "format": "date-time"
            },
          }
```

Remark: 
Since, I'm not too familiar with NSWAG code structur, so it could be that this changes does not fit in codewise.  So could very well be if this change is not ok. 

Related issues: 
#1211  (also see: #2084)

### DIFF
--- a/src/NSwag.CodeGeneration.CSharp.Tests/ParameterTests.cs
+++ b/src/NSwag.CodeGeneration.CSharp.Tests/ParameterTests.cs
@@ -1,4 +1,5 @@
-﻿using NJsonSchema;
+﻿using System;
+using NJsonSchema;
 using Xunit;
 
 namespace NSwag.CodeGeneration.CSharp.Tests
@@ -111,7 +112,7 @@ namespace NSwag.CodeGeneration.CSharp.Tests
             var code = generator.GenerateFile();
 
             //// Assert
-            Assert.Contains("RemoveElementAsync(string x_User, System.Collections.Generic.IEnumerable<long> elementId, string secureToken)", code);          
+            Assert.Contains("RemoveElementAsync(string x_User, System.Collections.Generic.IEnumerable<long> elementId, string secureToken)", code);
         }
 
         [Fact]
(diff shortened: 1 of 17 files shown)


RicoSuter:
You are my hero today 🙂 i think you didnt miss a single place where you’d add this. The only thing missing to ensure the feature is working is a unit test. Will review tomorrow i hope.

knoepdan:
yes.. unit test is missing. Ill try to find the right way to implement one. (might take me some time).

knoepdan:
Added a unit test.. not sure if it is in the right place though.

knoepdan:
Great thank you. 
Another note, is there a possibility to have a reference api that is to be used for integration tests? 
It would then be possible to use the various generated clients to call the api and one could test the outcome (api might write to a file for example). 
Not sure if its a good or realistic idea but something like this would have quite a big test coverage. Inital setup is probably a pain though.

RicoSuter:
It would be great to have a full integration test, a reference api, then generate clients and then calling the api with the clients - all automatic in a test without publishing a new version. 

At the moment we only check the output of the generators but not whether they actually work

But this is quite some work

Itori:
In the typeScript generator, there is still no differences between dates and date-times
I've started to take a look at it, but it's more complicated as we can choose the datetime type.

It would require to use some logic already present in NJsonSchema to render the correct typeScript.

jkone27:
is this working already ?
i still don't see the new date time format applied when using visual studio add service reference, and passing this as parameters, am i doing something wrong?

does anyone have a way to verify in vs2019/windows ?

```
<ItemGroup>
    <!-- https://github.com/RicoSuter/NSwag/wiki/CSharpClientGeneratorSettings -->
    <OpenApiReference Include="OpenApi\MySpecWithDateTimes.json">
      <Options>/GenerateClientClasses:false /GenerateClientInterfaces:false /AllowReferencesWithProperties:true /ParameterDateTimeFormat:"yyyy-MM-ddTHH:mm:ss"</Options>
    </OpenApiReference>
  </ItemGroup>
```

generated properties are still in DateTimeOffset (offset is giving us trouble as it's not in date-time swagger default), and, don't see any format attribute applied to them

knoepdan:
ParameterDateTimeFormat and ParameterDateFormat refer to the dates that are in the path and thus not serialized by newtonsoft (or by some other json serializer in the latest core version). If I'm not mistaken, the format of dates that are in the body can be controlled by the serializer settings. 


PS: jkone27: you can contact me directly. I dont consider myself an expert on NSWAG but still might be able to help in this matter.

PR #2538, opened 2019-11-21
allow angular http request to send withCredentials

dominicshaw:
When generating a typescript client you may want to be able to add withCredentails: true to the options, i.e.

```cs
let options_ : any = {
    observe: "response",
    responseType: "blob",
    withCredentials: true,
    headers: new HttpHeaders({
        "Accept": "application/json"
    })
};
```

this change allows the user to select whether or not to send this option


RicoSuter:
Thanks for the PR!

PR #2600, opened 2019-12-31
change default array instance type for CSharpControllerGeneratorSettings

### LINKED ISSUE #2599: CSharpControllerGeneratorSettings uses different array type from CSharpGeneratorSettings.ArrayInstanceType
.NET Framework 4.6.1
NSwag 13.2.0.0

With following controller settings:
```csharp
			var controllerSettings = new CSharpControllerGeneratorSettings
			{
				ClassName = "PetstoreController",
				ControllerBaseClass = "Microsoft.AspNetCore.Mvc.Controller",
				CSharpGeneratorSettings =
				{
					Namespace = "Pri.Testing"
				}
			};
```
The resulting code has a compile error because the instance type of an array is Collection<T>, but the property type is List<T>:
```
CS0029	Cannot implicitly convert type 'System.Collections.ObjectModel.Collection<string>' to 'System.Collections.Generic.List<string>'	407	...
```
Setting the instance type to match explicitly resolves the error:
```csharp
			var controllerSettings = new CSharpControllerGeneratorSettings
			{
				ClassName = "PetstoreController",
				ControllerBaseClass = "Microsoft.AspNetCore.Mvc.Controller",
				CSharpGeneratorSettings =
				{
					ArrayInstanceType = "System.Collections.Generic.List",
					Namespace = "Pri.Testing"
				}
			};
```

peteraritchie:
fixes #2599

### DIFF
--- a/src/NSwag.CodeGeneration.CSharp/CSharpControllerGeneratorSettings.cs
+++ b/src/NSwag.CodeGeneration.CSharp/CSharpControllerGeneratorSettings.cs
@@ -20,6 +20,7 @@ namespace NSwag.CodeGeneration.CSharp
         {
             ClassName = "{controller}";
             CSharpGeneratorSettings.ArrayType = "System.Collections.Generic.List";
+            CSharpGeneratorSettings.ArrayInstanceType = "System.Collections.Generic.List",
             ControllerStyle = CSharpControllerStyle.Partial;
             ControllerTarget = CSharpControllerTarget.AspNetCore;
             RouteNamingStrategy = CSharpControllerRouteNamingStrategy.None;


src/NSwag.CodeGeneration.CSharp/CSharpControllerGeneratorSettings.cs
```diff
@@ -20,6 +20,7 @@ public CSharpControllerGeneratorSettings()
         {
             ClassName = "{controller}";
             CSharpGeneratorSettings.ArrayType = "System.Collections.Generic.List";
+            CSharpGeneratorSettings.ArrayInstanceType = "System.Collections.Generic.List",
```
RicoSuter: Inst the fallback already the value of ArrayType?

RicoSuter: Nope, according to
https://github.com/RicoSuter/NJsonSchema/blob/96d37dc3abf1c154aaffb9dab2ca8f5c1f5bbdc9/src/NJsonSchema.CodeGeneration.CSharp/CSharpGeneratorSettings.cs#L28
We need to change it here too

RicoSuter:
This is required. Thx for the pr.

PR #2606, opened 2020-01-07
missed .data property i.e the ArrayBuffer

AbuDawood:
converting to blob then to Text of response is incorrect. and returns [object object]  the correct var is the data property


RicoSuter:
Thanks for the PR!

Issue #2580, opened 2019-12-12
Add partial methods prior to serialization and deserialization to allow ignoring of required fields


michsco:
**CONTEXT**
We generate clients that get used against multiple instances of our swagger services (e.g. pre-prod and production). The client is generated against the pre-production environment, which means there may be new values in the responses that will fail to deserialize. To address this, we want to change the `JsonSerializerSettings` to ignore required fields during deserialization.

**ISSUE**
There are existing partial methods:
* `PrepareRequest(HttpClient, HttpRequestMessage, StringBuilder)`: Runs *after* deserialization of the request body (if any) before the URL is built.
* `PrepareRequest(HttpClient, HttpRequestMessage, string)`: Runs *after* deserialization of the request body (if any) and the final URL is built.
* `ProcessResponse(HttpClient, HttpResponseMessage)`: Runs *after* the response is received, but *before* deserialization of the response body.

If we implement the `ProcessResponse` method, we can alter the `JsonSerializerSettings` for the deserialization of the response, but there isn't a second place where we can *undo* the `JsonSerializerSettings` (we still want to enforce required fields when serializing the request).

**REQUEST**
Can we add two additional partial methods?

* `void BeforeRequestSerialization<T>(T body, JsonSerializerSettings settings)`: Runs immediately prior to calling `JsonConvert.SerializeObject(...)`.
* `void BeforeResponseDeserialization<T>(string json, JsonSerializerSettings settings)`: Runs immediately prior to calling `JsonConvert.DeserializeObject<...>(....)`.

RicoSuter:
Have you tried ExposeJsonSerializerSettings?

https://github.com/RicoSuter/NSwag/blob/master/src/NSwag.CodeGeneration.CSharp/CSharpClientGeneratorSettings.cs#L100

michsco:
I am trying to come up with a solution that doesn’t require changes to the callers for each operation.

With that in mind, it’s not possible to have two different serialization settings for the request and response, even if the serialization settings are exposed. I thought about changing the serialization settings after the response comes back (eg ProcessResponse), but a second operation call wouldn’t be able to change them back (ie require the fields) before the request is serialized.

Michael
Sent from Outlook<https://aka.ms/qtex0l> on iOS

________________________________

michsco:
@RicoSuter -- if I were to create a PR with a new partial method to solve this, I assume you would accept it?

RicoSuter:
[...]

PR #2607, opened 2020-01-07
Add partial methods to allow changes prior to deserialization

michsco:
**BUG FIX:** #2580

1. Add a new partial method, `BeforeRequestSerialization<T>`, which can change either the `JsonSerializerSettings` or the content itself *prior* to any serialization occurring.
2. Add a new partial method, `BeforeResponseDeserialization<T>`, which can change either the `JsonSerializerSettings` or the serialized JSON *prior* to any deserialization occurring.

The big benefit of these methods is the ability to have different rules or contract resolution between the request and response (e.g. able to ignore missing required properties on responses, but not requests).

### DIFF
--- a/src/NSwag.CodeGeneration.CSharp/CSharpClientGeneratorSettings.cs
+++ b/src/NSwag.CodeGeneration.CSharp/CSharpClientGeneratorSettings.cs
@@ -26,7 +26,7 @@ namespace NSwag.CodeGeneration.CSharp
             ParameterDateTimeFormat = "s";
             ParameterDateFormat = "yyyy-MM-dd";
             GenerateUpdateJsonSerializerSettingsMethod = true;
-            GeneratePreSerializationMethods = false;
+            UseRequestAndResponseSerializationSettings = false;
             QueryNullValue = "";
             GenerateBaseUrlProperty = true;
             ExposeJsonSerializerSettings = false;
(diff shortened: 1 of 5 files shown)


michsco:
Given that partial methods are compiled away if not implemented, there didn't seem to be a reason to make these settings driven unless it might create a conflict with something in the `Client.Class.Body.liquid` template.

I had to make a change to the project in order to properly trigger the partial method.

michsco:
In retrospect, after considering potential breaks that can be added by `Client.Class.Body.liquid`, I decided to add settings for the new partial methods.

I'll update the [CSharpClientGeneratorSettings](https://github.com/RicoSuter/NSwag/wiki/CSharpClientGeneratorSettings) and [NSwag Configuration Document](https://github.com/RicoSuter/NSwag/wiki/NSwag-Configuration-Document) wiki pages once this PR is merged.

RicoSuter:
> If we implement the ProcessResponse method, we can alter the JsonSerializerSettings for the deserialization of the response, but there isn't a second place where we can undo the JsonSerializerSettings (we still want to enforce required fields when serializing the request).

The problem with that is that it makes the client instance thread unsafe. We should only change the serializer settings in the ctor. Another solution to your problem would be to have two serializer settings: one for serialization and one for deserialization, each individually configurable.

michsco:
Fair feedback. I updated the PR to create two different Serialization Settings for request and response. This ignores the partial methods and also solves the underlying issue.

RicoSuter:
Thanks for the PR. But I have to say that I try to avoid such features which are specific to a particular scenario - they make the templates more and more complex and can also be implemented by just replacing templates locally for this specific project. Keep in mind that each new feature we add has to be maintained forever and this is probably done by me and not the original creator :-)

michsco:
I totally get that, which is why I did multiple rounds (on the different PRs and issues) to create a generic solution. This specific solution actually seems quite common as services and clients roll out separately, which is why I am a little surprised it hasn't been an issue in the past. Regardless, I appreciate all the help.

How long before this shows up in the NuGet packages?

RicoSuter:
> How long before this shows up in the NuGet packages?

In the next minutes:
https://github.com/RicoSuter/NSwag/pull/2620

RicoSuter:
> This specific solution actually seems quite common as services and clients roll out separately, which is why I am a little surprised it hasn't been an issue in the past. Regardless, I appreciate all the help.

Yep, it's ok to have this flexibility..

Issue #1398, opened 2018-06-19
Epic: Multiple file output for code generators


RSuter:

vgb1993:
Hy Rico, what's the status of this epic? 

In my company we are generating an Angular client from our Core API definition. Because all the services are generated inside the same file we can not code-split each service in a separate bundle. This is very inconvenient as you can imagine. 

We would like to have each service in it's own file and all the type definitions in a single file so there are no type duplicates.

I'm not familiar with the Liquid template notation, but I've read in [this issue](https://github.com/RicoSuter/NSwag/issues/239) that you could write to multiple files. Do you think this could be achived this way?

Thanks for your time

ttma1046:
@vgb1993 
My approach is 
1. put the auto-generated giant service and just itself in one single NgModule. (e.g. api-autogenerated-module).
2. ask different service/facade in different ngModules uses/DI that gaint service in api-autogenerated-module.

Jehoel:
I might also suggest running a `sed` (or T4 script if you're on Windows and have `msbuild` available) that takes the monolithic output and splits it up into multiple output files. This can be done with a simple regex.

vgb1993:
> I might also suggest running a `sed` (or T4 script if you're on Windows and have `msbuild` available) that takes the monolithic output and splits it up into multiple output files. This can be done with a simple regex.

Hey thanks for the reply. I already read about this option somewhere in this repo, but I was a bit skeptical. Do you have any examples? Also I have a question:

- What happens if an interface shared by two Services in different files? Will the interfaces be declared all in one file or duplicated in multiple files.

- In case all the interfaces where in the same file (wich would make sense to me) how would you generate the imports to the interfaces the Service uses?

vgb1993:
@ttma1046 

I don't like the idea of writing the Service twice (autogenerated / facade) only to allow code-splitt, this is why we use the generator. 

Also are you sure this allows code splitting? I beliebe if the services are in the same .ts file they always get bundled together.

Cheers

RicoSuter:
Internally the feature is almost complete. It's just not exposed via CLI, here you can see that internally we already have multiple "files" (artifacts) which are then merged and written to a single file in the CLI

https://github.com/RicoSuter/NSwag/blob/master/src/NSwag.CodeGeneration/ClientGeneratorBase.cs#L89

vgb1993:
Hy @RicoSuter,

[...]

PR #2614, opened 2020-01-09
[Ts] Add possibility to generate separate contracts & implementation 

trejjam:
Hi, I did solve part of the #1398.

You can use it from `nswag.json` file like:
```json5
{
  "codeGenerators": {
    "openApiToTypeScriptClient": {
      "output": null, // works in the same manner as before (internally mapped to the 'outputs' field)
      "outputs": {
        "Contracts": "./restApi.contract.ts",
        "Implementation": "./restApi.implementation.ts"
      }
    }
  }
}
```

### DIFF
--- a/src/NSwag.Commands/Commands/CodeGeneration/OpenApiToTypeScriptClientCommand.cs
+++ b/src/NSwag.Commands/Commands/CodeGeneration/OpenApiToTypeScriptClientCommand.cs
@@ -11,7 +11,10 @@ using System.Threading.Tasks;
 using NConsole;
 using NJsonSchema.CodeGeneration.TypeScript;
 using NJsonSchema.Infrastructure;
+using NSwag.CodeGeneration;
 using NSwag.CodeGeneration.TypeScript;
+using System.Collections.Generic;
+using System.IO;
 
 #pragma warning disable 1591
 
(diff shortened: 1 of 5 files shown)


trejjam:
In case you like it, I can extend it for the c# client generator (it uses separate fields.

trejjam:
I found an issue from the land of javascript. I do not know how easily import all types from another module. Probably it is possible to fix that in the generator (search for contracts in an imported namespace)?

RicoSuter:
This feature must be based completely on the CodeArtifact feature, see `GenerateTypes()` here:

https://apiverse.io/organizations/nuget-org/nuget-feeds/public/packages/NJsonSchema.CodeGeneration/versions/latest?type=NJsonSchema.CodeGeneration.GeneratorBase&namespace=NJsonSchema.CodeGeneration&assembly=netstandard-2.0

currently these types are just concatenated based on the type (contract / implementation) but with multiple file output we would generate one file per CodeArtifact with the correct imports from  BaseTypeName: 

https://apiverse.io/organizations/nuget-org/nuget-feeds/public/packages/NJsonSchema.CodeGeneration/versions/latest?type=NJsonSchema.CodeGeneration.CodeArtifact&namespace=NJsonSchema.CodeGeneration&assembly=netstandard-2.0

PR #2616, opened 2020-01-10
Add template for Client method annotations

michsco:
Adds a new template to expose Annotations for the Client methods (on both the interface and classes) for more consistency and to unblock certain scenarios.


michsco:
If I am able to check in this PR (and get it into the NuGet package), I am able to workaround the other PRs I have made.

PR #2633, opened 2020-01-17
Allow custom json-based content type

### LINKED ISSUE #2586: Use a custom json based content type
Hello,

In our project, we are using a custom content type to version our REST api. Something like `application/myapp.api.v1+json`. We have run into serveral issues with nswag because of this.

There are many places in the code, where the `application/json` content type is hardcoded. If I'm correct it cannot be overriden by any settings.

If I specify only the custom content type in the swagger yml like this
```
    post:
      consumes:
        - application/myapp.api.v1+json 
```
The generated post method contains input parameters of type `System.IO.Stream` instead of a strongly typed object. It's a bit panful. I believe it's determined in this code [OpenApiParameter.cs#L208-L258](https://github.com/RicoSuter/NSwag/blob/dc80568236dd4d2984ea6e034c616d61ce07057c/src/NSwag.Core/OpenApiParameter.cs#L208-L258)

To overcome this. I thought we can specify both, the `application/json` and our custom content type like this
```
    post:
      consumes:
        - application/myapp.api.v1+json 
        - application/json   
```
This works for the input parameters but we ran into second issue. The accept header is by default set to `application/json`. I would expect that the first of the consumes collection is taken into account. The code [OperationModelBase.cs#L242-L245](https://github.com/RicoSuter/NSwag/blob/85ae862fd6d68173a201a79e0ad06e0be2ec5de1/src/NSwag.CodeGeneration/Models/OperationModelBase.cs#L242-L245) forces the `application/json`. 

What about removing the lines [OperationModelBase.cs#L242-L245](https://github.com/RicoSuter/NSwag/blob/85ae862fd6d68173a201a79e0ad06e0be2ec5de1/src/NSwag.CodeGeneration/Models/OperationModelBase.cs#L242-L245) (and L258-L261) so the logic is only dependend on the order of content types?

Or do you have any other ideas how to overcome this? 

Thank you

FelixS90:
This PR allows for custom (vendor extension) json-based content types to be treated as json and not as binary. This fixes #2586.

Instead of using [StartsWith](https://docs.microsoft.com/en-us/dotnet/api/system.string.startswith) and [EndsWith](https://docs.microsoft.com/en-us/dotnet/api/system.string.endswith), one could also use [Regex.IsMatch](https://docs.microsoft.com/en-us/dotnet/csharp/how-to/search-strings#finding-specific-text-using-regular-expressions). However, I saw that the former is already used in OpenApiResponse.cs so I decided to keep it same-style.

### DIFF
--- a/src/NSwag.Core/OpenApiParameter.cs
+++ b/src/NSwag.Core/OpenApiParameter.cs
@@ -222,7 +222,7 @@ namespace NSwag
 
                 return consumes?.Any() == true &&
                        consumes.Any(p => p.Contains("application/xml")) &&
-                       consumes.Any(p => p.Contains("application/json")) == false;
+                       consumes.Any(p => p.StartsWith("application/") && p.EndsWith("json")) == false;
             }
         }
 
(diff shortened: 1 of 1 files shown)


RicoSuter:
LGTM, thanks for the PR

PR #2638, opened 2020-01-19
Add catch error to axiosclient template file

ZhangNanBin:


RicoSuter:
@Meberem I don't know Axios, is this correct?

Meberem:
Certainly looks like it based on the documentation: https://github.com/axios/axios#handling-errors

Sorry, I don't know the ins and outs of Axios either.

thijscrombeen:
A bit more background information would be nice. 
What do you see when you don't have this catch handler. 
Is anything other than 2XX not handled?

I belive by default Axios only resolves with 2XX.
Another option would be to set a global validateStatus on the axios client like so:
`
const validateStatus = function(status: number): boolean { return true; // all http states should result in a resolved promise , so NSwag generated code can handle it
  };
`

One small remark in your catch. Is it needed to wrap it in another Error object? Why not just rethrow the original error ?

Issue #2652, opened 2020-01-26
Serialization casing not honored for new System.Text.Json in .net core 3.1


groogiam:
Control over the serialization casing looks like it is broken when using the new System.Text.Json in .NET core 3.1

https://github.com/RicoSuter/NSwag/issues/2216 says casing is controlled by the AddJsonOptions which prior to ASP.NET Core 3.1 the JSON.NET implementation. 

When using the following (which utilizes System.Text.Json) the casing changes back to the default. 

```c#
.AddJsonOptions(o =>
                {
                    //preserve poco casing
                    o.JsonSerializerOptions.PropertyNamingPolicy = null;
                })
```
The expected result is that the casing of the .NET classes be maintained when this setting is used.

RicoSuter:
Not possible st the moment, you need to describe the serializer with the neetondoft json settings

https://github.com/RicoSuter/NJsonSchema/issues/1014

PR #2655, opened 2020-01-27
Read some System.Text.Json options

Sinhk:
Get global string enum serialization and casing (#2652) from System.Text.Json options.

When testing this I noticed the GetService<IOptions<MvcNewtonsoftJsonOptions>>() always returns an IOptions<MvcNewtonsoftJsonOptions>, as long as MvcNewtonsoftJsonOptions exists. It didn't have to be registered with the service provider. So I changed some of the code that checks if it should load options for Newtonsoft.Json or System.Text.Json to look for a NewtonsoftJsonOutputFormatter instead.


src/NSwag.Commands/Commands/Generation/OpenApiGeneratorCommandBase.cs
```diff
@@ -224,12 +224,14 @@ public async Task<TSettings> CreateSettingsAsync(AssemblyLoader.AssemblyLoader a
             JsonSerializerSettings serializerSettings;
             try
             {
-                var mvcJsonOptions = serviceProvider?.GetRequiredService<IOptions<MvcNewtonsoftJsonOptions>>();
-                serializerSettings = mvcJsonOptions?.Value?.SerializerSettings;
+                if (mvcOptions.OutputFormatters.Any(f => f.GetType() == typeof(Microsoft.AspNetCore.Mvc.Formatters.NewtonsoftJsonOutputFormatter)))
```
RicoSuter: Is this safe (i.e. always registered)? Also for .NET Core 1.0, etc.?
Or better: Why can we use this try/catch fallback implementation?

src/NSwag.Generation.AspNetCore/NSwag.Generation.AspNetCore.csproj
```diff
@@ -1,6 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>netstandard1.6;net451;netstandard2.0</TargetFrameworks>
+    <TargetFrameworks>netstandard1.6;net451;netstandard2.0;netcoreapp3.0</TargetFrameworks>
```
RicoSuter: Adding "netcoreapp3.0" is not an option as it would to an explosion of TFs... please use reflection as we did it for the other things... isnt System.Text.Json a .net standard 2.0 package too?

Sinhk: This would not necessarily lead to an explosion of TFs. Projects targeting .net core 3.1 (and current .net 5 nightly) uses the netcoreapp3.0 target if that is the highest available in the library. I have no idea if this will be true for future versions.

System.Text.Json is .net standard, but the ASP.NET Core integration is .net core >= 3.0. AFAIK System.Text.Json support is not relevant for previous versions.  
My main reason for having the netcoreapp3.0 target was to use the #if NETCOREAPP3_0. So it's probably not required for this PR. I'll remove it and try to use reflection if that is still what you prefer.

src/NSwag.Commands/Commands/Generation/OpenApiGeneratorCommandBase.cs
Sinhk: The try/catch fallback relies on the assumption that System.Text.Json is the default, and will only fallback to Newtonsoft when the Newtonsoft output formatter is registered. 
Since the Microsoft.AspNetCore.Mvc.NewtonsoftJson package is referenced by this project, I don't know of any case where this will throw an exception. But there is probably some, so better safe than sorry.

This code is inside an #if NETCOREAPP3_0, so I don't think Core 1.0 support is required.

Sinhk:
@RicoSuter Finally got around to redoing this. Tried to keep only what’s necessary for reading the options

RicoSuter:
Looks good. Is there a possibility to add unit tests for that somehow? Just to avoid regressions and have it tested.

Sinhk:
@RicoSuter Added some tests, checking if the options are set. 
I didn't find any test project targeting netcoreapp3 so I added a new one. Is that fine or should the tests rather be added to one of the existing projects?

src/NSwag.Generation.AspNetCore/AspNetCoreOpenApiDocumentGenerator.cs
```diff
@@ -127,15 +127,51 @@ public async Task<OpenApiDocument> GenerateAsync(ApiDescriptionGroupCollection a
...
+            {
+                var optionsAssembly = Assembly.Load(new AssemblyName("Microsoft.AspNetCore.Mvc.Core"));
+                var optionsType = typeof(IOptions<>).MakeGenericType(optionsAssembly.GetType("Microsoft.AspNetCore.Mvc.JsonOptions", true));
+                options = serviceProvider?.GetService(optionsType) as dynamic;
+            }
+            catch
+            {
+            }
+
+            if (!(options?.Value?.JsonSerializerOptions is System.Text.Json.JsonSerializerOptions jsonOptions))
```
RicoSuter, line 157: Isn't the JsonOptions always registered, even when Newtonsoft stuff is enabled?

src/NSwag.Generation.AspNetCore/AspNetCoreOpenApiDocumentGenerator.cs
```diff
@@ -127,15 +127,51 @@ public async Task<OpenApiDocument> GenerateAsync(ApiDescriptionGroupCollection a
...
+            {
+                return settings;
+            }
+
+            // StringEnumConverter
+            if (jsonOptions.Converters.Any(c => c is System.Text.Json.Serialization.JsonStringEnumConverter))
+            {
+                settings.Converters.Add(new Newtonsoft.Json.Converters.StringEnumConverter());
```
RicoSuter, line 165: Maybe we should also support this new attribute here: 
https://github.com/RicoSuter/NJsonSchema/blob/31d67d7276b88074f472377f94cd1fa8647b2953/src/NJsonSchema/Generation/DefaultReflectionService.cs#L386

RicoSuter:
Sorry for my late reply, added two comments, what do you think?

src/NSwag.Generation.AspNetCore/AspNetCoreOpenApiDocumentGenerator.cs
Sinhk: Looks like the System.Text.Json attribute is similar enough that adding an IsAssignableToTypeName for the new converter name should do the trick. I could test it and make a pull request

src/NSwag.Generation.AspNetCore/AspNetCoreOpenApiDocumentGenerator.cs
Sinhk: JsonOptions is always registered from AspNetCore 3.0, it was MvcJsonOptions before.
Since this is currently called as a fallback when Newtonsoft options are not found, JsonOptions should always be registered when it gets to this point. Assuming Newtonsoft options are always found when running on AspNetCore pre 3.0

src/NSwag.Generation.AspNetCore/AspNetCoreOpenApiDocumentGenerator.cs
RicoSuter: Exactly should be just an or with another check, maybe with namespace. PR would be great.

RicoSuter:
I think it would make sense to extract a ConvertJsonSettingsToNewtonsoftSettings method and move it to njsonschema.. then we just need to map as good as possible, eventually maybe with a custom contract resolver which reads system.text.json metadata.. i think this would be a good path forward and users of json schema would also benefit.. what do you think?

Sinhk:
Sounds like a good plan. I'll try to make a PR for that tomorrow.

RicoSuter:
Another thing: We should somehow turn of reading of newtonsoft attributes - in the newtonsoft settings 😀. See https://github.com/RicoSuter/NSwag/issues/2762

RicoSuter:
Working on this now in your branch.

src/NSwag.Generation.AspNetCore/AspNetCoreOpenApiDocumentGenerator.cs
RicoSuter: Done.

RicoSuter:
Can we merge this?

Sinhk:
Yes, looks good 🙂

Issue #2653, opened 2020-01-27
FileResponse isnt annotated with GeneratedCodeAttribute


ffwdq:
FileResponse isnt annotated with GeneratedCodeAttribute - this triggers code analysis during build for me. I'm using NSwagStudio v13.2.2.0

Heres the sample spec:
```json
{
  "swagger": "2.0",
  "info": {
    "title": "API",
    "version": "v2"
  },
  "paths": {
    "/api/file": {
      "post": {
        "tags": [
          "Mills"
        ],
        "operationId": "PostFile",
        "consumes": [
          "application/json",
          "text/json",
          "application/*+json"
        ],
        "produces": [
          "application/json"
        ],
        "responses": {
          "200": {
            "schema": {
              "format": "binary",
              "type": "string"
            }
          }
        }
      }
    }
  },
}
```
and here are the results:
```c#
//----------------------
// <auto-generated>
//     Generated using the NSwag toolchain v13.2.2.0 (NJsonSchema v10.1.4.0 (Newtonsoft.Json v11.0.0.0)) (http://NSwag.org)
// </auto-generated>
//----------------------

#pragma warning disable 108 // Disable "CS0108 '{derivedDto}.ToJson()' hides inherited member '{dtoBase}.ToJson()'. Use the new keyword if hiding was intended."
#pragma warning disable 114 // Disable "CS0114 '{derivedDto}.RaisePropertyChanged(String)' hides inherited member 'dtoBase.RaisePropertyChanged(String)'. To make the current member override that implementation, add the override keyword. Otherwise add the new keyword."
#pragma warning disable 472 // Disable "CS0472 The result of the expression is always 'false' since a value of type 'Int32' is never equal to 'null' of type 'Int32?'
#pragma warning disable 1573 // Disable "CS1573 Parameter '...' has no matching param tag in the XML comment for ...
#pragma warning disable 1591 // Disable "CS1591 Missing XML comment for publicly visible type or member ..."

namespace MyNamespace
{
    using System = global::System;
    
    [System.CodeDom.Compiler.GeneratedCode("NSwag", "13.2.2.0 (NJsonSchema v10.1.4.0 (Newtonsoft.Json v11.0.0.0))")]
    public partial class Client 
    {
        private string _baseUrl = "";
        private System.Net.Http.HttpClient _httpClient;
        private System.Lazy<Newtonsoft.Json.JsonSerializerSettings> _settings;
    
[...]

PR #2669, opened 2020-02-03
added code generated attribute to file response

ffwdq:
edited the template so the FileResponse is annotated with the GeneratedCodeAttribute
#2653


RicoSuter:
Thx for the pr.

PR #2689, opened 2020-02-14
Fixed issue for OpenAPI 3 where collectionFormat was deprecated

lukepfeiffer10:
As mentioned here for OAS3 specification https://swagger.io/docs/specification/describing-parameters/#query-parameters

Query parameters no longer support `collectionFormat` and instead should use the `style` and `explode` properties to describe array type formatting

This fixes the issue for OAS3 documents while keeping the existing functionality for Swagger (OAS2) documents

### DIFF
--- a/src/NSwag.Generation.Tests/OpenApiDocumentGeneratorTests.cs
+++ b/src/NSwag.Generation.Tests/OpenApiDocumentGeneratorTests.cs
@@ -1,0 +1,48 @@
+﻿using System.Linq;
+using System.Reflection;
+using Namotion.Reflection;
+using NJsonSchema;
+using NJsonSchema.Generation;
+using Xunit;
+
+namespace NSwag.Generation.Tests
+{
+    public class OpenApiDocumentGeneratorTests
+    {
+        public void HasArrayParameter(string[] foo) { }
+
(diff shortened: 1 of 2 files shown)


RicoSuter:
Do we also need to fix that in the client code generators? i.e. do they need to be changed so that they pick this up correctly?

lukepfeiffer10:
Oh good point. I'll check on that.

RicoSuter:
This might be part of NJsonSchema

RicoSuter:
Shouldnt we check whether we generate Swagger 2.0 or OpenAPI 3.0 and then use one or the other? Or is already handled?

RicoSuter:
And maybe we should add some unit tests for this to avoid regressions in the future.

lukepfeiffer10:
> Shouldnt we check whether we generate Swagger 2.0 or OpenAPI 3.0 and then use one or the other? Or is already handled?

I used some pre-existing functions that were specific to Swagger 2.0 and OpenAPI 3.0 to implement this change while maintaining backwards compatibility. Hopefully I can get some time this week to investigate your other notes and add some unit tests.

RicoSuter:
Is this PR ready to be reviewed again/merged?

PR #2701, opened 2020-02-25
Don't use DTO classes when generating interfaces

### LINKED ISSUE #2341: TypeScript + Axios + typeStyle: Interface results in static fromJS calls on Interface type
Using the Axios template to generate TypeScript with the typeStyle option set to Interface results in return types generated as interfaces, but the process methods still try to call fromJS on the interface like a static method.

Example generated code:

```typescript
    protected processUpdate(response: AxiosResponse): Promise<CustomerDto> {
        let _headers: any = {}; 
        if (response.headers && response.headers.forEach) { 
            response.headers.forEach((v: any, k: any) => _headers[k] = v);
        };
        if (status === 200) {
            let result200: any = null;
            result200 = CustomerDto.fromJS(resultData200); // <-- 'CustomerDto' only refers to a type, but is being used as a value here.ts(2693)
            return result200;
        }
        return Promise.resolve<CustomerDto>(<any>null);
    }
```

Meberem:
Fixes #2341. The Typescript `DataConversionCode` [relies on a class to be generated](https://github.com/RicoSuter/NSwag/blob/85ae862fd6d68173a201a79e0ad06e0be2ec5de1/src/NSwag.CodeGeneration.TypeScript/TypeScriptClientGenerator.cs#L132). If there is no need to call this, as is the case when using interfaces, we can just parse the JSON result.

### DIFF
--- a/src/NSwag.CodeGeneration.TypeScript/Templates/Client.ProcessResponse.HandleStatusCode.liquid
+++ b/src/NSwag.CodeGeneration.TypeScript/Templates/Client.ProcessResponse.HandleStatusCode.liquid
@@ -35,7 +35,11 @@ return response.blob().then(blob => { return { fileName: fileName, data: blob, s
 let result{{ response.StatusCode }}: any = null;
 {%         if Framework.IsAxios -%}
 let resultData{{ response.StatusCode }}  = _responseText;
+{%             if response.UseDtoClass -%}
 {{ response.DataConversionCode }}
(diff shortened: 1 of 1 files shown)


RicoSuter:
I think the initial idea of UseDtoClass == false is that they are not generated but you need to provide/write them yourself.

So this would be quite a breaking change.

RicoSuter:
v13.2.5

PR #2709, opened 2020-03-02
Merge 2xx responses with the same type (Issue #1259)

Trentm95:
I took a stab at #1259. 2xx responses with the same type as the primary success response are merged into a single block. Other 2xx responses use the fall through response. I ran the integration tests and from what I saw the output is the same as the current functionality when there is only a single 2xx response. Happy to make some changes if needed. Thanks!

Example generated code (From the two unit tests I added.)

C# Client

```csharp
...
var status_ = ((int)response_.StatusCode).ToString();
if (status_ == "200" || status_ == "201") 
{   
    var objectResponse_ = await ReadObjectResponseAsync<BusinessException>(response_, headers_).ConfigureAwait(false);
}
else
if (status_ == "204") 
{

    return default(BusinessException);
}
else
if (status_ == "404") 
{
    string responseText_ = ( response_.Content == null ) ? string.Empty : await response_.Content.ReadAsStringAsync().ConfigureAwait(false);
    throw new ApiException("Not Found", (int)response_.StatusCode, responseText_, headers_, null);
}
else
if (status_ != "200" && status_ != "204")
{
    var responseData_ = response_.Content == null ? null : await response_.Content.ReadAsStringAsync().ConfigureAwait(false); 
    throw new ApiException("The HTTP status code of the response was not expected (" + (int)response_.StatusCode + ").", (int)response_.StatusCode, responseData_, headers_, null);
}

return default(BusinessException);
...
```
TypeScript Client (Fetch)
```TypeScript 
...
if (status === 200 || status === 201) { 
    return response.text().then((_responseText) => {
    let result200: any = null;
    let resultData200 = _responseText === "" ? null : JSON.parse(_responseText, this.jsonParseReviver);
    result200 = BusinessException.fromJS(resultData200);
    return result200;
    });
} else if (status === 204) {
    return Promise.resolve<BusinessException>(<any>null);
} else if (status === 404) {
    return response.text().then((_responseText) => {
    return throwException("Not Found", status, _responseText, _headers);
    });
    return response.text().then((_responseText) => {
    });
}
return Promise.resolve<BusinessException>(<any>null);
...
```

### DIFF
--- a/src/NSwag.CodeGeneration.CSharp.Tests/ResponseTests.cs
+++ b/src/NSwag.CodeGeneration.CSharp.Tests/ResponseTests.cs
@@ -120,5 +120,69 @@ namespace NSwag.CodeGeneration.CSharp.Tests
             Assert.DoesNotContain("System.Threading.Tasks.Task<object>", code);
             Assert.Contains("class BusinessException", code);
         }
+
+        [Fact]
+        public async Task When_multiple_2xx_responses_of_same_type_they_are_merged_correctly()
+        {
+            // Arrange
+            var json = @"{
+  ""openapi"": ""3.0.0"",
+  ""paths"": {
+    ""/v1/exceptions/get"": {
+      ""post"": {
+        ""operationId"": ""Exceptions_GetException"",
+        ""responses"": {
+          ""200"": {
+            ""content"": {
+              ""application/json"": {
+                ""schema"": {
+                  ""$ref"": ""#/components/schemas/BusinessException""
+                }
+              }
+            }
+          },
+          ""201"": {
+            ""content"": {
+              ""application/json"": {
+                ""schema"": {
+                  ""$ref"": ""#/components/schemas/BusinessException""
+                }
+              }
+            }
+          },
+          ""204"": {
+            ""description"":""No Content""
+          },
+          ""404"": {
+            ""description"":""Not Found""
+          }
+        }
+      }
+    }
+  }, 
+  ""components"": {
+    ""schemas"": {
+      ""BusinessException"": {
+        ""type"": ""object"",
+        ""additionalProperties"": false
+      }
+    }
+  }
+}";
+
+            // Act
+            var document = await OpenApiDocument.FromJsonAsync(json);
+
+            //// Act
+            var settings = new CSharpClientGeneratorSettings { ClassName = "MyClass" };
+            var generator = new CSharpClientGenerator(document, settings);
+            var code = generator.GenerateFile();
+
+            // Assert
+            Assert.Contains(@"if (status_ == ""200"" || status_ == ""201"")", code);
+            Assert.Contains(@"if (status_ == ""204"")", code);
+            Assert.Contains(@"if (status_ == ""404"")", code);
+            Assert.Contains(@"if (status_ != ""200"" && status_ != ""204"")", code);
+        }
     }
 }
--- a/src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.liquid
+++ b/src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.liquid
@@ -263,7 +263,31 @@
                     ProcessResponse(client_, response_);
 
                     var status_ = ((int)response_.StatusCode).ToString();
-{%     for response in operation.Responses -%}
+{%     if operation.SameTypeSuccessResponses.size > 0 -%}
+                    if ({% for response in operation.SameTypeSuccessResponses -%}status_ == "{{ response.StatusCode }}"{% if response.CheckChunkedStatusCode %} || status_ == "206"{% endif %}{% if operation.SameTypeSuccessResponses.last.StatusCode != response.StatusCode %} || {% endif %}{% endfor -%}) 
+                    {   {% assign response = operation.SameTypeSuccessResponses.first %}
+                        {% template Client.Class.ProcessResponse %}
+                    }
+                    else
+{%     endif -%}
+{%     if operation.OtherSuccessResponses.size > 0 -%}
+                    if ({% for response in operation.OtherSuccessResponses -%}status_ == "{{ response.StatusCode }}"{% if response.CheckChunkedStatusCode %} || status_ == "206"{% endif %}{% if operation.OtherSuccessResponses.last.StatusCode != response.StatusCode %} || {% endif %}{% endfor -%}) 
+                    {
+{%         if operation.HasResultType -%}
+
+{%             if operation.WrapResponse and operation.UnwrappedResultType != "FileResponse" -%}
+                        return new {{ ResponseClass }}<{{ operation.UnwrappedResultType }}>((int)response_.StatusCode, headers_, default({{ operation.UnwrappedResultType }})); 
+{%             else -%}
+                        return default({{ operation.UnwrappedResultType }});
+{%             endif -%}
+{%         elseif operation.WrapResponse -%}
+
+                        return new {{ ResponseClass }}((int)response_.StatusCode, headers_); 
+{%         endif -%}
+                    }
+                    else
+{%     endif -%}
+{%     for response in operation.OtherResponses -%}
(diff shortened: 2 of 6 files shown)


Remixninja:
Just curious, what about 202? Only asking because I'm greedy and I want it included.

Trentm95:
> Just curious, what about 202? Only asking because I'm greedy and I want it included.

This should  handle any 2xx response. It uses this function to classify a response code as such. 
https://github.com/RicoSuter/NSwag/blob/de6de358cf7b3b56d57a9c47afffb74c4d1c5454/src/NSwag.Core/HttpUtilities.cs#L12-L21

Jogai:
I'm looking at this issue because I'm running into trouble with the ts client.
My api look like 

        [HttpPost]
        public async Task<ActionResult<MyType>> PostMy(MyType myType)
        {
            _context.MyTypes.Add(myType);
            await _context.SaveChangesAsync();

            return CreatedAtAction("GetMy", new { id = myType.Id }, myType);
        }

But on the return the 'response.text()' is invalid and throws an exception. I dont think this code accounts for that does it?

jeremyVignelles:
Hi @Trentm95 ,
It seems that I introduced some conflicts with your code, and I will likely create another PR with similar changes to be able to merge success response types.

Feel free to come and chat on gitter, we've already discussed with Rico about what the new feature should look like, but I'm sure there will be places that you can help with.

I'm curious about that part : https://github.com/RicoSuter/NSwag/pull/2709/files#diff-e3c05a0ed66970dd40f8b9de436b2831R273

How can it return something with a different type than the other sucesses ? Did I miss something ?

Thanks for the PR!

Trentm95:
> Hi @Trentm95 ,
> It seems that I introduced some conflicts with your code, and I will likely create another PR with similar changes to be able to merge success response types.
> 
> Feel free to come and chat on gitter, we've already discussed with Rico about what the new feature should look like, but I'm sure there will be places that you can help with.
> 
> I'm curious about that part : https://github.com/RicoSuter/NSwag/pull/2709/files#diff-e3c05a0ed66970dd40f8b9de436b2831R273
> 
> How can it return something with a different type than the other sucesses ? Did I miss something ?
> 
> Thanks for the PR!

Thanks for the reply @jeremyVignelles !

The intent with the "OtherSuccessResponses" was essentially to take any 2xx response that did not have the same type as the primary response (Relevant code https://github.com/RicoSuter/NSwag/pull/2709/files#diff-f24a7701007bec3c27fff8b8d4ab3470R147) and return that fall through section of code that was after all the "if else" blocks before the changes in #2959. I did this mostly for backwards compatibility with the way undeclared 204s worked (essentially the situation you mention as a breaking change in your PR) and also because I thought it would prevent the need for making the return type of the generated method nullable which would probably a better solution but possibly a breaking change. Reading through the comment you made on #1259 I may have missed some things with wrapped responses in that other success block and I do think what you proposed is cleaner.

jeremyVignelles:
Thanks for your reply. My main concern is that your code returns another type in the "Others" section. How could that compile?

Trentm95:
@jeremyVignelles The "OtherSuccessResponses" block you linked still returns the same ` operation.UnwrappedResultType` as the rest of the code just the `default` value of that type.

https://github.com/RicoSuter/NSwag/blob/83720b82016c1b24f2787bdb9139a119298884f8/src/NSwag.CodeGeneration/Models/OperationModelBase.cs#L104-L122


This is the method generated by this test https://github.com/RicoSuter/NSwag/pull/2709/files#diff-40f6347b03e522a42ce178d795a1d5b6R9 for example returning a type of `BusinessException`.
```csharp
public async System.Threading.Tasks.Task<BusinessException> GetExceptionAsync(System.Threading.CancellationToken cancellationToken)
{
    urlBuilder_.Append(BaseUrl != null ? BaseUrl.TrimEnd('/') : "").Append("/v1/exceptions/get");

    var client_ = _httpClient;
    try
    {
        {
            request_.Content = new System.Net.Http.StringContent(string.Empty, System.Text.Encoding.UTF8, "application/json");
            request_.Method = new System.Net.Http.HttpMethod("POST");
            request_.Headers.Accept.Add(System.Net.Http.Headers.MediaTypeWithQualityHeaderValue.Parse("application/json"));

            PrepareRequest(client_, request_, urlBuilder_);
            var url_ = urlBuilder_.ToString();
            request_.RequestUri = new System.Uri(url_, System.UriKind.RelativeOrAbsolute);
            PrepareRequest(client_, request_, url_);

            var response_ = await client_.SendAsync(request_, System.Net.Http.HttpCompletionOption.ResponseHeadersRead, cancellationToken).ConfigureAwait(false);
            try
            {
                var headers_ = System.Linq.Enumerable.ToDictionary(response_.Headers, h_ => h_.Key, h_ => h_.Value);
                if (response_.Content != null && response_.Content.Headers != null)
                {
                    foreach (var item_ in response_.Content.Headers)
                }

                var status_ = ((int)response_.StatusCode).ToString();
                if (status_ == "200" || status_ == "201")
                {
                    var objectResponse_ = await ReadObjectResponseAsync<BusinessException>(response_, headers_).ConfigureAwait(false);
                }
                else
                if (status_ == "204")
                {
                    return default(BusinessException);
                }
                else
                if (status_ == "404")
                {
                    string responseText_ = (response_.Content == null) ? string.Empty : await response_.Content.ReadAsStringAsync().ConfigureAwait(false);
                    throw new ApiException("Not Found", (int)response_.StatusCode, responseText_, headers_, null);
                }
                else
                if (status_ != "200" && status_ != "204")
                {
                    var responseData_ = response_.Content == null ? null : await response_.Content.ReadAsStringAsync().ConfigureAwait(false);
                    throw new ApiException("The HTTP status code of the response was not expected (" + (int)response_.StatusCode + ").", (int)response_.StatusCode, responseData_, headers_, null);
                }

                return default(BusinessException);
            }
            finally
            {
                if (response_ != null)
                    response_.Dispose();
            }
        }
    }
    finally
    {
    }
}
```
Though I might be misunderstanding as I am not very familiar with this code base.

jeremyVignelles:
Indeed, so if 200 returns T and 201 returns U, how is that handled?

Trentm95:
@jeremyVignelles 
200 will return T deserialized from the Response Body
201 will return default(T)

If wrap response is on then basically same thing except wrapped

jeremyVignelles:
Yes, so that means that 201 can't have a result object, because you are expecting 201 to return an empty response (or you discard the response). I still need to think about all the edge cases like this.

ErikPilsits-RJW:
> Indeed, so if 200 returns T and 201 returns U, how is that handled?

Should this be a supported scenario? In a well designed api I would not expect different response schemas.

To support multiple response types, I'm just brainstorming, I don't exactly love this idea.
- Methods return the wrapped response that includes the data as an `object`, the response code, and a Type property of the actual response.
- Caller handles the cast.

My initial thought on the aggregate implementation would be to extend `SwaggerResponse` (or create a new attribute like `SwaggerSuccessResponse`) to define additional 2xx responses and whether they expect response content (this would include 204). If yes, then it uses the type from the `ProducesResponse` attribute to deserialize, otherwise it returns `default`.

RicoSuter:
This story was about 

> Merge 2xx responses with the same type

which is an edge case where it would work with a single response type...

What @ErikPilsits-RJW is talking about is 2xx responses with different response types (which to me seems like a strange API design, yes) - this case cannot be nicely supported in C# as it does not have intersection types (`T | U`) like eg TypeScript or F#

jeremyVignelles:
I have implemented Type unions in https://github.com/jeremyVignelles/ooak , so this is theoretically possible to generate a method that deserializes to a `TypeUnion<T, U>`. That would probably mean some changes in NSwag to handle multiple return types though. Haven't looked at it, and to be honnest, I'm in the process of writing my own code generator.

codexnishant:
I dont think this handles for 202 status codes yet.

PR #2733, opened 2020-03-18
Implement TypeScriptClientGeneratorSettings.UseLeafType

trejjam:
This PR implement changes in NSwag related to https://github.com/RicoSuter/NJsonSchema/pull/1148.

It implements leaf types for client arguments. I do not think that leaf types can be useful for return types.

It can not pass before the NJsonSchema is updated.


RicoSuter:
Very nice PR. Notting missing I think. 
Will merge after NJS is updated.

RicoSuter:
v13.3.0

PR #2738, opened 2020-03-20
Fix TypeScriptClientGeneratorSettings

trejjam:
Sorry, It was a relict from the initial commit...


RicoSuter:
Too late, already fixed.

PR #2742, opened 2020-03-22
Nullable query parameters

l0lcat:
### Description
These commits make a valid OpenAPI 3 specification when query parameters originate from a complex type and ignores parameters in the query string for TypeScript clients when their value is null.

### OpenAPI 3 specification
OpenAPI 3 parameter before fix:
- `{
            "name": "id",
            "in": "query",
            "schema": {
              "type": "integer",
              "format": "int32",
              "nullable": true
            },
            "nullable": true
}`

OpenAPI 3 parameter after fix (nullable on root level removed, which is invalid):
- `{
            "name": "id",
            "in": "query",
            "schema": {
              "type": "integer",
              "format": "int32",
              "nullable": true
            }
}`
### TypeScript client
Generated TypeScript query string before fix:
- `http://something.com/foo?parameter=null`
Generated TypeScript query string after fix: 
- `http://something.com/foo`
### Related issues
I believe these commits fix these issues:
- [Question regarding client TS generation and nullable query params](https://github.com/RicoSuter/NSwag/issues/2000)
- ["Null" string passed to controllers, despite option for custom null value](https://github.com/RicoSuter/NSwag/issues/1830)

### DIFF
--- a/src/NSwag.CodeGeneration.TypeScript.Tests/TypeScriptOperationParameterTests.cs
+++ b/src/NSwag.CodeGeneration.TypeScript.Tests/TypeScriptOperationParameterTests.cs
@@ -8,10 +8,19 @@ namespace NSwag.CodeGeneration.TypeScript.Tests
 {
     public class TypeScriptOperationParameterTests
     {
-        public class OptionalParameterController
+        public class NullableParameterController
         {
             [Route("foo")]
             public string Test(int a, int? b)
+            {
+                return null;
+            }
+        }
+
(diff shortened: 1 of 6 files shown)


l0lcat:
@RicoSuter, I was wondering if this is going to be merged, or do I need to follow some procedure or something?

Kind regards,

Erik

RicoSuter:
Hope to have time to look into prs soon. But this one looks fine so far...

l0lcat:
Thanks! I was just checking if I missed something.

Issue #2772, opened 2020-04-03
Base interface for client interface in c# client generation


michauzo:
I have a following use case:
I'm creating clients using ClientBase class and I have some common methods exposed through the ClientBase. However, I'm using Clients through their generated interfaces. 
I've come up with an Interface to ClientBase class that can serve as a base for client interface. But this means I need to specify this manually every time I regenerate a client, as there's no such option to provide base interface for client interface.
I would like to add such option. I will work on a pull request for this.

That would mean following class hierarchy:
```C#
interface IClientBase {}
class ClientBase: IClientBase {}
interface IClient: **IClientBase** {}
class Client: ClientBase, IClient {}
```
Highlighted with stars is the part I'm missing generated.

mmmmmarco:
I'm running into exactly the same issue. Thank you for fixing this 👍

Do you have any updates when this will be released? The PR has been merged on 16 April 2020.

jeremyVignelles:
If the PR is merged, this issue must be closed @RicoSuter / @michauzo  .

The PR has been merge a long time ago, so it should already been there. If not, please open a new issue.

mmmmmarco:
It was a configuration issue on my side. The changes are in the latest release, yes. It wasn't mentioned in the release notes, so that confused me 😊

This issue can indeed be closed imo.

PR #2775, opened 2020-04-06
New option: define a base interface for C# client interfaces.

michauzo:
This adds an option discribed in issue #2772.

### DIFF
--- a/src/NSwag.CodeGeneration.CSharp.Tests/CSharpClientSettingsTests.cs
+++ b/src/NSwag.CodeGeneration.CSharp.Tests/CSharpClientSettingsTests.cs
@@ -111,5 +111,44 @@ namespace NSwag.CodeGeneration.CSharp.Tests
             //// Assert
             Assert.Contains("var client_ = new CustomNamespace.CustomHttpClient();", code);
         }
+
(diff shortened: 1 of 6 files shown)


RicoSuter:
What is the actual scenario which is solved by this?

michauzo:
It's decribed in the issue #2772.
The use case is as follows:
I have a base client class exposing what's needed in the client. 
'''
interface IClientBase {}
class ClientBase: IClientBase {}
interface IClient: IClientBase {}
class Client: ClientBase, IClient {}
'''
I have some common functionality for all clients (common endpoints, that are not covered in swagger specification). I'm exposing calls to these endpoints as extension methods for 'IClientBase'. I'm using 'Client' through an interface (injected). This means I need to have 'IClient' extending 'IClientBase'. Otherwise I'd need to cast these explicitly.
Currently I cannot get the 'IClient' extending 'IClientBase' in the generated code. That's the piece of functionality I'm adding here as an option for C# client generation. A user can specify a name for the base interface. If left empty - this option is disabled. The input doesn't show if client interface is not generated.

RicoSuter:
Thanks for the PR.

PR #2804, opened 2020-04-17
Throw an error if the supplied template directory does not exist

Meberem:
I had an issue where a build script was supplying an invalid path when building under docker, cue the "but it works on my machine". This pull request throws an error if the supplied template directory is not null (i.e. the user wants to use their own templates) but the folder doesn't exist.


RicoSuter:
Thanks for the PR.

PR #2806, opened 2020-04-17
Disable status validation (axios)

### LINKED ISSUE #2726: Axios client doesn't handle errors properly
Axios throws errors on 4xx/5xx status codes by default, so when using the generated client the actual API errors are never properly processed.

RicoSuter:
... so that exception handling is done by the NSwag generated client.

Closes #2726


MariuszKogut:
LGTM

PR #2819, opened 2020-04-23
Support query params expansion when explode is true and style is form

### LINKED ISSUE #2807: Generated C# client does not correctly encode query parameters for objects
Given the following openapi segment:

```json
"paths": {
  "/batches": {
    "get": {
      "parameters": [
          {
            "name": "filterSortPage",
            "in": "query",
            "required": true,
            "schema": {
              "$ref": "#/components/schemas/getBatchListParameters"
            },
            "style": "form",
            "explode": true
        }
      ]
    }
  },
}
```

where `getBatchListParameters` is an object defining several query parameters, the generated C# Client code for formatting this object as a query parameter string looks like this:

```csharp
// ...
urlBuilder_.Append(BaseUrl != null ? BaseUrl.TrimEnd('/') : "").Append("/batches?");
urlBuilder_.Append(System.Uri.EscapeDataString("filterSortPage") + "=").Append(System.Uri.EscapeDataString(ConvertToString(filterSortPage, System.Globalization.CultureInfo.InvariantCulture))).Append("&");
urlBuilder_.Length--;    
// ...
```

I was expecting the query string to consist of this object's properties where they query parameter names are consistent with my openapi spec (or the generated `Newtonsoft.Json.JsonProperty` annotations).

vadimi:
According to the spec object properties should be serialized to query string parameters if `explode: true` and `style: form`:

https://swagger.io/docs/specification/serialization/#query

For example ` {"role": "admin", "firstName": "Alex"}` should be serialized as `/users?role=admin&firstName=Alex`

This should resolve #2807

### DIFF
--- a/src/NSwag.CodeGeneration.CSharp.Tests/QueryParameterTests.cs
+++ b/src/NSwag.CodeGeneration.CSharp.Tests/QueryParameterTests.cs
@@ -1,0 +1,82 @@
+﻿using Xunit;
+
+namespace NSwag.CodeGeneration.CSharp.Tests
+{
+    public class QueryParameterTests
+    {
+        [Fact]
+        public void When_query_parameter_is_set_to_explode_and_style_is_form_object_parameters_are_expanded()
+        {
+            var spec = @"{
+  ""openapi"": ""3.0.0"",
+  ""info"": {
+    ""version"": ""1.0.0"",
+    ""title"": ""Query params tests""
+  },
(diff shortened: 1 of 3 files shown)


RicoSuter:
Thanks for the PR. So this was not supported at all before and this is not a breaking change?

vadimi:
yes, this wasn't supported and it's not a breaking change.

PR #2846, opened 2020-05-15
Add support for axios cancellation token

### LINKED ISSUE #2844: Add CancelToken to AxiosClient
Axios has build in support for canceling an ongoing call: https://github.com/axios/axios#cancellation It's necessary for some more or less complex fetching scenarios. So this feature should be supported by NSwag.

I'm still working on the required changes an will contribute a Pull-Request. Please hold the line... ;-)

MariuszKogut:
Fixes #2844 

I have also noticed that src/NSwag.CodeGeneration.TypeScript.Tests/AxiosTests.cs did not use the axios template. I have switched to axios to make sure, that everything works fine.

```
function isAxiosError(obj: any | undefined): obj is AxiosError {
  return obj && obj.isAxiosError === true
}
```

was tweaked, because I noticed, that the thrown error is undefined, if an ongoing was canceled.

Please let me know if I have to tweak something.

### DIFF
--- a/src/NSwag.CodeGeneration.TypeScript.Tests/AxiosTests.cs
+++ b/src/NSwag.CodeGeneration.TypeScript.Tests/AxiosTests.cs
@@ -40,7 +40,7 @@ namespace NSwag.CodeGeneration.TypeScript.Tests
             //// Act
             var codeGen = new TypeScriptClientGenerator(document, new TypeScriptClientGeneratorSettings
             {
-                Template = TypeScriptTemplate.Fetch,
+                Template = TypeScriptTemplate.Axios,
                 GenerateClientInterfaces = true,
                 TypeScriptGeneratorSettings =
                 {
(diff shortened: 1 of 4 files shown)


RicoSuter:
LGTM, thanks for the PR.

NewDark90:
@MariuszKogut Yep, I think there's a problem with it being passed into the options...

Example generation:
```
            method: "GET",
            url: url_,
            headers: {
                "Accept": "text/plain"
            },
            cancelToken
        };
```

It's subtle, and syntactically sound, but a problem.
`cancelToken` should be `cancelToken: cancelToken`.

cancelToken is valid as undefined, so the property not getting a value is valid, technically.

NewDark90:
Also, it would be cool if that bit checked to make sure that the CancelToken parameter exists before writing that out. If someone were to opt out of the functionality with `excludedParameterNames` or some other possible configuration, that shouldn't break.

RicoSuter:
@NewDark90 Can you create a PR with this change?

MariuszKogut:
Hey @NewDark90

I have tested this implementation in one of my applications using the cancel token behavior of Axios and it works fine. ``cancelToken`` in an object literal is a short hand and will be transpiled into ``cancelToken: cancelToken`` by typescript (when your target is es5):

image

A parameter opt out is not necessary in my point of view. It's an optional parameter therefore undefined will be passed and everything is as before. So in my point of view it's already opted out. ;-)

Any thoughts or did I missed something?

NewDark90:
Nope, all good. TIL

PR #2848, opened 2020-05-16
Demo: Add snapshooter

MariuszKogut:
During development of #2844 I missed an easy way to see the generated typescript file during writing a test. With this PR I want to demonstrate, how snapshot testing can help during development time and is also a kind of guarantee during a refactoring session.

I used https://github.com/SwissLife-OSS/snapshooter which is a small extensions and extends Xunit.

So please take a brief look and ping me, if you like it. I could integrate Snapshooter to any Generation-Tests.


MariuszKogut:
One more thing: If you want to use it in the future, consider to set SNAPSHOOTER_STRICT_MODE=true for a ci build

> In order to fail tests that are without a snapshot on your CI-build you can set the snapshooter behavior to strict-mode by setting the environment variable SNAPSHOOTER_STRICT_MODE to on or true.

MariuszKogut:
If you don't like this idea, feel free to close this PR. :-)

PR #2896, opened 2020-06-12
Aligned OpenAPI code generation for file more with OpenAPI specification

nicro950:
This PR modifies the multipart-form request for the request body schema to be an object instead of none, since this is more like the open api specification https://swagger.io/docs/specification/describing-request-body/file-upload/ 

This will also fix a problem with SwaggerUI where the browse button for file upload does not work, or does not appear.


RicoSuter:
Thanks for the PR.

Issue #2891, opened 2020-06-11
NSwag should take into account parameter position even if GenerateOptionalParameters is false


asvishnyakov:
https://github.com/RicoSuter/NSwag/blob/895e0ca52238de11868423b148283e67e475370f/src/NSwag.CodeGeneration.CSharp/Models/CSharpOperationModel.cs#L53-L82
https://github.com/RicoSuter/NSwag/blob/895e0ca52238de11868423b148283e67e475370f/src/NSwag.CodeGeneration.TypeScript/Models/TypeScriptOperationModel.cs#L39-L54

Parameters should be ordered by position before this `if`, otherwise `Position` is meaningless, because optional parameters are always last.

> Use case: for backward compatibility with Open API v2 I want to specify `x-position` for `requestBody` in v3, but body is optional. I can't do it required and I can't make it first without specifying `GenerateOptionaParameters` to `true`, which will move it to last, because optional parameters should be last, so it meaningless.

PR #2905, opened 2020-06-15
Fix parameters position calculation

asvishnyakov:
#2891


RicoSuter:
I think this is a breaking change and cannot be merged - also tests are failing.

asvishnyakov:
@RicoSuter I don't think failed tests are related to PR changes. Please check your tests and CI.

Yes, it's breaking changes, but without that fix `Position` parameter will work only if you have `GenerateOptionalParameters` option enabled, which is not useful, because optional parameters are always last, so you never can have your parameters correctly ordered. We probably need introduce new option for that case

asvishnyakov:
@RicoSuter As I said tests failed not because of these changes. After merging latest changes from master, everything passed. Please review and merge is possible or ask me for changes

RicoSuter:
I'm very afraid to break many people as this might reorder parameters.

PR #2908, opened 2020-06-17
Fix incorrect code generation for cancellation tokens

jacobpovar:
The non compile-time default value for CancellationToken added in https://github.com/RicoSuter/NSwag/pull/2888 resulted in compilation errors:

```
 Default parameter value for 'cancellationToken' must be a compile-time constant
```

This pr modifies generated code to compiler-friendly value.


azure-pipelines.yml
```diff
@@ -40,6 +40,13 @@ steps:
     script: 'dnt bump-versions preview "$(Build.BuildNumber)"'
     failOnStderr: true
 
+- task: DotNetCoreCLI@2
```
jacobpovar, line 43: This fixes build errors first appeared in https://dev.azure.com/rsuter/NSwag/_build/results?buildId=3706&view=results

goldsam:
My mistake on the breaking change. I agree with this fix. Sorry for the inconvenience.

PR #2912, opened 2020-06-18
Added a `#nullable enable` directive

jeremyVignelles:
Added a #nullable enable directive in generated C# source only if GenerateNullableReferenceTypes is set to true.

Added nullable reference annotations for generated code if needed.

Related to https://github.com/RicoSuter/NJsonSchema/pull/1201


jeremyVignelles:
I tried this with one of my projects and it generated a .cs file that can be built. However, I'm pretty sure I didn't test every case, so please tell me if anything is missing.

PR #2949, opened 2020-07-10
Add support for using controllers' summary as tag description

Leon99:
This will simplify migration from Swashbuckle (it always uses controllers' summary as tag description) while avoiding introducing a breaking change (new behaviour is off by default).


RicoSuter:
Thanks for the PR.

PR #2959, opened 2020-07-17
Reworked C# client response handling

### LINKED ISSUE #2944: C# client code returns default(T)! while result is not nullable
The implementation of NRT works great so far, but I've been bothered by the C# generated code :

```cs
public async System.Threading.Tasks.Task<ApiModel> GetModelAsync(System.Threading.CancellationToken cancellationToken)
{
    // --- snip ---
    try
    {
        {
            // --- snip ---    
            var response_ = await client_.SendAsync(request_, System.Net.Http.HttpCompletionOption.ResponseHeadersRead, cancellationToken).ConfigureAwait(false);
            try
            {
                // --- snip ---
                if (status_ == "200") 
                {
                    var objectResponse_ = await ReadObjectResponseAsync<ApiModel>(response_, headers_).ConfigureAwait(false);
                }
                else
                if (status_ != "200" && status_ != "204")
                {
                    var responseData_ = response_.Content == null ? null : await response_.Content.ReadAsStringAsync().ConfigureAwait(false); 
                    throw new ApiException("The HTTP status code of the response was not expected (" + (int)response_.StatusCode + ").", (int)response_.StatusCode, responseData_, headers_, null);
                }
    
                return default(ApiModel)!; // <-------------- HERE
            }
            finally
            {
                if (response_ != null)
                    response_.Dispose();
            }
        }
    }
    finally
    {
    }
}
```

The corresponding API:

```cs
[HttpGet("Model")]
[ProducesResponseType(typeof(ApiModel), 200)]
public ApiModel GetModel()
{
    return new ApiModel
    {
        Message = "Hello world!"
    };
}
```

The method explicitely returns `default(T)!`, which might not be an expected value (`null` while the expected output is not nullable).
Moreover, 204 is not an expected return code in my case, so that code should never be triggered.

Should the `if`/`else if`... change?
Should the unexpected status code exception be thrown outside the `else if` or in a `else` ?

I think that all the expected cases must be handled in the previous `if`/`else if`, and that is something remains there, it should throw because it's unexpected, but I'm pretty sure that if it is done that way, there must be a hidden reason, burried into all the options combination, so please tell me :) 

Another remark lies here:

```cs
var objectResponse_ = await ReadObjectResponseAsync<ApiModel>(response_, headers_).ConfigureAwait(false);
```

Due to the use of generics in ReadObjectResponseAsync, there are places where `!` is used, because we can't know if `T` is nullable or not.
Should we use `return objectResponse_.Object ?? throw new ...;` instead, to make sure that no nullable references are returned ?

jeremyVignelles:
This PR reworks how the response parsing works in the generated C# client:

- Keeps `status_` as an `int`, instead of applying `.ToString()` and comparing strings.
- Uses the `status_` where `(int)response_.StatusCode` was used
- Reordered code in the if/else sequence, fixes #2944 (see below)

Before:
```cs
var status_ = ((int)response_.StatusCode).ToString();
if (status_ == "200") 
{
    var objectResponse_ = await ReadObjectResponseAsync<ApiModel>(response_, headers_).ConfigureAwait(false);
}
else
if (status_ != "200" && status_ != "204")
{
    var responseData_ = response_.Content == null ? null : await response_.Content.ReadAsStringAsync().ConfigureAwait(false); 
    throw new ApiException("The HTTP status code of the response was not expected (" + (int)response_.StatusCode + ").", (int)response_.StatusCode, responseData_, headers_, null);
}

return default(ApiModel)!
```

The status code 200 was controlled twice, and `default(ApiModel)!` was returned while a null was not expected (#2944 )

The new code looks like this:

```cs
var status_ = (int)response_.StatusCode;
if (status_ == 200)
{
    var objectResponse_ = await ReadObjectResponseAsync<ApiModel>(response_, headers_).ConfigureAwait(false);
}
else
{
    var responseData_ = response_.Content == null ? null : await response_.Content.ReadAsStringAsync().ConfigureAwait(false); 
    throw new ApiException("The HTTP status code of the response was not expected (" + status_ + ").", status_, responseData_, headers_, null);
}
```

now, the `else` is the default case where the "unexpected HTTP status code" exception is thrown. There is no need to `return default(T)!;` anymore.

I tried to be careful not to break things, and as far as I know, there is only one thing that might break:
If there is an explicit success status code declared in the openApi definition, the `else if (status_ == 200 || status_ == 204)` that returns a `default(T)!` (or a wrapped instance) is not written anymore, and it will fall in the default `else` case.
In that case where a 200 is declared, but the server returns a 204 (undeclared), the code will no longer return `null` but will throw an "unexpected HTTP status code". I'd argue that this is cleaner and, in fact, expected : the fix is to declare the 204 on the server side.

### DIFF
--- a/src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.ConvertToString.liquid
+++ b/src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.ConvertToString.liquid
@@ -5,7 +5,7 @@ private string? ConvertToString(object? value, System.Globalization.CultureInfo 
 private string ConvertToString(object value, System.Globalization.CultureInfo cultureInfo)
 {%              endif -%}
 {
-    if (value is null)
+    if (value == null)
     {
         return null;
     }
--- a/src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.ConvertToString.liquid
+++ b/src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.ConvertToString.liquid
@@ -1,9 +1,15 @@
 ﻿{%              if GenerateNullableReferenceTypes -%}
+[return: System.Diagnostics.CodeAnalysis.NotNullIfNotNull("value")]
 private string? ConvertToString(object? value, System.Globalization.CultureInfo cultureInfo)
 {%              else -%}
 private string ConvertToString(object value, System.Globalization.CultureInfo cultureInfo)
 {%              endif -%}
 {
+    if (value is null)
+    {
+        return null;
+    }
+
     if (value is System.Enum)
     {
         var name = System.Enum.GetName(value.GetType(), value);
@@ -25,17 +31,18 @@ private string ConvertToString(object value, System.Globalization.CultureInfo cu
     }
     else if (value is bool) 
     {
-        return System.Convert.ToString(value, cultureInfo)?.ToLowerInvariant();
+        return System.Convert.ToString((bool)value, cultureInfo).ToLowerInvariant();
     }
     else if (value is byte[])
     {
         return System.Convert.ToBase64String((byte[]) value);
     }
-    else if (value != null && value.GetType().IsArray)
+    else if (value.GetType().IsArray)
     {
         var array = System.Linq.Enumerable.OfType<object>((System.Array) value);
         return string.Join(",", System.Linq.Enumerable.Select(array, o => ConvertToString(o, cultureInfo)));
     }
 
-    return System.Convert.ToString(value, cultureInfo);
+    var result = System.Convert.ToString(value, cultureInfo);
+    return (result is null) ? string.Empty : result;
 }
--- a/src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.ProcessResponse.liquid
+++ b/src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.ProcessResponse.liquid
@@ -7,7 +7,7 @@ client_ = null; response_ = null; // response and client are disposed by FileRes
 return fileResponse_;
 {%         else -%}
 var objectResponse_ = await ReadObjectResponseAsync<{{ response.Type }}>(response_, headers_).ConfigureAwait(false);
-throw new {{ ExceptionClass }}("{{ response.ExceptionDescription }}", status_, objectResponse_.Text, headers_, objectResponse_.Object);
+throw new {{ ExceptionClass }}<{{ response.Type }}>("{{ response.ExceptionDescription }}", status_, objectResponse_.Text, headers_, objectResponse_.Object, null);
 {%         endif -%}
 {%     elsif response.IsPlainText -%}
 var responseData_ = response_.Content == null ? null : await response_.Content.ReadAsStringAsync().ConfigureAwait(false);
@@ -21,6 +21,12 @@ return result_;
 {%         endif -%}
 {%     else -%}
 var objectResponse_ = await ReadObjectResponseAsync<{{ response.Type }}>(response_, headers_).ConfigureAwait(false);
+{%         if response.IsNullable == false -%}
+if (objectResponse_.Object == null)
+{
+    throw new {{ ExceptionClass }}("Response was null, but return type is not declared as nullable", status_, objectResponse_.Text, headers_, null);
+}
+{%         endif -%}
 {%         if response.IsSuccess -%}
 {%             if operation.WrapResponse -%}
 return new {{ ResponseClass }}<{{ operation.UnwrappedResultType }}>(status_, headers_, objectResponse_.Object); 
--- a/src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.ProcessResponse.liquid
+++ b/src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.ProcessResponse.liquid
@@ -2,19 +2,19 @@
 {%     if response.IsFile -%}
 {%         if response.IsSuccess -%}
 var responseStream_ = response_.Content == null ? System.IO.Stream.Null : await response_.Content.ReadAsStreamAsync().ConfigureAwait(false);
-var fileResponse_ = new FileResponse((int)response_.StatusCode, headers_, responseStream_, {% if InjectHttpClient %}null{% else %}client_{% endif %}, response_); 
+var fileResponse_ = new FileResponse(status_, headers_, responseStream_, {% if InjectHttpClient %}null{% else %}client_{% endif %}, response_); 
 client_ = null; response_ = null; // response and client are disposed by FileResponse
 return fileResponse_;
 {%         else -%}
 var objectResponse_ = await ReadObjectResponseAsync<{{ response.Type }}>(response_, headers_).ConfigureAwait(false);
-throw new {{ ExceptionClass }}("{{ response.ExceptionDescription }}", (int)response_.StatusCode, objectResponse_.Text, headers_, objectResponse_.Object);
+throw new {{ ExceptionClass }}("{{ response.ExceptionDescription }}", status_, objectResponse_.Text, headers_, objectResponse_.Object);
 {%         endif -%}
 {%     elsif response.IsPlainText -%}
 var responseData_ = response_.Content == null ? null : await response_.Content.ReadAsStringAsync().ConfigureAwait(false);
 var result_ = ({{ response.Type }})System.Convert.ChangeType(responseData_, typeof({{ response.Type }}));
 {%         if response.IsSuccess -%}
 {%             if operation.WrapResponse -%}
-return new {{ ResponseClass }}<{{ operation.UnwrappedResultType }}>((int)response_.StatusCode, headers_, result_); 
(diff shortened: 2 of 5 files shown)


src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.ProcessResponse.liquid
```diff
@@ -2,63 +2,69 @@
...
 {%         if response.IsSuccess -%}
 var responseStream_ = response_.Content == null ? System.IO.Stream.Null : await response_.Content.ReadAsStreamAsync().ConfigureAwait(false);
-var fileResponse_ = new FileResponse((int)response_.StatusCode, headers_, responseStream_, {% if InjectHttpClient %}null{% else %}client_{% endif %}, response_); 
+var fileResponse_ = new FileResponse(status_, headers_, responseStream_, {% if InjectHttpClient %}null{% else %}client_{% endif %}, response_); 
 client_ = null; response_ = null; // response and client are disposed by FileResponse
 return fileResponse_;
 {%         else -%}
 var objectResponse_ = await ReadObjectResponseAsync<{{ response.Type }}>(response_, headers_).ConfigureAwait(false);
-throw new {{ ExceptionClass }}("{{ response.ExceptionDescription }}", (int)response_.StatusCode, objectResponse_.Text, headers_, objectResponse_.Object);
+throw new {{ ExceptionClass }}<{{ response.Type }}>("{{ response.ExceptionDescription }}", status_, objectResponse_.Text, headers_, objectResponse_.Object, null);
```
jeremyVignelles, line 10: Unless I'm missing something obvious, there was a bug here:
The non-generic exception class doesn't take an object as a parameter, so I added the response type generic argument.
The constructor of that generic class takes an inner exception parameter that must be filled, be it with `null`.

Do you agree?

src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.ProcessResponse.liquid
```diff
@@ -2,63 +2,69 @@
...
+return new {{ ResponseClass }}<{{ operation.UnwrappedResultType }}>(status_, headers_, objectResponse_.Object); 
 {%             else -%}
 return objectResponse_.Object; 
 {%             endif -%}
 {%         endif -%}
 {%         if response.IsSuccess == false -%}
 {%             if response.InheritsExceptionSchema -%}
 var responseObject_ = objectResponse_.Object != null ? objectResponse_.Object : new {{ response.Type }}();
-responseObject_.Data.Add("HttpStatus", status_);
+responseObject_.Data.Add("HttpStatus", status_.ToString());
```
jeremyVignelles, line 40: That's the only place where a string is needed (admittedly, I didn't even test without the ToString()). All other places should work fine (if not better) with ints.

src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.liquid
```diff
@@ -284,28 +284,44 @@
...
-                    if (status_ != "200" && status_ != "204")
-                    {
-                        var responseData_ = response_.Content == null ? null : await response_.Content.ReadAsStringAsync().ConfigureAwait(false); 
-                        throw new {{ ExceptionClass }}("The HTTP status code of the response was not expected (" + (int)response_.StatusCode + ").", (int)response_.StatusCode, responseData_, headers_, null);
-                    }
-{%     endif -%}
-{%     if operation.HasDefaultResponse == false or (operation.DefaultResponse.HasType == false and operation.HasSuccessResponse == false) -%}
-{%         if operation.HasResultType -%}
-        
+{%        elseif operation.HasResultType -%}
```
jeremyVignelles, line 289: I admit that this is quite hard to review with the diff tools.
I took the part inside `if operation.HasDefaultResponse == false or (operation.DefaultResponse.HasType == false and operation.HasSuccessResponse == false)` and moved that to the two separate places where it was needed.

This duplicates some code, but in the end, I find it clearer and this allows the generated code to be cleaner.

Then, see below for the second major change

src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.liquid
```diff
@@ -284,28 +284,44 @@
...
+                    return new {{ ResponseClass }}(status_, headers_); 
+{%         endif -%}
+{%     else -%}
+{%         if operation.HasSuccessResponse == false -%}
+{% comment %} 
+    If the success response has already been explicitely declared, there is no need for this default code (because handled above).
+    Otherwise, return default values on success because we don't want to throw on "unknown status code".
+    Success is always expected
+{% endcomment -%}
+                    if (status_ == 200 || status_ == 204)
```
jeremyVignelles, line 305: The logic has been inverted there, and the second code is pasted in here. (and only enabled if no success code has been handled yet)
The unexpected throw is in the final else.

src/NSwag.CodeGeneration.CSharp/Templates/File.liquid
```diff
@@ -74,7 +93,11 @@ namespace {{ Namespace }}
             get { return StatusCode == 206; }
         }
 
+{%          if GenerateNullableReferenceTypes -%}
+        public FileResponse(int statusCode, System.Collections.Generic.IReadOnlyDictionary<string, System.Collections.Generic.IEnumerable<string>> headers, System.IO.Stream stream, System.IDisposable? client, System.IDisposable? response)
```
jeremyVignelles, line 97: The Stream is not nullable since a NullStream is given here upstream.

src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.ConvertToString.liquid
```diff
@@ -1,9 +1,15 @@
 ﻿{%              if GenerateNullableReferenceTypes -%}
+[return: System.Diagnostics.CodeAnalysis.NotNullIfNotNull("value")]
 private string? ConvertToString(object? value, System.Globalization.CultureInfo cultureInfo)
 {%              else -%}
 private string ConvertToString(object value, System.Globalization.CultureInfo cultureInfo)
 {%              endif -%}
 {
+    if (value is null)
```
RicoSuter, line 8: I try to use C# 6 and below (I think) isnt this a more recent feature?

src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.ProcessResponse.liquid
RicoSuter: Looks like a bug to me, yes. Before the object param would needed an exception => compile error.

src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.ProcessResponse.liquid
```diff
@@ -2,63 +2,69 @@
...
 {%             else -%}
 return result_; 
 {%             endif -%}
 {%         endif -%}
 {%     else -%}
 var objectResponse_ = await ReadObjectResponseAsync<{{ response.Type }}>(response_, headers_).ConfigureAwait(false);
+{%         if response.IsNullable == false -%}
+if (objectResponse_.Object == null)
+{
+    throw new {{ ExceptionClass }}("Response was null, but return type is not declared as nullable", status_, objectResponse_.Text, headers_, null);
```
RicoSuter, line 27: Maybe change it to "Response was null which was not expected."

src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.liquid
```diff
@@ -267,9 +267,9 @@
 
                     ProcessResponse(client_, response_);
 
-                    var status_ = ((int)response_.StatusCode).ToString();
+                    var status_ = (int)response_.StatusCode;
 {%     for response in operation.Responses -%}
-                    if (status_ == "{{ response.StatusCode }}"{% if response.CheckChunkedStatusCode %} || status_ == "206"{% endif %}) 
+                    if (status_ == {{ response.StatusCode }}{% if response.CheckChunkedStatusCode %} || status_ == 206{% endif %})
```
RicoSuter, line 272: This here is exactly the problem why it's a string => The OpenAPI status code might not be a number but "2XX" or "default" which now might lead to compile errors: 

https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.3.md#responses-object

So now we probably need to fix that.

src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.ConvertToString.liquid
jeremyVignelles: According to https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/is#constant-pattern , it is C#7, will fix.

src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.ProcessResponse.liquid
jeremyVignelles: Done

RicoSuter:
Just checked with some projects, lgtm. 
Thank you very much for the PR and sorry for the delay (holidays).

Issue #2950, opened 2020-07-10
FileParameter class definition missing when controller has an API with List<IFormFile>


gilzig:
Hi,

It took some time to get the scenario right, but I see that if I have a net core controller with only one API for uploading multiple files, for example:
```
       [SwaggerOperation(OperationId = "UploadMultipleFile")]
        [Consumes("multipart/form-data")]
        [Produces("application/json")]
        [ProducesResponseType((int)HttpStatusCode.OK, Type = typeof(Dictionary<string, bool>))]
        [ProducesResponseType((int)HttpStatusCode.BadRequest, Type = typeof(string))]
        [HttpPost("upload/{sessionId}")]
        public IActionResult UploadMultipleFile(string sessionId, List<IFormFile> files)
        {
                        ........
        }
```
Then nswag studio does not generate the FileParameter class definition.
However, if I add another API which received a single IFormFile (not a list), then the FileParameter class definition is correctly added. For example, if I add another API as below then it will create the class definition:

```
        [SwaggerOperation(OperationId = "UploadSingleFile")]
        [Consumes("multipart/form-data")]
        [Produces("application/json")]
        [ProducesResponseType((int)HttpStatusCode.OK, Type = typeof(bool))]
        [ProducesResponseType((int)HttpStatusCode.BadRequest, Type = typeof(string))]
        [HttpPost("uploadsingle/{sessionId}")]
        public IActionResult UploadSingleFile(string sessionId, IFormFile file)
        {
                  ........
        }
```

Please let me know what you think,

Best,
Gil

alexdresko:
Confirmed.

Issue #2797, opened 2020-04-15
IFormFile collections not binding to typescript correctly


gjhurray:
We have various methods in our api controllers that take `IFormFile` as a parameter. The resulting typescript changes the `IFormFile` to a `FileParameter`. However, when I try and change the api methods to take a parameter of type `IFormFileCollection` (or similar), the resulting typescript parameter is always a string array.

For example, I have this method:

`[HttpPost]`
`[ProducesResponseType(typeof(bool), 200)]`
`[Route("uploadFile")]`
`public bool UploadFile(IFormFile file, ScanDocumentCategory category) {`

Which produces the following typescript:

`uploadFile(file: FileParameter | null | undefined, category: ScanDocumentCategory): Observable<boolean> {`

Which is great, as the `FileParameter` type lets me do what I want. 
However, when I try and change my api method to take multiple files like so:

`[HttpPost]`
`[ProducesResponseType(typeof(bool), 200)]`
`[Route("uploadFiles")]`
`public bool UploadFiles(IFormFileCollection files, ScanDocumentCategory category) {`

The resulting typescript looks like this:

`uploadFiles(files: string[] | null | undefined, category: ScanDocumentCategory): Observable<boolean> {`

I have tried using `IEnumerable<IFormFile>`, `List<IFormFile>`, `IFormFile[]`, with no success. 
I have also tried creating a wrapper C# class to hold multiple IFormFiles and changing the api method to take this new class as a parameter, but the resulting class in typescript still has the files stored as a `string[]`.

Thank you.

DenisJNewb:
Im facing this issue too. Any advices?

gjhurray:
I ended up calling my api method once per file, but using rxjs` finalize()` I could wait until the last one had finished before showing the user the results:

```
let uploadResultMessages: string[] = []; // Collection of result messages from each file upload

range(0, this.files.length).pipe(
    concatMap(i => {
        let fileContents = <FileParameter>{ data: this.files[i], fileName: this.files[i].name };
        return this.api.uploadFile(fileContents, this.scanDocumentCategory);
    }),
    finalize(() => {
        // do stuff with uploadResultMessages
    }),
    .subscribe(resultMessage => uploadResultMessages.push(resultMessage)); // where resultMessage is message from uploadFile method

```

DenisJNewb:
My solution is better. I threw nswag

RicoSuter:
Working on this in the next days

Issue #2259, opened 2019-06-21
Issue with IFormFile and IFormFileCollection using AddOpenApiDocument()


skolmer:
Thanks for this great project! 
Unfortunately I have some issues with the Open API document generator in ASP.NET Core. 
If I use `AddOpenApiDocument` instead of `AddSwaggerDocument`, all `IFormFile` and `IFormFileCollection` parameters are exported as `string` or `ICollection<string>`. 
If I switch to `AddSwaggerDocument` the generated document has the parameter type `file` and the generated client code uses `FileParameter` instead of `string`, which is the expected behavior.

Is this a limitation in the OpenAPI Spec or the implementation or did I miss some configuration detail?
thanks in advance.

Znow:
We are facing a similar issue, but here FileParameter is generated when we use IFormFile in C#. What is FileParameter? it is no existing class anywhere...

jeremyVignelles:
On the controller side, you declare a Stream or a IFormFile parameter (Stream + file name). The Stream class doesn't serialize correctly into json, so you can't use it on the client side, so NSwag generates a FileParameter class.

Znow:
The IFormFile is also generated as a FileParameter class... Which is nowhere to be found, not on any page on the internet. What is this class, why is it generated, and why can it not be found?

jeremyVignelles:
If you can't serialize a Stream, you can't serialize something that contains a stream (IFormFile), so FileParameter is generated on your C# client file.

IFormFile -> FileParameter is the expected behavior, IFormFile -> string isn't.

@skolmer : Can you post a repro project somewhere? I've never seen such issue.

Znow:
@jeremyVignelles That didnt quite answer my question: "What is this class, why is it generated, and why can it not be found?" FileParameter does not exist anywhere...

jeremyVignelles:
I mean that the class is in your C# generated code (if you are using the C# generator).
It is declared like this : https://www.github.com/RicoSuter/NSwag/tree/master/src%2FNSwag.CodeGeneration.CSharp%2FTemplates%2FFile.liquid#L32

It can't really be defined as such because:
- It's not exactly the same class between C# and Typescript
- It's a partial class, so you can add your own behavior
- You can tell NSwag not to generate it and provide your own implementation.

As for the why it is being generated, I think I answered the question properly, but let me correct something. IFormFile is not serialized as JSON. That still stands:
[...]

PR #2972, opened 2020-07-31
Fix for File Parameter interface not generated in some instances

craigngu:
- Split the Binary tests into 3 so each endpoint can be generated properly without missing the FileParameter interface
- changed RequiresFileParameterInterface logic so it checks isBinary on arrays and nested objects

Fixes [#2973](https://github.com/RicoSuter/NSwag/issues/2973)

### DIFF
--- a/src/NSwag.CodeGeneration.TypeScript.Tests/BinaryTests.cs
+++ b/src/NSwag.CodeGeneration.TypeScript.Tests/BinaryTests.cs
@@ -58,7 +58,7 @@ components:
         }
 
         [Fact]
-        public async Task WhenSpecContainsFormData_ThenFormDataIsUsedInTypeScript()
+        public async Task WhenSpecContainsFormDataInSingleMultipartFile_ThenFormDataIsUsedInTypeScript()
         {
             var json = @"{
   ""x-generator"": ""NSwag v13.5.0.0 (NJsonSchema v10.1.15.0 (Newtonsoft.Json v11.0.0.0))"",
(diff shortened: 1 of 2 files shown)


jeremyVignelles:
Do you know which issues are fixed by this PR? I've seen a lot of FileParameter-related issues lately

craigngu:
For this PR its trying to fix the issue that the FileParameter interface in Typescript generated clients are missing when
1. There are no endpoints that have a single file input

So What I did is to check one layer deeper for the isBinary tags so now it will appear when there are
1. Array of files
2. File input under a nested structure 

Check the unit tests I've changed for examples, 
the original unit test combined all 3 use cases in one test, but the second and third endpoint won't work properly without the first one existing.

jeremyVignelles:
I mean, please link the issues resolved by your PR in the PR description like `Fixes #`.

craigngu:
I can create an issue for this, couldn't find any that matches this

jeremyVignelles:
#2950 ?
#2259 ?
#2797 ?

Just looked at the title, don't know if it's really related, I'm on mobile.

jeremyVignelles:
Your fix seems typescript specific, but would it be possible to include a fix for other issues as well?

craigngu:
yeah my fix is typescript specific for my use case, non of the ones referenced above are really related.

I don't think I understand the other issues fully atm, and its pretty different to how we were using it on the controller side

RicoSuter:
Thanks for the PR.

ElgersNiels:
I'm encountering this still for v14.5.0, is this a regression or am I missing something.

NSwag.AspNetCore 14.5.0
NSwag.CodeGeneration.TypeScript 14.5.0

NSwagStudio 14.5.0

PR #2983, opened 2020-08-05
fix for FileParameter class not being generated for FormFile collections

### LINKED ISSUE #2950: FileParameter class definition missing when controller has an API with List<IFormFile>

Sinhk:
Same as #2972, but for C# clients

Should fix #2950

### DIFF
--- a/src/NSwag.CodeGeneration.CSharp.Tests/BinaryTests.cs
+++ b/src/NSwag.CodeGeneration.CSharp.Tests/BinaryTests.cs
@@ -165,10 +165,10 @@ components:
         }
 
         [Fact]
-        public async Task WhenSpecContainsFormData_ThenFormDataIsUsedInCSharp()
+        public async Task WhenSpecContainsFormDataInSingleMultipartFile_ThenFormDataIsUsedInCSharp()
         {
             var json = @"{
-  ""x-generator"": ""NSwag v13.5.0.0 (NJsonSchema v10.1.15.0 (Newtonsoft.Json v11.0.0.0))"",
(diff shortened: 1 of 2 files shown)


emilmork:
This would solve our problem with generating clients with file collections. 

Do you take a look @Sinhk ?

Thanks for a greate library 👌

RicoSuter:
Thanks for the PR.

PR #2989, opened 2020-08-07
Added option to use standard C# naming conventions (and more)

kendallb:
Added options to use standard C# naming conventions as well as the ability to generate a client without the Async suffix on functions when Async is the only API method supported. Also trimmed new lines from all the .liquid files to avoid conflicts on dangling spaces in diffs in the future.

### DIFF
--- a/src/NSwag.CodeGeneration.CSharp/CSharpGeneratorBaseSettings.cs
+++ b/src/NSwag.CodeGeneration.CSharp/CSharpGeneratorBaseSettings.cs
@@ -6,17 +6,19 @@
 // <author>Rico Suter, mail@rsuter.com</author>
 //-----------------------------------------------------------------------
 
+using System.Reflection;
 using Newtonsoft.Json;
 using NJsonSchema;
 using NJsonSchema.CodeGeneration;
 using NJsonSchema.CodeGeneration.CSharp;
-using System.Reflection;
 
 namespace NSwag.CodeGeneration.CSharp
 {
     /// <summary>Settings for the <see cref="CSharpGeneratorBase"/>.</summary>
     public abstract class CSharpGeneratorBaseSettings : ClientGeneratorBaseSettings
     {
+        private bool _cSharpNamingConvention;
+
         /// <summary>Initializes a new instance of the <see cref="CSharpClientGeneratorSettings"/> class.</summary>
         protected CSharpGeneratorBaseSettings()
         {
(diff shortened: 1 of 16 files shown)


kendallb:
Just checked in support for automatic rate limit handling (http status code 429) with the UI to optionally enable it.

src/NSwag.CodeGeneration.CSharp/CSharpGeneratorBaseSettings.cs
```diff
@@ -49,6 +51,22 @@ protected CSharpGeneratorBaseSettings()
...
+            get => _cSharpNamingConvention;
+            set
+            {
+                _cSharpNamingConvention = value;
+
+                // Install the C# style name generators if desired
+                if (!_cSharpNamingConvention) return;
+                CSharpGeneratorSettings.PropertyNameGenerator = new CSharpPropertyNameGenerator();
+                CSharpGeneratorSettings.TypeNameGenerator = new CSharpTypeNameGenerator();
+                CSharpGeneratorSettings.EnumNameGenerator = new CSharpEnumNameGenerator();
```
RicoSuter, line 66: I think these three implementations should not be added here but we should change the defaults in NJsonSchema. This just adds another layer of complexity and if you need to do that you already have the CSharpGeneratorSettings class as the extension point. Please create a PR in NJsonSchema with your improvements.

kendallb:
I believe when I did this, it was not possible to modify the naming generator in NJsonSchema to support this, because the generator stuff is all handles in NSwag itself? Or are you suggesting that the default naming convention in NJsonSchema be changed to use correct C# naming conventions and get rid of the options and overrides to change it?

RicoSuter:
> Or are you suggesting that the default naming convention in NJsonSchema be changed to use correct C# naming conventions and get rid of the options and overrides to change it?

Yes, i wonder how much we’d need to break?

kendallb:
Well I am all for making it the default naming convention since the other way was wrong :) But I do wonder if other folks using it might do a double take if they can't go back to what it was before?

fsamiec:
are there any plans to resolve this? me and @christianerbsmehl would love to use nswag, but that naming issue is a showstopper for us.

kendallb:
I would like to see this merged in also. We are still using our own custom build of NSwag as it works for the purposes we need it for with all my patches. You can use my tree and build against that as an alternative?

https://github.com/kendallb/NSwag

PR #2996, opened 2020-08-12
Fixes for NuGet package cache

### LINKED ISSUE #1646: Add local Nuget's cache folder to reference paths resolves to the wrong path
Trying to fix some problems that probably are due to the work in progress state of the API explorer generation support I tried and enabled "Add local Nuget's cache folder to reference paths" (both for Reflection based generation and API explorer. Let's ignore the latter).

I get the error ```System.IO.DirectoryNotFoundException: Could not find a part of the path 'D:\Users\Ben\.nuget\packages'```.

That folder doesn't exist. Compare:

```
C:\Users\Ben>dotnet nuget locals all --list
info : http-cache: C:\Users\Ben\AppData\Local\NuGet\v3-cache
info : global-packages: C:\Users\Ben\.nuget\packages\
info : temp: C:\Users\Ben\AppData\Local\Temp\NuGetScratch
info : plugins-cache: C:\Users\Ben\AppData\Local\NuGet\plugins-cache
```
and
```
C:\Users\Ben>echo %USERPROFILE%/.nuget/packages
C:\Users\Ben/.nuget/packages
```

Where is the D:\ (that drive does exist) coming from? How does NSwag resolve that path if not using either of the two ways above?

AndrewMcLachlan:
Respects the NUGET_PACKAGES environment variable and falls back to the USERPROFILE environment variable instead of HOMEPATH on Windows machines.
Fixes #1646.

### DIFF
--- a/src/NSwag.Commands/Commands/IsolatedCommandBase.cs
+++ b/src/NSwag.Commands/Commands/IsolatedCommandBase.cs
@@ -87,9 +87,14 @@ namespace NSwag.Commands
 
         private static string[] LoadDefaultNugetCache()
         {
-            var envHome = RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? "HOMEPATH" : "HOME";
-            var home = Environment.GetEnvironmentVariable(envHome);
-            var path = Path.Combine(home, ".nuget", "packages");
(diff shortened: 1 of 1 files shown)


RicoSuter:
Thanks for the PR.

PR #3000, opened 2020-08-12
In process gui code generation

kendallb:
Reference this discussion:

https://github.com/RicoSuter/NSwag/issues/2988

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -72,3 +72,6 @@ src/packages/**
 .editorconfig
 /samples/packages
 /src/NSwag.Console/Properties/launchSettings.json
+
+# Ignore files from JetBrainds Rider
+/src/.idea/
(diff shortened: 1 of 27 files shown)


RicoSuter:
Please do not change newtonsoft versions as this might break a lot of users...

kendallb:
I believe it was only a minor version bump but mostly trying to make it consistent across the application. I don't think it was changed for anything in user libraries though? I thought I left it alone for the Schema library.

kendallb:
Actually it looks like I did change it from 10.0.2 to 12.0.3, however the main reason I did it was to fix the warnings related to different versions of Json.net across the solution. A lot of the executables were still referencing 9.0.1. I can see that some of those are nuget packages, so I will simply revert that change and let you decide how to fix the conflicting versions.

Issue #1607, opened 2018-09-14
Generating Controller Methods with WrapResponses - true, for Asp.Net Core


alex-leroux:
Hi,
If I use SwaggerToCSharpControllerGenerator, with SwaggerToCSharpControllerGeneratorSettings.WrapResponses = true, to generate a controller method for the following definition:

		"/api/v2/contact/getcontactsbyname":
		{
			"post":
			{
				"tags":
				[
					"Contact"
				],
				"operationId": "Contact_GetContactsByName",
				"parameters":
				[
					{
						"name": "names",
						"in": "body",
						"required": true,
						"schema":
						{
							"type": "array",
							"items":
							{
								"type": "string"
							}
						},
						"x-nullable": true
					}
				],
				"responses":
				{
					"200":
					{
						"x-nullable": true,
						"description": "",
						"schema":
						{
							"type": "array",
							"items":
							{
								"$ref": "#/definitions/Contact"
							}
						}
					}
				}
			}
		},

the resulting methods appears as follows:

        [Microsoft.AspNetCore.Mvc.HttpPost, Microsoft.AspNetCore.Mvc.Route("api/v2/contact/getcontactsbyname")]
        public async System.Threading.Tasks.Task<HttpResponseMessage> GetContactsByName([Microsoft.AspNetCore.Mvc.FromBody] System.Collections.Generic.IEnumerable<string> names)
        {
            var result = await _implementation.GetContactsByNameAsync(names).ConfigureAwait(false);
    
            var status = (System.Net.HttpStatusCode)result.StatusCode;
            HttpResponseMessage response = Request.CreateResponse(status, result.Result);
    
            foreach (var header in result.Headers)
                response.Headers.Add(header.Key, header.Value);
    
            return response;
        }

This code works in Asp.Net, but not in **Asp.Net Core** because there is no Request.CreateResponse method.

Is there currently a way to generate code for Asp.Net Core?

It should probably look something like this:

        [Microsoft.AspNetCore.Mvc.HttpPost, Microsoft.AspNetCore.Mvc.Route("api/v2/contact/getcontactsbyname")]
        public async System.Threading.Tasks.Task<Microsoft.AspNetCore.Mvc.IActionResult> GetContactsByName([Microsoft.AspNetCore.Mvc.FromBody] System.Collections.Generic.IEnumerable<string> names)
        {
            var result = await _implementation.GetContactsByNameAsync(names).ConfigureAwait(false);
    
            var status = result.StatusCode;
[...]

PR #3004, opened 2020-08-13
Add AspNetCore WrapResponse in Controllers.liquid

Jeroen-VdB:
Added support for WrapResponse in AspNetCore Controllers based on the example in #1607


RicoSuter:
Thx for the PR.

Issue #2981, opened 2020-08-04
Bug: Client will be disposed in FileResponse but "Dispose HttpClient" is not activated


DominikPf:
We are having some Issues in a generated NSwag Client with File Downloads.

If i create a NSwag Client over NSwagStudio and don't check the box here:

![image](https://user-images.githubusercontent.com/20923469/89294372-b9765a80-d65f-11ea-84d3-df557b8e54f8.png)

The Client will be Disposed in FileResponse:
![image](https://user-images.githubusercontent.com/20923469/89294427-ceeb8480-d65f-11ea-867e-fb1ec73a5b4a.png)
In my oppinion this should not happen if i unchecked this option. Or do i understand something wrong?

DominikPf:
I created a PR for this #3009 

@RicoSuter Can you please check if everything is alright and merge the PR? 
As for now i will use my own Build of NSwagStudio

rbedell99:
This issue still occurs on 13.9.

![image](https://user-images.githubusercontent.com/20542305/98965523-c4971d00-24cf-11eb-8cb3-921f172d8234.png)
![image](https://user-images.githubusercontent.com/20542305/98965577-d24ca280-24cf-11eb-80d2-bead71ef3d8e.png)
![image](https://user-images.githubusercontent.com/20542305/98965646-e6909f80-24cf-11eb-93f3-c5a37ca8f4dc.png)

PR #3009, opened 2020-08-17
Prevent dispose of HttpClient in FileResponse

DominikPf:
This PR adresses #2981 

The if Statement is the same as in the Client Template (Check for InjectHttpClient == false and DisposeHttpClient == true)

https://github.com/RicoSuter/NSwag/blob/fb48ea9e48ef8ccacf9e7b8fa35af1f1dfa0c81b/src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.liquid#L336-L339

### DIFF
--- a/src/NSwag.CodeGeneration.CSharp/Templates/File.liquid
+++ b/src/NSwag.CodeGeneration.CSharp/Templates/File.liquid
@@ -111,8 +111,10 @@ namespace {{ Namespace }}
             Stream.Dispose();
             if (_response != null)
                 _response.Dispose();
+{%     if InjectHttpClient == false and DisposeHttpClient -%}
             if (_client != null)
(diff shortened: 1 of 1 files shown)


RicoSuter:
> The if Statement is the same as in the Client Template

Where is this already present?

DominikPf:
I meant the Dispose in the Client Template (Marked Lines in Client.liquid as mentioned in my comment above)

jeremyVignelles:
Sorry, I didn't see your fix at first when creating #3016. The way things were written made me think that the proper fix is to pass `null` the FileResult constructor, instead of testing in the FileResult template.

DominikPf:
@jeremyVignelles No Problem, i'm ok with that. I am closing this PR in favor of #3016

PR #3015, opened 2020-08-18
Fixed potential NullReferenceException due to a race condition

jeremyVignelles:
While using the CommandLine project, my VisualStudio hit a breakpoint because this was not initialized in `TryLoadByVersion`.

This will likely help too if the assembly is annotated with NRT someday.


bbarry:
This code is not remotely thread safe. Are you able to get a call stack for that breakpoint to figure out why TryLoadByVersion is called before AssemblyLoader.RegisterReferencePaths gets called (the only method in the code that sets this property)?

jeremyVignelles:
Possibly, I didn't look deep into it.
I should probably document the change better, but IIRC, this code is called from AssemblyLoader.RegisterReferencePaths, before the assignment (inside the GetDirectory or something).
IIRC, it was trying to load the System.IO assemblies.
Will have a look later if needed

PR #3016, opened 2020-08-18
Fixed code generation for FileResponse

### LINKED ISSUE #2981: Bug: Client will be disposed in FileResponse but "Dispose HttpClient" is not activated

jeremyVignelles:
When enabling NullReferenceTypes generation with an API that can return a FileResponse, the code tried to set `client_ = null` and `response_ = null`, which is now disallowed.

Instead, I introduce two booleans `disposeClient_` and `disposeResponse_` that controls whether the variables must be disposed at the end.

By fixing this issue, I found a that the FileResponse gets the client only if InjectHttpClient is false, which is not consistent with this condition in the Client.Class.liquid:
```cs
{%     if InjectHttpClient == false and DisposeHttpClient -%}
```

I decided to fix it there, and I then found #3009 and #2981 . This fixes #2981.

Note : This PR is based on #3015, the first commit exists in both branches.

### DIFF
--- a/src/NSwag.AssemblyLoader/CustomAssemblyLoadContext.cs
+++ b/src/NSwag.AssemblyLoader/CustomAssemblyLoadContext.cs
@@ -26,7 +26,7 @@ namespace NSwag.AssemblyLoader
 
         internal Dictionary<string, Assembly> Assemblies { get; } = new Dictionary<string, Assembly>();
 
-        internal List<string> AllReferencePaths { get; set; }
+        internal List<string> AllReferencePaths { get; set; } = new List<string>();
 
         public Assembly Resolve(AssemblyName args)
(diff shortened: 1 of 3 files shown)


RicoSuter:
Thx for the PR

dltnbs:
My HttpClient is not disposable. Previously an injected client had no `Dispose` call generated against it. Now it is generated, and being wrapped in an `if (false)` isn't enough to stop it failing to compile.

PR #3030, opened 2020-08-30
Add type="module" into script html tag for CustomJavaScriptPath 

Rynaret:
Adds the possibility to use more advanced additional custom javascript files.


RicoSuter:
Thx, i hope this is not a breaking change?

Rynaret:
Actually, it can give some disadvantages:
1. A module is a file. To make import/export work, browsers need <script type="module">. Modules have several differences:
    - Deferred by default.
    - Async works on inline scripts.
    - To load external scripts from another origin (domain/protocol/port), CORS headers are needed.
    - Duplicate external scripts are ignored.
2. Modules have their own, local top-level scope and interchange functionality via import/export.
3. Modules always use strict.
4. Module code is executed only once. Exports are created once and shared between importers.
(from here https://javascript.info/modules-intro#summary)

Add a new merge request that gives more options https://github.com/RicoSuter/NSwag/pull/3053.

PR #3043, opened 2020-09-09
Use file parameter when multiple mime types are allowed (CSharp)

RicoSuter:


jakubbloksa:
Hi, is there a way to still use Stream instead of FileParameter even if there are multiple mime types?

PR #3061, opened 2020-09-16
Fixed complex parameter generation for custom json content types.

dhowardimpinj:
This PR restores support for mime-type parameters for json-based content types. I takes the suggestion from the previous PR #2633 and uses [Regex.IsMatch](https://docs.microsoft.com/en-us/dotnet/csharp/how-to/search-strings#finding-specific-text-using-regular-expressions). Additionally, OpenApiResponse.cs is updated to have consistency as was mentioned in the previous PR as well.


RicoSuter:
Looks good to me, Thx

Issue #2392, opened 2019-09-06
TypeScript baseUrl behaviour - coalescence FTW!


johnnyreilly:
Hello!

First of all, thanks for all your fine work on NSwag. Just lately my team has made the journey from handrolling clients to generating TypeScript and C# clients for our platform using your marvellous tooling.

We bumped on one issue related to how the baseUrl for TypeScript clients is set. It's slightly related to this issue:

https://github.com/RicoSuter/NSwag/issues/1198

But subtly different. It goes like this: when we're developing our application locally we're serving up the front end of our app at http://localhost:8080 - webpack dev server. However, our API is being served up by ASP.NET Core at http://localhost:5000

The generated clients contain this line in the generated client code:

```
    this.baseUrl = baseUrl ? baseUrl : "http://localhost:5000";
```

This code is problematic for one of our use cases. We have 2 potential `baseUrl`s that we `new` up clients with. 

1. When developing locally we pass `"http://localhost:5000"`. 

2. When deployed our complete app, API, HTML, CSS and JavaScript generated from TypeScript all get served from the same location. So in that case we pass an empty string: `""`.

The first use case is fine. The second is problematic because of the behaviour of `this.baseUrl = baseUrl ? baseUrl : "http://localhost:5000";`. Namely that this is a falsey check. If `baseUrl` is `null` or `undefined` then we fall back to use `"http://localhost:5000"`. That's desired behaviour. However, we also do the same for other falsey values *including* empty strings.

You can probably see where I'm going with this :smile:

Would you be open to a PR that changed the constructor generated code in this case to something like this:

```
    this.baseUrl = baseUrl !== undefined && baseUrl !== null ? baseUrl : "http://localhost:5000";
```

This is code that I believe covers the intent of the current generated code but does it in a more specific way.

If you're open to this I'd be happy to try pick this up.

pglx:
any updates?? I have the same issue!

johnnyreilly:
No response as yet - I'm still happy to raise a PR, but I don't want to do that if it's not desired

PR #3065, opened 2020-09-19
TypeScript generate clients: this.baseUrl to cater for empty strings

johnnyreilly:
Hello!

First of all, thanks for all your fine work on NSwag. My team has happily made the journey from handrolling clients to generating TypeScript and C# clients for our platform using your marvellous tooling.

We bumped on one issue related which I raised in #2392.

The generated clients contain this line in the generated client code:

```
this.baseUrl = baseUrl ? baseUrl : "http://fallbackUrl";
```

This code is problematic in the case where empty strings are supplied; this relates to JavaScript's `false`-y behaviour (where empty strings are treated as `false`). Consider the following 2 cases: 

1. When developing locally we pass a `baseUrl` of `"http://localhost:5000"`. - **this is fine**

```ts
// Development mode - the supplied baseUrl is "http://localhost:5000"  
this.baseUrl = baseUrl ? baseUrl : "http://fallbackUrl"; //"http://localhost:5000" - yay!
```

2. When deployed, our complete app: API, HTML, CSS and JavaScript generated from TypeScript all get served from the same location. So in that case we pass a `baseUrl` of an empty string: `""`. - **this has a problem**

```ts
// Production mode - the supplied baseUrl is ""  
this.baseUrl = baseUrl ? baseUrl : "http://fallbackUrl"; //"http://fallbackUrl" - oh no! we wanted to use the empty string; that was the behaviour we needed!
```

The first use case is fine. The second is problematic because of the *false-y* check. If `baseUrl` is `null` or `undefined` then we fall back to use `"http://fallbackUrl"`. That's desired behaviour. However, we also do the same for other falsey values *including* empty strings. That's troublesome!

You can probably see where I'm going with this :smile:

This PR changes the constructor in generated TypeScript clients to this:

```
this.baseUrl = baseUrl !== undefined && baseUrl !== null ? baseUrl : "http://fallbackUrl";
```

This does the same as the previous code but crucially treats empty strings as legitimate values. This is a good thing :tm: 

The behaviour is similar to the nullish coalescing operator which is now part of JS / TS: https://github.com/microsoft/TypeScript/issues/26578 / https://github.com/tc39/proposal-nullish-coalescing

Having this merged would remove our current workaround of doing some `string.Replace` on our generated TypeScript client code to get to this place.

What do you think?

Thanks for all your work by the way - we really love it!

cc @pglx

### DIFF
--- a/src/NSwag.CodeGeneration.TypeScript/Templates/AngularClient.liquid
+++ b/src/NSwag.CodeGeneration.TypeScript/Templates/AngularClient.liquid
@@ -26,9 +26,9 @@
 {%     endif -%}
         this.http = http;
 {%     if UseGetBaseUrlMethod -%}
-        this.baseUrl = baseUrl ? baseUrl : this.getBaseUrl("{{ BaseUrl }}");
+        this.baseUrl !== undefined && baseUrl !== null ? baseUrl : this.getBaseUrl("{{ BaseUrl }}");
 {%     else -%}
-        this.baseUrl = baseUrl ? baseUrl : "{{ BaseUrl }}";
+        this.baseUrl !== undefined && baseUrl !== null ? baseUrl : "{{ BaseUrl }}";
 {%     endif -%}
     }
 {% endif -%}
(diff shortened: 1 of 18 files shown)


RicoSuter:
Makes sense, I hope we dont break someone with this...
Thanks for the PR.

PR #3101, opened 2020-10-09
[TS] Add AbortSignal to fetch template

nulladdict:
This adds support for [AbortSignal](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal) for Fetch template
I've mostly followed [this PR](https://github.com/RicoSuter/NSwag/pull/2846/) for axios

One thing I'm concerned about is that it seems like `AbortSignal` typings were [added](https://github.com/microsoft/TypeScript/issues/19418) in TypeScript 2.7, so I wonder if signal generation should be opt-in through some kind of setting

### DIFF
--- a/src/NSwag.CodeGeneration.TypeScript.Tests/FetchTests.cs
+++ b/src/NSwag.CodeGeneration.TypeScript.Tests/FetchTests.cs
@@ -105,5 +105,52 @@ namespace NSwag.CodeGeneration.TypeScript.Tests
             Assert.DoesNotContain("FormData", code);
             Assert.Contains("\"Content-Type\": \"application/x-www-form-urlencoded\"", code);
         }
+
+        [Fact]
+        public async Task When_abort_signal()
+        {
+            //// Arrange
+            var generator = new WebApiOpenApiDocumentGenerator(new WebApiOpenApiDocumentGeneratorSettings());
+            var document = await generator.GenerateForControllerAsync<UrlEncodedRequestConsumingController>();
+            var json = document.ToJson();
+
+            //// Act
+            var codeGen = new TypeScriptClientGenerator(document, new TypeScriptClientGeneratorSettings
+            {
+                Template = TypeScriptTemplate.Fetch,
+                TypeScriptGeneratorSettings =
+                {
+                    TypeScriptVersion = 2.7m
+                }
+            });
+            var code = codeGen.GenerateFile();
+
+            //// Assert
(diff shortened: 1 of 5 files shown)


nulladdict:
I've added conditional generation, depending on TypeScript version.
I wasn't sure if this should've been a separate setting, or computed from TypeScript version,
but I guess this should be fine, since generated `signal` is optional parameter

RicoSuter:
> I wasn't sure if this should've been a separate setting, or computed from TypeScript version,

Is this a feature which is supported in all browsers/fetch implementations?
Why is this dependent on the TS version?
If it's not supported in older versions/not in all fetch versions than we should probably put this behind a config..

RicoSuter:
It's even experimental: 

image

nulladdict:
The original discussion about aborting fetch ([1](https://github.com/whatwg/fetch/issues/27)) ([2](https://github.com/whatwg/fetch/issues/447)) ended up using `AbortSignal` for [`Request`](https://fetch.spec.whatwg.org/#request-class), but I guess `AbortSignal` & `AbortController` are still considered experimental
Generally speaking, modern browsers that support `fetch` tend to also support this api, but additional polyfills might be [necessary](https://github.com/github/fetch#aborting-requests)

I don't mind hiding this behind a config setting

Also, it's dependent on the TS version, because earlier versions didn't have types for `AbortSignal` & `AbortController`, but I think if we're putting this behind a setting this might not be relevant

RicoSuter:
As its experimental and potentially a breaking change i think its better to put this behind a config (UseAbortSignal, default: false). Is that ok? Can you update the PR?

nulladdict:
Sounds good to me, I'll update the PR

RicoSuter:
@nulladdict can we merge this PR?

nulladdict:
I think so, if there are no problems with liquid template.
I am not very experienced with it, it looks fine for me, but I might be missing something

RicoSuter:
Lgtm, thanks for the PR.

nulladdict:
No problem, should we also update the documentation for new `UseAbortSignal` config?
I'm not sure what (if anything) needs to be done to update [the wiki](https://github.com/RicoSuter/NSwag/wiki/TypeScriptClientGeneratorSettings)

RicoSuter:
image

RicoSuter:
Just add a new group there and document it -that's probably fine

PR #3124, opened 2020-10-23
Fixes an issue where nullable ENUM query parameter had double OneOf -schema

Havunen:
Fixes an issue where nullable ENUM query parameter had double OneOf -schema
Fixes an issue where nullable collection of Enums query parameter was defined as request body in GET method

### DIFF
--- a/src/NSwag.Generation.WebApi.Tests/Nullability/ParameterNullabilityTests.cs
+++ b/src/NSwag.Generation.WebApi.Tests/Nullability/ParameterNullabilityTests.cs
@@ -1,7 +1,9 @@
-﻿using System.Linq;
+﻿using System.Collections.Generic;
+using System.Linq;
 using System.Threading.Tasks;
 using System.Web.Http;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Newtonsoft.Json;
 using NJsonSchema;
 using NJsonSchema.Annotations;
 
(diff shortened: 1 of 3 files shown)


Havunen:
@RicoSuter Can you review this change please :)

Havunen:
@RicoSuter Can you please review this its roadblocker for our OpenAPI 3 project

RicoSuter:
Can you post a part of a spec how it's currently generated vs how it's done in the changes?

Havunen:
>Can you post a part of a spec how it's currently generated vs how it's done in the changes?

I was trying to give example from the test case I added, and I noticed it produces identical code before and after the fix. So the test case is not doing what is was supposed to do. I will look into it. 

Anyway; currently the NSwag generator OpenAPI 3 schema like this for some enum parameters:

```
        "parameters": [
          {
            "name": "integrationName",
            "in": "query",
            "description": "Optional: The integration of which logs you want.",
            "schema": {
              "oneOf": [
                {
                  "nullable": true,
                  "oneOf": [
                    {
                      "$ref": "#/components/schemas/Integration"
                    }
                  ]
                }
              ]
            },
            "x-position": 1
          },
```
It should have only one of those "oneOf" and have nullable property at schema. I will look into the test case and change it accordingly

RicoSuter:
I think the changes in CreatePrimitiveOpenApiParameter will remove the inner oneOf and not the outer one...
I think the fix needs to be done somewhere else (ie avoid the outer oneOf ref completely).

Havunen:
I will debug this tomorrow and create proper test case first to reproduce the issue.

Havunen:
I created new PR for test case here: https://github.com/RicoSuter/NSwag/pull/3124

Issue #3108, opened 2020-10-15
CS8825 warnings when building generated C# client code with the .NET 5 RC2 SDK


Numpsy:
Hi,

I think this is an issue with the new SDK, but i'm not entirely certain if it's that or the latest preview release of Visual Studio 2019, but:

I have a project that uses NSwag to generate a C# OpenApi client with the C#8 nullability options enabled, and compiles the generated source into an ASP.Net Core 3.1 project (targetting .NET Core 3.1).

After I updated my dev system to the latest VS2019 preview release, building the project gives me multiple CS8825 warnings from the ```ConvertToString``` methods in the generated client code:

![image](https://user-images.githubusercontent.com/1178570/96142104-98e93d00-0ef9-11eb-90ab-8e2d76fb6558.png)

I don't know if there would ever really be a null here, but looking at the template code at https://github.com/RicoSuter/NSwag/blob/285e1c09c8f3fa17dd435be469ec2a046a4e0895/src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.ConvertToString.liquid#L29, I wonder if that should be checking for the result of ToString() being null and returning string.Empty if so as the other returns in the function to?

Numpsy:
I've tried my build with the 13.9 release and it builds ok now, so I'll close this.

PR #3125, opened 2020-10-24
Updated ConvertToString to always return a non-nullable string

jeremyVignelles:
Broadens compatibility with older frameworks (like netstandard 2.0) that doesn't support `NotNullIfNotNull`.

I checked all usages of ConvertToString to avoid breaking changes:

https://github.com/RicoSuter/NSwag/blob/3787156acfc9ec6b519289d54fad312d1963da4b/src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.QueryParameter.liquid#L10
was guarded by `(item_ == null) ? "" : ` . I removed the check for simplicity, because it will return an empty string. No functional change here

https://github.com/RicoSuter/NSwag/blob/3787156acfc9ec6b519289d54fad312d1963da4b/src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.QueryParameter.liquid#L15 Already guarded by an if, no change here

https://github.com/RicoSuter/NSwag/blob/3787156acfc9ec6b519289d54fad312d1963da4b/src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.QueryParameter.liquid#L24 Removed the check the same way as above.

https://github.com/RicoSuter/NSwag/blob/3787156acfc9ec6b519289d54fad312d1963da4b/src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.ConvertToString.liquid#L43 Null are filtered out by the line above, no change.

https://github.com/RicoSuter/NSwag/blob/ec30ee0c862ee345e869e8a0da5b82155cdc5294/src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.PathParameter.liquid#L12 Would throw an `ArgumentNullException` if a null was given. Now transforms the null to "". Not sure if that's checked upstream or whether nobody faced the issue. Or maybe they just convert all their null parameters to "" by themself.

https://github.com/RicoSuter/NSwag/blob/cffdb99fd36e39cb7a3441336d9472e32cda40e1/src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.HeaderParameter.liquid#L4 Seems to be OK :  https://github.com/dotnet/runtime/blob/0e402bcd28f38a3b65720474dcd7a338cab0841d/src/libraries/System.Net.Http/src/System/Net/Http/Headers/HttpHeaders.cs#L139

https://github.com/RicoSuter/NSwag/blob/6ce40aa83b5b46f98fb089a0622cd55bf9d777c7/src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.liquid#L216 Not executed if it is null

https://github.com/RicoSuter/NSwag/blob/6ce40aa83b5b46f98fb089a0622cd55bf9d777c7/src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.liquid#L249 If I read it correctly, it cannot be null in that block

### DIFF
--- a/src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.ConvertToString.liquid
+++ b/src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.ConvertToString.liquid
@@ -1,13 +1,12 @@
 ﻿{%              if GenerateNullableReferenceTypes -%}
-[return: System.Diagnostics.CodeAnalysis.NotNullIfNotNull("value")]
-private string? ConvertToString(object? value, System.Globalization.CultureInfo cultureInfo)
+private string ConvertToString(object? value, System.Globalization.CultureInfo cultureInfo)
 {%              else -%}
 private string ConvertToString(object value, System.Globalization.CultureInfo cultureInfo)
 {%              endif -%}
 {
     if (value == null)
     {
-        return null;
+        return "";
     }
 
     if (value is System.Enum)
--- a/src/NSwag.CodeGeneration.CSharp.Tests/ArrayParameterTests.cs
+++ b/src/NSwag.CodeGeneration.CSharp.Tests/ArrayParameterTests.cs
@@ -66,7 +66,7 @@ namespace NSwag.CodeGeneration.CSharp.Tests
             var code = generator.GenerateFile();
 
             //// Assert
-            Assert.Contains(@"foreach (var item_ in elementId) { urlBuilder_.Append(System.Uri.EscapeDataString(""elementId"") + ""="").Append((item_ == null) ? """" : System.Uri.EscapeDataString(ConvertToString(item_, System.Globalization.CultureInfo.InvariantCulture))).Append(""&""); }", code);
+            Assert.Contains(@"foreach (var item_ in elementId) { urlBuilder_.Append(System.Uri.EscapeDataString(""elementId"") + ""="").Append(System.Uri.EscapeDataString(ConvertToString(item_, System.Globalization.CultureInfo.InvariantCulture))).Append(""&""); }", code);
         }
     }
 }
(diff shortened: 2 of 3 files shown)


Numpsy:
Hi,

Would this effect #3108?

(I was wondering about sending a PR that makes it return string.Empty if https://github.com/RicoSuter/NSwag/blob/285e1c09c8f3fa17dd435be469ec2a046a4e0895/src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.ConvertToString.liquid#L29 did produce a null under some circumstances, but haven't had chance to look at it yet)

jeremyVignelles:
Not sure about what the warning you saw means. Maybe you could checkout the branch and see if it fixes something for you?

Numpsy:
I think it's just pointing out that the variant of ```System.Convert.ToString()``` that takes an ```object``` has a nullable return value, and ```ConvertToString``` isn't supposed to return null:

image

(Actually, on a related note that I just noticed - is that line I liked in ConvertToString.liquid supposed to be passing ```cultureInfo``` to  ```ChangeType``` but not to ```ToString``` ?)

jeremyVignelles:
You're right, I probably need to check a few things again...

Numpsy:
fwiw, I've put a small change in #3144 that seems to fix the warnings for me.

RicoSuter:
So we'd first merge #3144 and then this PR?

Numpsy:
If the signature of ```ConvertToString()``` is going to be changed to always return a non-null string, then I'd guess it would make sense to do #3144 first.

src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.ConvertToString.liquid
```diff
@@ -1,13 +1,12 @@
...
-private string? ConvertToString(object? value, System.Globalization.CultureInfo cultureInfo)
+private string ConvertToString(object? value, System.Globalization.CultureInfo cultureInfo)
 {%              else -%}
 private string ConvertToString(object value, System.Globalization.CultureInfo cultureInfo)
 {%              endif -%}
 {
     if (value == null)
     {
-        return null;
+        return "";
```
Numpsy, line 9: should that be string.Empty?

jeremyVignelles: I once was on the "string.Empty" side, but then I heard once that nowadays, the compiler is smart enough to understand that "" and string.Empty are the same.
If there's no perf benefit of using string.Empty, I find "" easier to read.

RicoSuter:
Can we do #3144 directly in this PR to avoid conflicts?
@jeremyVignelles can you ping me when the PR is ready?

RicoSuter:
PR: https://github.com/jeremyVignelles/NSwag/pull/1

src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.ConvertToString.liquid
RicoSuter: Up to you, we can keep it as is or update it.. for me it's fine either way - it's generated code :-)

Issue #3112, opened 2020-10-16
disposeResponse_ and disposeClient_ are always true, why check?


paulovila:
The generated C# client is declaring two local variables ```disposeResponse_``` and ```disposeClient_```
and initialising them to true.
Later on, in a ```finally``` block they are read to check if the ```response_``` and ```client_``` can be disposed respectively.

Between the initialization and the read, both variables are not updated. Or at least it is not very intuitive .

Either the check is unnecessary or there is missing code that updates the aforementioned variables.
See screenshot:
![image](https://user-images.githubusercontent.com/888391/96273075-d667cc00-0fc6-11eb-90d7-1bbec96b0fc0.png)

jeremyVignelles:
Please have a look at the template file and you'll understand how convoluted the cases are.

In your case, the check is not useful, but in some cases, e.g. when a file is downloaded from the API and the stream is returned, you don't want the code to dispose the response and let the caller do that.

It would be unnecessarily complex to precompute whether this is needed in the template.

paulovila:
Hi @jeremyVignelles thanks for your quick response.

It doesn't matter the complexity of the template. That's a result of the complexity of the problem that you guys are solving. From what you are saying I understand that there are different scenarios for disposing each variable. The problem is that this code is critical and needs to be addressed on a case basis.

On a side note, it's  _less usual_ to download files than CRUD operations on a web API, therefore, normal requests shouldn't be penalised at runtime because of the special case/cases.

Are there more cases other than the file download?

jeremyVignelles:
Did you benchmark the cost implied by this code? it's possible that the compiler is smart enough to remove the useless code

paulovila:
It is not responsibility of the compiler make optimisations across different blocks ```try```  and ```finally```.

Even more, in the previous version there was a null check before disposing it, why it could be null?

jeremyVignelles:
If you really care about perf, there are a lot of ways you could improve the template to fit your need. I'd argue that it would be way more performant to not use the code generation and do the http call by hand.

NSwag just tries to work  for many use cases out of the box, and if it introduces a useless variable, I'm fine with that.

Now, if you want to try and make the PR that would solve that without breaking anything nor adding complexity to the code base, please do 🙂

paulovila:
[...]

PR #3136, opened 2020-10-30
C# client class template: simplify disposal of HttpResponseMessage

0xced:
Since the response is always disposed, there is no point in having and testing the `disposeResponse_` variable which is always `true`. The template can thus be simplified with a `using` statement.

### DIFF
--- a/src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.liquid
+++ b/src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.liquid
@@ -267,9 +267,7 @@
                 PrepareRequest(client_, request_, url_);
 
                 {% template Client.Class.BeforeSend %}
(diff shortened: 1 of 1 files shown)


jeremyVignelles (changes requested):
Hi,

Thanks for the PR. Unfortunately, as explained in #3112, things are not that easy.

Did you miss that line from the included ProcessResponse template ? https://github.com/RicoSuter/NSwag/blob/285e1c09c8f3fa17dd435be469ec2a046a4e0895/src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.ProcessResponse.liquid#L6

0xced:
Thanks for the quick reply.

I totally missed that, I thought the `disposeResponse_` was left from a previous version. I should have grepped the whole project first!

jeremyVignelles:
It is in fact pretty new ;) https://github.com/RicoSuter/NSwag/pull/3016/files

It was introduced to fix NullableReferenceType issues.

PR #3144, opened 2020-11-05
Change ConvertToString to handle the possibility of System.Convert.ToString returning null

Numpsy:
refs #3108 - ```System.Convert.ToString()``` is documented as being able to return null, but ```ConvertToString()``` isn't supposed to return null in this case, which provokes a ```CS8825``` warning with some compiler versions.

I'm not sure if this specific case (converting an enum) will return null in any normal case, but in case it does, this change adopts the approach used in the default case at the end of the function, which is to return string.Empty rather than null in this case.

### DIFF
--- a/src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.ConvertToString.liquid
+++ b/src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.ConvertToString.liquid
@@ -26,7 +26,8 @@ private string ConvertToString(object value, System.Globalization.CultureInfo cu
                 }
             }
 
-            return System.Convert.ToString(System.Convert.ChangeType(value, System.Enum.GetUnderlyingType(value.GetType()), cultureInfo));
+            var converted = System.Convert.ToString(System.Convert.ChangeType(value, System.Enum.GetUnderlyingType(value.GetType()), cultureInfo));
+            return converted == null ? string.Empty : converted;
         }
     }
     else if (value is bool) 


src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.ConvertToString.liquid
```diff
@@ -26,7 +26,8 @@ private string ConvertToString(object value, System.Globalization.CultureInfo cu
                 }
             }
 
-            return System.Convert.ToString(System.Convert.ChangeType(value, System.Enum.GetUnderlyingType(value.GetType()), cultureInfo));
+            var converted = System.Convert.ToString(System.Convert.ChangeType(value, System.Enum.GetUnderlyingType(value.GetType()), cultureInfo));
+            return (converted is null) ? string.Empty : converted;
```
RicoSuter, line 30: I think we should use 

    return converted == null ? string.Empty : converted;

to keep compatibility with older C# versions.

Numpsy: I did it in that style because the existing code in the template is like that:
https://github.com/RicoSuter/NSwag/blob/285e1c09c8f3fa17dd435be469ec2a046a4e0895/src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.ConvertToString.liquid#L47

Will change if you want though.

RicoSuter:
@jeremyVignelles do you think this is safe to merge?

jeremyVignelles:
Looks good to me, but I didn't test yet. Agreed with @RicoSuter, it would be a good idea to merge this with #3125 
Maybe @Numpsy you could make a PR on my fork, against my branch? Otherwise, I should be able to cherry-pick the commit.

Hope to have time this week but it's been busy here 🙂

Numpsy:
Ok, I'll try to send a PR to your fork.

RicoSuter:
Closed in favor of #3125

PR #3150, opened 2020-11-10
Added test case to reproduce issue 3149

Havunen:
Added test case to reproduce issue https://github.com/RicoSuter/NSwag/issues/3149


Havunen:
The expected result is probably still not good enough.

https://swagger.io/docs/specification/data-models/enums/

PR #3153, opened 2020-11-10
Added test case and proposed fix to issue 3149 

Havunen:
Added test case and proposed fix to issue https://github.com/RicoSuter/NSwag/issues/3149

This PR includes the test case and removes other of those OneOf wrappers. @RicoSuter

### DIFF
--- a/src/NSwag.Generation.WebApi/Processors/OperationParameterProcessor.cs
+++ b/src/NSwag.Generation.WebApi/Processors/OperationParameterProcessor.cs
@@ -495,14 +495,6 @@ namespace NSwag.Generation.WebApi.Processors
                 {
                     operationParameter.Default = defaultValue;
                 }
-                else if (operationParameter.Schema.HasReference)
-                {
-                    operationParameter.Schema = new JsonSchema
-                    {
-                        Default = defaultValue,
-                        OneOf = { operationParameter.Schema }
-                    };
-                }
                 else
                 {
                     operationParameter.Schema.Default = defaultValue;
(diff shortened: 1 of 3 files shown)


Havunen:
@RicoSuter what do you think about this PR, it now removes the other oneOf wrapper

src/NSwag.Generation.WebApi/Processors/OperationParameterProcessor.cs
```diff
@@ -495,14 +495,6 @@ private OpenApiParameter AddBodyParameter(OperationProcessorContext context, str
                 {
                     operationParameter.Default = defaultValue;
                 }
-                else if (operationParameter.Schema.HasReference)
```
RicoSuter, line 498: I think we need to keep this in case there is actually a default value, but only then: 

    else if (operationParameter.Schema.HasReference && defaultValue != null)

RicoSuter, line 498: I think removing it is wrong, but you can just synchronize it with the non-legacy (ASP.NET Core) implementation: 

image

https://github.com/RicoSuter/NSwag/blob/master/src/NSwag.Generation.AspNetCore/Processors/OperationParameterProcessor.cs#L450

And then set AllowReferencesWithProperties to false

Havunen: Thanks for the input, but how should this be synchronized? Is there some way to recognize if the other place already added schema OneOf wrapper?

Issue #3127, opened 2020-10-26
CS8073 build warnings when building generated C# clients against as .NET5 build target


Numpsy:
(Note: Related to/same code as #3108)

I have a generated C# client which is compiled into an ASP.Net Core 3.1 project, and am currently testing with building it with .NET 5.
I get multiple of these warnings with non-nullable Guid parameters that weren't there previously:

![image](https://user-images.githubusercontent.com/1178570/97163274-51c03f00-1778-11eb-9883-d25835175888.png)

Which seems to be one of the new Net5 warnings mentioned in this [Microsoft blog post](https://devblogs.microsoft.com/dotnet/automatically-find-latent-bugs-in-your-code-with-net-5/#all-new-warnings-and-errors-coming-in-analysis-level-5).

I don't know if there is a simple way to remove the null checks in these cases, or if the warning (sh/c)ould be disabled in the generated code as CS0472 is now @ https://github.com/RicoSuter/NSwag/blob/285e1c09c8f3fa17dd435be469ec2a046a4e0895/src/NSwag.CodeGeneration.CSharp/Templates/File.liquid#L17

?

jeremyVignelles:
We should handle the case in the code generator.

Probably a duplicate of #2855

RicoSuter:
We could just add a ! for simplicity (unnecessary null check doesnt hurt)

jeremyVignelles:
like `jobId == null!` ? There should be a better way I think, because we could test if the type is nullable or not, right?

Numpsy:
Just changing it to ```== null!``` doesn't seem to have an effect, if that's what you mean:

![image](https://user-images.githubusercontent.com/1178570/98264902-52c74c80-1f80-11eb-87c1-a250d23507fb.png)

I don't know enough about how the templates/generation work to really say about checking if the type is nullable in them (I see that ```CSharpParameterModel``` has an ```IsSystemNullable``` property that just works goes on whether the type name ends with ```?```, but I don't think that would be useful for reference types?)

Numpsy:
The supression in the in the generated code in the 13.9 has silenced the warnings I was seeing.

Is it worth leaving this open for any further discussion of fixing the warnings rather than disabling them, or will #2855 do for that?

PR #3156, opened 2020-11-11
Disable CS8073 warnings in the generated C# client files

Numpsy:
refs #3127 

It's not as 'clean' as changing the code generation to remove the null checks for value types alltogether, but it is inkeeping with the existing suppression of CS0472 which silences similar warnings for parameters of types like ```int```, so maybe it's acceptable at least in the short term?

Just a thought really (the currently generated code breaks the build for a project I work on that has WarningsAsErrors enabled if you try to build it againt .NET 5, so it's helpful to get something done with it)


RicoSuter:
Thanks for the PR. If feasible it surely makes sense to fix the warnings but also we need to keep the template complexity as low as possible and we are talking about generated code... so it's always a bit a compromise.

PR #3186, opened 2020-11-20
Interface Annotations

matthawley:
We are in need of annotating the generated interface with our own attributes based on document level extension data to be able to handle more dynamic client creation.

# Additions
1. Adding `ExtensionData` to the client template model.
2. Modifying the client interface template to call a new client interface annotations template.

# Example Usage

Input from `Client.Interface.Annotations.liquid`
```
[Services.Documentation.ShortName("{{ ExtensionData["x-service-short-name"] }}")]
```

Generated Output
```
[Services.Documentation.ShortName("discovery")]
```


RicoSuter:
Maybe we should do the same with Client.Class.Annotations just that its consistent?

matthawley:
@RicoSuter added the class annotations.

RicoSuter:
Thanks for the PR.

PR #3198, opened 2020-11-29
Added multiple file open support & close all documents option

HiTechMagic:
Added close all files support
Adjusted file extension to match wider range of filenames (still compatible, but allowing version prefixes)


RicoSuter:
Thanks for the PR.

PR #3203, opened 2020-12-02
Fix Null Reference when Schema is null in a specific circumstance.

ksmithRenweb:
The GetTypeName function handles a null Schema when it is passed in but the hint parameter is attempting to use the Schema assuming it is not null. This is where the NullReferenceException is generated. With a couple of extra lines of code, that case is handled. 

The other option is to make the GetTypeName hint default with null and inside the method if Schema is not null but hint is null. Then do the code, "hint = !response.Value.Schema.HasTypeNameTitle ? "Response" : null"


src/NSwag.CodeGeneration/Models/OperationModelBase.cs
```diff
@@ -117,7 +117,9 @@ public string UnwrappedResultType
                 }
 
                 var isNullable = response.Value.IsNullable(_settings.CodeGeneratorSettings.SchemaType);
-                return _generator.GetTypeName(response.Value.Schema, isNullable, !response.Value.Schema.HasTypeNameTitle ? "Response" : null);
+                var schemaHasTypeNameTitle = response.Value.Schema?.HasTypeNameTitle;
+                var hint = schemaHasTypeNameTitle.HasValue && !schemaHasTypeNameTitle.Value ? "Response" : null;
```
jeremyVignelles, line 121: why not `hasTypeNameTitle == false`, using the fact that false != null

jeremyVignelles (commented):
Not sure what the behavior should be here, I don't have enough context, but the change seems to improve null safety here

src/NSwag.CodeGeneration/Models/OperationModelBase.cs
ksmithRenweb: The exception is in this code segment: response.Value.Schema.HasTypeNameTitle
The problem is that repsonse.Value.Schema is null.
So Schema.HasTypeNameTitle throws a NullReferenceException because the Schema is null.

inside of _generator.GetTypeName its first code segment checks for a null Schema appropriately. Unfortunately to populate the hint parameter, someone used a Schema object to decide the hint without handling the edge case of Schema being null.

RicoSuter: Yep you can replace `schemaHasTypeNameTitle.HasValue && !schemaHasTypeNameTitle.Value`
with `schemaHasTypeNameTitle == false`

ksmithRenweb: i made that change.

ksmithRenweb: @RicoSuter resolving comment after that change.

RicoSuter:
I'd like to merge but there are conflicts. Can you fix those?

ksmithRenweb:
> I'd like to merge but there are conflicts. Can you fix those?

It looks like the conflicts are from equivalent code that you commited an hour before your comment. See below. 
Do you prefer the more verbose style I have, or the concise style you commited?

<<<<<<< fixNullRef
                var schemaHasTypeNameTitle = response.Value.Schema?.HasTypeNameTitle;
                var hint = schemaHasTypeNameTitle != true ? "Response" : null;
                return _generator.GetTypeName(response.Value.Schema, isNullable, hint);
=======
                return _generator.GetTypeName(response.Value.Schema, isNullable, response.Value.Schema?.HasTypeNameTitle != true ? "Response" : null);
>>>>>>> master

RicoSuter:
I think the more verbose one is better.

ksmithRenweb:
@RicoSuter Merge conflict fixed.

RicoSuter:
Thx.

PR #3211, opened 2020-12-08
Re-throw exceptions preserving their stack traces

### LINKED ISSUE #3210: Exceptions re-thrown on subsequent document accesses get their stack traces erased
Basically doing `throw exceptionThatHasBeenCreatedMuchEarlier;` is a big no-no, but this is what happens here: https://github.com/RicoSuter/NSwag/blob/4e94b9d2bd558eb7bc2fa5f944b6a3ff08d62c68/src/NSwag.AspNetCore/Middlewares/OpenApiDocumentMiddleware.cs#L89

`ExceptionDispatchInfo` should be used instead. I will try to submit a PR for this.

LordJZ:
Fixes #3210


RicoSuter:
Thanks for the PR.

Issue #3208, opened 2020-12-08
DateTime headers uri escaped


prrrtpieptoet:
Recently formatting of DateTime headers was fixed. ConvertToString used to be applied and the only way to apply a custom format was to override ConvertToString in a custom partial client, next to the generated client.

In addition to applying the format specified by configuration setting parameterDateTimeFormat (which is a great fix), the resulting string is now uri escaped. 

My question: why is the string uri escaped? I cannot find any rfc specifying header values should be uri escaped. The api I communicate with doesn't expect that either, resulting in all requests from the client generated with NSwagStudio being rejected by the api.

ScottKaye:
This is my fault - [#3174](https://github.com/RicoSuter/NSwag/pull/3174).  I have submitted another PR #3220 to fix this.  Sorry about this!

In the meantime, you can use the `TemplateDirectory` to specify your own overridden template with the fixed value.

prrrtpieptoet:
No worries Scott, many thanks for the reply and the fix!

PR #3220, opened 2020-12-15
Fix issue #3208 - headers should not be URI encoded

ScottKaye:
My previous pull request [#3174](https://github.com/RicoSuter/NSwag/pull/3174) mistakenly URI-encoded the datetime header.  This issue is pointed out in #3208.

My original pull request worked for me, but #3208 is correct in that the date should not be encoded (no other headers are).


RicoSuter:
Thanks for the PR.

Issue #3247, opened 2021-01-13
CSharp client generator - Add the ability to define PrepareRequest and ProcessResponse methods using the Async Pattern


francozuca:
### Current behaviour
**PrepareRequest** and **ProcessResponse** methods are currently being generated as below:
```csharp
partial void PrepareRequest(System.Net.Http.HttpClient client, System.Net.Http.HttpRequestMessage request, string url);
partial void PrepareRequest(System.Net.Http.HttpClient client, System.Net.Http.HttpRequestMessage request, System.Text.StringBuilder urlBuilder);
partial void ProcessResponse(System.Net.Http.HttpClient client, System.Net.Http.HttpResponseMessage response);
```
### What's the problem with it?
On our current implementation, we are getting the response content using **ReadAsStringAsync**, and we are forced to use **GetAwaiter().GetResult()** instead of **await**, because the method is not **async**:
```csharp
partial void ProcessResponse(System.Net.Http.HttpClient client, System.Net.Http.HttpResponseMessage response)
     //...
     var responseString = response.Content.ReadAsStringAsync().GetAwaiter().GetResult();
     //...     
}
```

### What's the solution we propose? (PR #3247)
We propose to make the following changes:

1. Add a boolean property **GeneratePrepareRequestAndProcessResponseAsAsyncMethods** to **NSwag.CodeGeneration.CSharp.CSharpClientGeneratorSettings** as following:
```csharp
/// <summary>
/// Gets or sets a value indicating whether to create PrepareRequest and ProcessResponse as async methods, or as partial synchronous methods.
/// If value is set to true, PrepareRequestAsync and ProcessResponseAsync methods must be implemented as part of the client base class (if it has one) or as part of the partial client class.
/// If value is set to false, PrepareRequest and ProcessResponse methods will be partial methods, and implement them is optional.
/// </summary>
public bool GeneratePrepareRequestAndProcessResponseAsAsyncMethods { get; set; }
```
2. Add a boolean property **GeneratePrepareRequestAndProcessResponseAsAsyncMethods** to **NSwag.CodeGeneration.CSharp.Models.CSharpClientTemplateModel** as following;
```csharp
/// <summary>
/// Gets or sets a value indicating whether to create PrepareRequest and ProcessResponse as async methods, or as partial synchronous methods.
/// If value is set to true, PrepareRequestAsync and ProcessResponseAsync methods must be implemented as part of the client base class (if it has one) or as part of the partial client class.
[...]

PR #3248, opened 2021-01-13
CSharp client generator - Implementing Async Pattern on PrepareRequest and ProcessResponse 

francozuca:
Issue #3247

### DIFF
--- a/src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.liquid
+++ b/src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.liquid
@@ -265,7 +265,7 @@
                 request_.Headers.Accept.Add(System.Net.Http.Headers.MediaTypeWithQualityHeaderValue.Parse("{{ operation.Produces }}"));
 {%     endif -%}
 
-{% if GeneratePrepareRequestAndProcessResponseAsAsyncMethods == true %}
+{% if GeneratePrepareRequestAndProcessResponseAsAsyncMethods %}
                 await PrepareRequestAsync(client_, request_, urlBuilder_);
 {% else -%}
                 PrepareRequest(client_, request_, urlBuilder_);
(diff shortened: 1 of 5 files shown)


src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.liquid
```diff
@@ -279,7 +292,11 @@
                             headers_[item_.Key] = item_.Value;
                     }
 
+{% if GeneratePrepareRequestAndProcessResponseAsAsyncMethods %}
```
RicoSuter, line 295: Change `%}` to `-%}` in all ifs so that the line is removed in the output...

francozuca: Done!

RicoSuter:
We would also need to expose the new config wia CLI command and in the UI (XAML).. can do that tomorrow or you update the PR for that as well. But looks good to me. Thanks.

francozuca:
> We would also need to expose the new config wia CLI command and in the UI (XAML).. can do that tomorrow or you update the PR for that as well. But looks good to me. Thanks.

I just added **GeneratePrepareRequestAndProcessResponseAsAsyncMethods** to CLI and UI. Please let me know if you want me to rephrase the option description/label. Thanks.

cmd
ui

RicoSuter:
Lgtm, thank you very much.

jclijmans:
There seems to be an error in this PR.
In /src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.liquid when GeneratePrepareRequestAndProcessResponseAsAsyncMethods == true, no partial method definition for PrepareRequestAsync is generated
This results in a generated class that doesn't compile because the PrepareRequestAsync is used later on.

francozuca:
> There seems to be an error in this PR.
> In /src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.liquid when GeneratePrepareRequestAndProcessResponseAsAsyncMethods == true, no partial method definition for PrepareRequestAsync is generated
> This results in a generated class that doesn't compile because the PrepareRequestAsync is used later on.

Hi @jclijmans! 
You're right: if GeneratePrepareRequestAndProcessResponseAsAsyncMethods == true, no partial method definition for PrepareRequestAsync is generated. 

That is because partial methods cannot be marked as async. So what you should do is to create a partial definition of your client class and then implement those methods in that particular. You can take a look to #3247 where I tried to explain this properly.

Thanks!

PR #3250, opened 2021-01-15
Fixing FormData generated for typescript with array property (v13.10.0)

felipeprabor:
The scenario is when the content type is form-data and there is an array property in the schema it was generating something like:
content_.append("arrayOfIds", arrayOfIds.toString());
Which was wrong because in the Api request it was always an array of one item with the string concat in comma.
To fix this I've changed the template to check if the parameter is an array it should generate the typescript contract with the code:
arrayOfIds.forEach(item_ => content_.append("arrayOfIds", item_.toString());


RicoSuter:
Thanks for the PR. Will merge and also release with v13.10.0...

Issue #3258, opened 2021-01-18
TS Generation: Null cant be passed to nullable query param could not be passed


FreeFrags:
The behavior changed between NSwag.MSBuild 13.4.2 and 13.5.0

This is what gets generated for a nullable parameter 
```
getPrivate(folderId: number | null,
        let url_ = this.baseUrl + "/api/FolderContent/GetPrivate?";
        if (folderId === undefined)
            throw new Error("The parameter 'folderId' must be defined.");
       else if(folderId !== null) **<----- this should be an else not else if**
            url_ += "folderId=" + encodeURIComponent("" + folderId) + "&";
```

else if(folderId !== null) **<----- this should be an else not else if**

The error if undefined is correct but the else if should be an else because a nullable param can be null and should then be added to the url as null

FreeFrags:
Ill create a pull request for this

bbarry:
Objectively that file is a rats nest of potential bugs waiting to happen. Maintaining branch states across languages might make for microscopically more efficient code but it comes at the cost of significant increased burden on all future contributors.

It would be much better if the file were something like:

```
let url_ = this.baseUrl + "/{{ operation.Path }}{% if operation.HasQueryParameters %}?{% endif %}";
{% for parameter in operation.PathParameters -%}
  url_ = addPath(
    url_,
    '{{ parameter.VariableName }}',
    '{{ parameter.Name }}',
    {{ parameter.VariableName }},
    {{ parameter.IsRequired }},
    {{ parameter.IsDateOrDateTimeArray}},
    {{ parameter.IsDateOrDateTime }},
    {{ parameter.IsArray }},
    {{ parameter.IsOptional }}
    );
{% endfor -%}
{% for parameter in operation.QueryParameters -%}
  url_ = addQuery(
    url_,
    '{{ parameter.VariableName }}',
    '{{ parameter.Name }}',
    {{ parameter.VariableName }},
    {{ parameter.IsRequired }},
    {{ parameter.IsNullable }},
    {{ parameter.IsDateOrDateTimeArray}},
    {{ parameter.IsDateOrDateTime }},
    {{ parameter.IsArray }},
    {{ parameter.IsOptional }}.
    {{ QueryNullValue}}
    );
{% endfor -%}
url_ = url_.replace(/[?&]$/, "");
```

and the logic of actually adding the parameter was in TS.

vocko:
I second this, just tested with 13.18.2 and it still generates the (invalid) check.

[...]

PR #3259, opened 2021-01-18
#3258 changed an else if to an else for nullable query parameters

FreeFrags:
Unfortunately im not sure how to test this, so i haven't but this should fix the issue.

if the param is undefined an error should definitely be shown but if a null is passed the null should be added to the url (because its a nullable param)


RicoSuter:
The problem is that you cannot send a `null` via query parameter - i.e. how can you distinguish between a `"null"`,  `""` and `null` for a string-query parameter in the backend? I think this is the reason to omit it completely so that for example ASP uses the default (null) when the parameter is not defined, "" when the param is defined as empty string (`&foo=&bar=abc`).

PR #3261, opened 2021-01-19
Inherit from ControllerBase for .NET Core/5

gillesdb:
Inherit from ControllerBase for .NET Core instead of ApiController


RicoSuter:
Is this safe and not breaking for anyone?

gillesdb:
Hi @RicoSuter thanks for your quick reply!
I don't think so because in the current implementation it doesn't work for .NET Core, 'ApiController' doesn't exist in Microsoft.AspNetCore.Mvc. 
image

RicoSuter:
Ok then, you're the only one using it it seems ;-)
Thanks for the PR.

gillesdb:
@RicoSuter Or they use the controllerBaseClass property to avoid this issue. 
If we're the only ones using it it's a shame, cause it works perfectly!  Thanks to the controller generation of NSwag we could build a complete contract-first API development process. 
image

Anyway thanks for all your effort! ;)

PR #3268, opened 2021-01-24
Update Microsoft.AspNetCore.Mvc.* references in NSwag.AspNetCore to 1.0.6

Numpsy:
Hi,

A project that I work on which uses NSwag.AspNetCore (13.9.4) has recently been run through Whitesource, and that has complained that it depends on Microsoft.AspNetCore.Mvc.Core version 1.0.4, which suffers from [CVE-2017-8700](https://github.com/aspnet/announcements/issues/279).

The Microsoft announcement says that CVE was fixed in the 1.0.6 update, so this is a simple go at silencing the warning by updating the dependencies to that version.


RicoSuter:
Lgtm, thx for the PR.

PR #3283, opened 2021-02-03
Adding support to the object parameter for multipart/form-data

felipeprabor:
Adding support to the object parameter for multipart/form-data. 
The object parameter should be added in the content as json stringify.


RicoSuter:
Should we do the same for the C# client?

felipeprabor:
I've added the changes for the C# client. Let me know if it is enough.

RicoSuter:
Lgtm. Thank you very much.

felipeprabor:
Hey, when this PR will be released?

RicoSuter:
@felipeprabor can release today or tomorrow.. see also my small change (commit) attached above, is that ok?

felipeprabor:
Hi, today would be great! thanks

PR #3289, opened 2021-02-05
Use Asp.Net Core 2.1 in .NET Framework NSwag.Console

older:
Asp.Net Core 2.1 is only supported version for .NET Framework 4 as of today.
Asp.Net Core 2.2 is EOL as of December 23, 2019


RicoSuter:
v13.10.2

PR #3329, opened 2021-02-24
Use latest version of DiagnosticSource in NSwag.Console for .NET 4.

### LINKED ISSUE #3292: Unable to resolve service for type 'System.Diagnostics.DiagnosticSource' on Asp.Net Core 2.1 on .Net 4
Starting with 13.10.2 NSwag.MSBuild works now with freshly generated from template Asp.Net Core 2.1 project targeting .Net 4. But as soon as reference to Microsoft.ApplicationInsights.AspNetCore is added it starts producing the following exception:

    NSwag command line tool for .NET 4.6.1+ WinX64, toolchain v13.10.2.0 (NJsonSchema v10.3.4.0 (Newtonsoft.Json v11.0.0.0))
    NSwag bin directory: C:\Users\me\.nuget\packages\nswag.msbuild\13.10.2\tools\Win
    System.InvalidOperationException: Unable to resolve service for type 'System.Diagnostics.DiagnosticSource' while attempting to activate 'Microsoft.AspNetCore.Mvc.Razor.RazorPageActivator'.

Repro solution is [attached](https://github.com/RicoSuter/NSwag/files/5945746/AspNetCore21OnNet4.zip).
`System.Diagnostics.DiagnosticSource.dll` in NSwag.exe directory has `FileVersion: 4.6.26919.02` while the one in project's output has   `FileVersion: 5.0.20.51904` so it looks like reference to System.Diagnostics.DiagnosticSource needs to be updated, but I don't see any direct references, so it must be transitive reference, but from which project?

older:
This makes it possible to use it with projects which reference latest version of DiagnosticSource.
Fixes #3292

### DIFF
--- a/src/NSwag.Console.x86/NSwag.Console.x86.csproj
+++ b/src/NSwag.Console.x86/NSwag.Console.x86.csproj
@@ -39,6 +39,7 @@
     <PackageReference Include="NConsole" Version="3.9.6519.30868" />
     <PackageReference Include="Newtonsoft.Json" Version="11.0.2" />
     <PackageReference Include="System.Net.Http" Version="4.3.2" />
+    <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="5.0.1" />
     <PackageReference Include="Microsoft.AspNetCore.TestHost" Version="2.1.1" />
     <PackageReference Include="Microsoft.AspNetCore.Mvc.ApiExplorer" Version="2.1.3" />
     <PackageReference Include="Microsoft.AspNetCore.Mvc.Formatters.Json" Version="2.1.18" />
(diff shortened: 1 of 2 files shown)


RicoSuter:
I'm a bit worried that this is a breaking change for some.

older:
@RicoSuter NSwag.Console already had transitive reference to `System.Diagnostics.DiagnosticSource`, so all this does is makes it explicit in order to bump up the referenced version. In what scenario do you think it might be a breaking change?

RicoSuter:
I think it's fine. Thanks for the PR.

> In what scenario do you think it might be a breaking change?

Usually i'm quite hesitant to upgrade libs in own libs - as this might force consumers to upgrade... but here it's the host and probably fine.

PR #3331, opened 2021-02-25
fix C# Client nullable body guard

RytisLT:
Due to changes introduced in commit e5d6903edd0b82044dccbdf84d86adeb4ee53829 generatic C# Client from swagger.json below will add `null` guard for optional body parameter. Due to recent changes to `IsRequired` and `IsNullable` evaluation from commit b35310290ed944dc7af40088564197e2f2efbf4a it makes sense to revert change from e5d6903edd0b82044dccbdf84d86adeb4ee53829 and use `IsRequired` instead of `IsNullable` to evaluate whether a guard is needed

```
{
  "openapi": "3.0.1",
  "paths": {
    "testOp": {
      "get": {
        "operationId": "testOp",
        "requestBody": {
          "content": {
            "*/*": {
              "schema": {
                "type": "integer",
                "format": "int64"
              }
            }
          }
        },
        "responses": {
          "200": {
            "description": "Success"
          }
        }
      }
    }
  }
}
```

### DIFF
--- a/src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.liquid
+++ b/src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.liquid
@@ -127,7 +127,7 @@
 
 {%         endif -%}
 {%     endfor -%}
-{%     if operation.HasContent and operation.ContentParameter.IsNullable != true -%}
+{%     if operation.HasContent and operation.ContentParameter.IsRequired -%}
         if ({{ operation.ContentParameter.VariableName }} == null)
             throw new System.ArgumentNullException("{{ operation.ContentParameter.VariableName }}");
 
--- a/src/NSwag.CodeGeneration.CSharp.Tests/AllowNullableBodyParametersTests.cs
+++ b/src/NSwag.CodeGeneration.CSharp.Tests/AllowNullableBodyParametersTests.cs
@@ -1,13 +1,60 @@
+using Microsoft.AspNetCore.Mvc;
+using NSwag.CodeGeneration.OperationNameGenerators;
+using NSwag.Generation.WebApi;
 using System.ComponentModel.DataAnnotations;
 using System.Threading.Tasks;
-using Microsoft.AspNetCore.Mvc;
-using NSwag.Generation.WebApi;
 using Xunit;
 
 namespace NSwag.CodeGeneration.CSharp.Tests
 {
     public class AllowNullableBodyParametersTests
     {
+        [Fact]
+        public async Task TestNoGuardForOptionalBodyParameter()
+        {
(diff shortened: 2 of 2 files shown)


src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.liquid
```diff
@@ -127,7 +127,7 @@
 
 {%         endif -%}
 {%     endfor -%}
-{%     if operation.HasContent and operation.ContentParameter.IsNullable != true -%}
+{%     if operation.HasContent and operation.ContentParameter.IsRequired -%}
```
RicoSuter, line 130: I'm not sure about that, mainly because "required" means that "null" is allowed - ie required says that a value is required but null is also a valid value.

RytisLT: Incorrect, the spec says it's nullable int64 aka `long?`. The produced Client method has the correct signature:

`public async System.Threading.Tasks.Task TestOpAsync(long? body, System.Threading.CancellationToken cancellationToken)`

but the not null guard is added which doesn't make sense for a nullable parameter.

```
if (body == null)
    throw new System.ArgumentNullException("body");
```

If body IS required:

```
{
  "openapi": "3.0.1",
  "paths": {
    "/testOp": {
      "get": {
        "operationId": "testOp",
        "requestBody": {
          "content": {
            "*/*": {
              "schema": {
                "type": "integer",
                "format": "int64"
              }
            }
          },
          "required": true
        },
        "responses": {
          "200": {
            "description": "Success"
          }
        }
      }
    }
  }
}
```

then the generated signature is also correct

`public async System.Threading.Tasks.Task TestOpAsync(long body, System.Threading.CancellationToken cancellationToken)`

the null guard still is added which is not needed for non-nullable parameters but it doesn't break the logic. My pull request fixes the first case. Removes null guard for the optional body parameter.

RytisLT: @RicoSuter bump

RicoSuter: Same PR: https://github.com/RicoSuter/NSwag/pull/3358
Ref: https://github.com/RicoSuter/NSwag/pull/3195/files

RicoSuter: Seems that some months ago it was like that...

RytisLT: thanks!

PR #3338, opened 2021-03-02
Fix JsonSerializerParameterCode property for System.Text.Json

unchase:


unchase:
@RicoSuter ready for review

RicoSuter:
Lgtm. Thanks very much.
How well are we with the STJ feature? Is a lot missing?

Issue #2168, opened 2019-05-13
using NSwag inside docker build fails


bleissem:
Hello,
I just try to generate a typescript file based on NSwag
as you can see here:
https://github.com/bleissem/SampleNSwagBuildsInsideDocker
I added a afterbuild target to the WebApplication1.csproj as it can be seen on your wiki
https://github.com/RicoSuter/NSwag/wiki/MSBuild
NSwag is included as nuget packages
    <PackageReference Include="NSwag.AspNetCore" Version="12.2.5" />
    <PackageReference Include="NSwag.MSBuild" Version="12.2.5">
When I build the WebApplication1.csproj locally NSwag code generation works perfect. 
But when I use Docker to generate Typescript the build fails. 
The build fails as you can see here
https://bleissem.visualstudio.com/opensource/_build/results?buildId=3703
with the error message:
Project outputs could not be located in '/src/WebApplication1/bin/Release/netcoreapp2.1/'. Ensure that the project has been built.

Also I add a azure-pipelines.yml so you should be able to reproduce the errormessage yourself

RicoSuter:
> When I build the WebApplication1.csproj locally NSwag code generation works perfect.

Does it also work locally when setting configuration to release? 

`dotnet build --configuration Release`

bleissem:
Hi @RicoSuter ,
in the Dockerfile at line 11 
https://github.com/bleissem/SampleNSwagBuildsInsideDocker/blob/master/Dockerfile#L11
the configuration being used to run dotnet build is Release.
So I should have said that when I build the WebApplication1.csproj locally with Release configuration, NSwag code generation works perfect.

RicoSuter:
But is running a release build outside of docker working?

bleissem:
Hi @RicoSuter ,
building with release configuration works without using docker
Just added another build to demonstrate:
https://bleissem.visualstudio.com/opensource/_build/results?buildId=3819

bleissem:
Hi @RicoSuter, do you have any further questions ? Regards Alex B.

RicoSuter:
Seems like the used paths are wrong

RicoSuter:
Strange, must be something in the project loader... (NSwag.Commands and the entry point alp)

petermein:
Another interresting finding when running outside of the docker container on Linux. 
When running 
`dotnet build "GXP.Api.csproj" -c Release` 

The files path becomes 
[...]

PR #3357, opened 2021-03-11
Add option, OutputPath, to allow seting the location where the project output is located

miasvanklei:
NSwag currently uses the default OutputPath from MSBuild. For normal builds this is sufficient, however if a project is build with a different outputPath (dotnet build -o or msbuild .. /p:OutputPath), it breaks with the following error: _'Project outputs could not be located in ..'_.

This PR fixes the problem, by adding an extra option called OutputPath, which contains the location where the project output is located.

Fixes: at least #2168

### DIFF
--- a/src/NSwag.Commands/Commands/Generation/AspNetCore/AspNetCoreToOpenApiCommand.cs
+++ b/src/NSwag.Commands/Commands/Generation/AspNetCore/AspNetCoreToOpenApiCommand.cs
@@ -61,8 +61,8 @@ namespace NSwag.Commands.Generation.AspNetCore
         [Argument(Name = nameof(NoBuild), IsRequired = false, Description = "Don't build the project. Only use this when the build is up-to-date.")]
         public bool NoBuild { get; set; }
 
-        [Argument(Name = nameof(OutputPath), IsRequired = false, Description = "Path to the MSBUILD output directory")]
-        public string OutputPath { get; set; }
+        [Argument(Name = nameof(MSBuildOutputPath), IsRequired = false, Description = "The MSBuild output path")]
+        public string MSBuildOutputPath { get; set; }
 
         [Argument(Name = nameof(Verbose), IsRequired = false, Description = "Print verbose output.")]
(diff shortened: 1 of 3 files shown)


src/NSwag.Commands/Commands/Generation/AspNetCore/AspNetCoreToOpenApiCommand.cs
```diff
@@ -61,6 +61,9 @@ public class AspNetCoreToSwaggerCommand : OpenApiGeneratorCommandBase<AspNetCore
         [Argument(Name = nameof(NoBuild), IsRequired = false, Description = "Don't build the project. Only use this when the build is up-to-date.")]
         public bool NoBuild { get; set; }
 
+        [Argument(Name = nameof(OutputPath), IsRequired = false, Description = "Path to the MSBUILD output directory")]
+        public string OutputPath { get; set; }
```
RicoSuter, line 65: Not sure whether "OutputPath" is a good name here as I'd expect the OpenAPI.json output path with that... 
Maybe "BuildOutputPath" (msbuild is legacy :-)?

miasvanklei: OutputPath is indeed a bit confusing. I have renamed it to MSBuildOutputPath

RicoSuter:
Can you also quickly add the text field to the UI?

https://github.com/RicoSuter/NSwag/blob/master/src/NSwagStudio/Views/SwaggerGenerators/AspNetCoreToSwaggerGeneratorView.xaml#L54

miasvanklei:
> Can you also quickly add the text field to the UI?
> 
> https://github.com/RicoSuter/NSwag/blob/master/src/NSwagStudio/Views/SwaggerGenerators/AspNetCoreToSwaggerGeneratorView.xaml#L54
No problem

RicoSuter:
Thanks for the PR. Looks good now, can we merge?

miasvanklei:
As far as I'm concerned, yes.

PR #3366, opened 2021-03-17
Drop deprecated Microsoft.AspNetCore.*

trejjam:
Hi, this PR add new targets `netcoreapp3.1;net5.0` in `NSwag.AspNetCore` so it can remove dependency on obsolete packages (`Microsoft.AspNetCore.Mvc.Core`, `Microsoft.AspNetCore.Mvc.Formatters.Json`, `Microsoft.AspNetCore.StaticFiles`)

### DIFF
--- a/src/NSwag.AspNetCore/NSwag.AspNetCore.csproj
+++ b/src/NSwag.AspNetCore/NSwag.AspNetCore.csproj
@@ -68,10 +68,6 @@
     <ProjectReference Include="..\NSwag.Generation\NSwag.Generation.csproj" />
   </ItemGroup>
 
-  <ItemGroup>
-    <Folder Include="obj" />
-  </ItemGroup>
-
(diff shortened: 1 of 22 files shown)


src/NSwag.AspNetCore/NSwag.AspNetCore.csproj
```diff
@@ -63,6 +68,10 @@
     <ProjectReference Include="..\NSwag.Generation\NSwag.Generation.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
```
RicoSuter, line 71: This probably needs to be removed, no?

trejjam: You are right, removed.

Sorry for the delay, I was out of the keyboard.

PR #3371, opened 2021-03-19
Add support for free form objects

### LINKED ISSUE #2257: Support for free form / additional properties in query string.
Hi,

I would like to have a controller with an action such as:

```
[HttpGet("Foo/{id}")]
public async Task<Foo> GetFoo(string id, [FromQuery] Dictionary<string, IEnumerable<string>> additionalProperties)
{
}
```

So that I can invoke it with a URL such as `foo/1?x=y&x=z`. ASP.NET Core supports this and populates the `additionalProperties` as expected with multiple values for the key `x`. 

However, when we invoke the endpoint from the swagger page, say with a value like

```
{
  x: [y, z]
}
```

The URL that is generated and invoked is `foo/1?x=y,z`, but the desired URL is `foo/1?x=y&x=z` as mentioned above. Which when bound in the controller leads to the `additionalProperties` dictionary containing `{ "x", ["y,z"] }` rather than the desired `{ "x", ["y", "z"] }`.

The schema for the `additionalProperaties` in the swagger file that is generated by NSwag looks like:

```
{
  "name": "additionalProperties",
  "in": "query",
  "schema": {
    "type": "object",
    "nullable": true,
    "additionalProperties": {
      "type": "array",
      "items": {
        "type": "string"
      }
    }
  },
  "x-position": 6
}
```

And the swagger input for that param looks like:

image

Both of which seem to be OK to me, but it is just the way that the value for each additional property is being serialised that seems incorrect.

Is there a way to change this behaviour so that it treats each additional property as if it had `CollectionFormat = OpenApiParameterCollectionFormat.Multi`, as it would do if we had a non-free form parameter of type `IEnumerable<string>`, or is this not supported by NSwag? We are also generating a C# client and would like the same behaviour there too.

NSwag.AspNetCore version is 13.0.1 and we have enabled OpenApi3 schemas.

Falco20019:
Fixes #2257

`AdditionalProperties` were supported for the Body but missing for `QueryParameter`.

It seems, that the following `schema` definition will result in `IDictionary<string, string>` until `properties` is explicitly defined. If no `type` is defined (`additionalProperties: true`) or if `properties` are defined, a separate class `ExtendedProperties` is generated.
```json
{
  "ExtendedProperties": {
    "type": "object",
    "additionalProperties": {
      "type": "string"
    }
  }
}
```

Therefore I needed to handle `parameter.IsDictionary` and `parameter.HasAdditionalProperties` seperately. It would be easier if the class is always generated when `parameter.Schema.AllowAdditionalProperties` is `true`. But I'm not too deep into `NJsonSchema` to be able to do this.

### DIFF
--- a/src/NSwag.CodeGeneration.CSharp.Tests/QueryParameterTests.cs
+++ b/src/NSwag.CodeGeneration.CSharp.Tests/QueryParameterTests.cs
@@ -78,5 +78,225 @@ namespace NSwag.CodeGeneration.CSharp.Tests
                 "urlBuilder_.Append(System.Uri.EscapeDataString(\"limit\") + \"=\").Append(System.Uri.EscapeDataString(ConvertToString(paging.Limit, System.Globalization.CultureInfo.InvariantCulture))).Append(\"&\");",
                 code);
         }
+
+        [Fact]
+        public void When_query_parameter_is_untyped_free_form_object_parameters_are_expanded()
+        {
+            var spec = @"{
+  ""openapi"": ""3.0.0"",
+  ""info"": {
+    ""version"": ""1.0.0"",
+    ""title"": ""Query params tests""
+  },
+  ""servers"": [
+    {
+      ""url"": ""http://localhost:8080""
+    }
+  ],
+  ""paths"": {
+    ""/settings"": {
+      ""get"": {
+        ""summary"": ""List all settings"",
+        ""operationId"": ""listSettings"",
+        ""parameters"": [
+          {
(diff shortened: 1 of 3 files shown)


AndriiKysylytsia (approved):
LGTM

We also faced with this issue, so it would be great to megre it ASAP, thanks.

ivandobrynskyi (approved):
Confirm, need to merge for Our project  too.

jramut (approved):

PR #3373, opened 2021-03-22
Removed fall-back as it is hidning errors happening in the program class

Jens-H-Eriksen:
The problem is decribed here: 
https://stackoverflow.com/questions/66710156/why-do-nswag-fail-on-my-computer-and-not-on-my-colleagues-for-the-same-c-sharp-p


Jens-H-Eriksen:
build error

PR #3375, opened 2021-03-23
Removed fall-back as it is hidning errors happening in the program class

Jens-H-Eriksen:
You can find more info about the reason for this PR here: 
https://stackoverflow.com/questions/66710156/why-do-nswag-fail-on-my-computer-and-not-on-my-colleagues-for-the-same-c-sharp-p/66747604#66747604


RicoSuter:
Doesnt this turn of needed fallbacks for certain scenarios/users?
Can you fix the conflict?

RicoSuter:
assembly loading not supported anymore

PR #3389, opened 2021-03-31
Features/allow ref in path item

jellelicht:
Any reason #3086 was merged into a feature branch, instead of master? Either way, a zero-creative-input-from-me PR, from me, to you :smile:


jellelicht:
Woops, it got merged already. It still doesn't work for me if the the files are in a different directory though :/

PR #3402, opened 2021-04-09
Release v13.10.9

RicoSuter:


Numpsy:
Do the new TFMs added to ```NSwag.Generation.AspNetCore``` need to import the ```Microsoft.AspNetCore.Mvc.NewtonsoftJson``` nuget package when the older ones (still) don't?

PR #3404, opened 2021-04-10
Support skipping output generation based on API specification checksum

lahma:
Opt-in feature to allow checking for source API specification checksum. If output file exists and both source checksum and toolchain version matches in the file, generation is skipped altogether. Requires `nswag.json` to have `"checksumCacheEnabled": true` for TS/CS generation configuration for checks to kick in.

If user has created custom templates that can change on disk this feature isn't for them, but if only built-in ones are used the cache should be safe.


RicoSuter:
Thanks for the PR.

PR #3405, opened 2021-04-11
remove invalid C# identifier characters from operation and class names, fixes #11

### LINKED ISSUE #11: Option to choose action methods
Provide option to choose the action methods to generate...

jamesmanningrapidscale:
likely needs work, including:

- need to add tests for validating class names, only validates operation names currently
- probably need to move the tests to a more appropriate place
- not sure if ClientGeneratorBase.cs is really the right place for this cleanup

### DIFF
--- a/src/NSwag.CodeGeneration.Tests/CodeGenerationIdentifierTests.cs
+++ b/src/NSwag.CodeGeneration.Tests/CodeGenerationIdentifierTests.cs
@@ -1,0 +1,155 @@
+﻿using System;
+using System.ComponentModel.DataAnnotations;
+using System.Text.RegularExpressions;
+using System.Threading.Tasks;
+using NJsonSchema;
+using NJsonSchema.CodeGeneration.TypeScript;
+using NJsonSchema.Generation;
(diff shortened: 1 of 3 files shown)


RicoSuter:
Is this PR ready do be reviewed/merged?

jamesmanningrapidscale:
@RicoSuter at least to get feedback on whether its approach is acceptable, yes please.

RicoSuter:
So this probably only applies to C# but you added it in NSwag.CodeGeneration which is for any output language...
If this applies for TS and C# we should rename CSharpIdentifierValidator to something more generic otherwise move up into the NSwag.CodeGeneration.CSharp lib

mrange:
Hi. We have a very similar issue and is blocked. With this PR applied it seems to resolve the issues for us. Naturally from our point of view it would be great if it could be merged so we don't have to rely on homebrew version.

I also did a quick review of the PR and to me the implementation looks clean but I don't know NSwag in detail so my opinion might in practice mean little.

If it could help I could provide the rebased version that I used in my local testing

Issue #3384, opened 2021-03-29
Change generators minimal supported versions to .NET 4.6.1 and .NET Standard 2.0


lahma:
## Background

* .NET 4.6.1 is the oldest supported full framework version, this has API parity with .NET Standard 2.0 even though has binding redirect problems why separate target for full framework is needed
* Microsoft has stated that we should forget < .NET Standard 2.0 and use 2.0 or later
* .NET Core < 3.1, except for 2.1 is no longer supported
* If there still are people targeted against earlier versions I would argue they can probably live with the latest version that still supported said old versions

## Benefits

* Core gets cleaner, less ifdefs
* Performance improves, things like `DynamicApis.XPathEvaluate` can be changed to use direct invocation which is quite a lot faster
    * Also other such DynamicApis implementations can be removed/replaced with invoking the actual API

## Actions to take

* Remove all < NET Standard 2.0 targets from projects, make sure netstandard2.0 is present for libraries
* Remove all < NET 4.6.1 targets from projects, replace with net461 where lower was supported
* Remove supported from exe projects for anything lower than netcoreapp3.1or is netcoreapp2.1

Will you accept a PR which changes these targets? This would make optimization story a lot easier too.

lahma:
If this is accepted I would then apply the same logic to dependencies (NJsonSchema, Namotion.Reflection).

lahma:
@RicoSuter can you please share your thoughts about this, are there pressing reasons to retain the old frameworks and targets? Say if v14 would drop support for older and if there are depending projects they could continue using v13?

Numpsy:
For what it's worth:

The projects i'm using NSwag in that are in active development at this point are .NETCore3.1 and up, but there are some older things using 2.1, and I don't think 

> .NET Core < 3.1 is no longer supported

is actually true, because 2.1 is still under LTS support until August?

2.2 and 3.0 are out of support though, and 2.2 has been a source of some security warnings from tools like Whitesource (because MS aren't fixing CVEs in it as they still are for 2.1, so there are no fixed Nuget packages about to upgrade to)

lahma:
Thanks @Numpsy , I stand corrected, 2.1 is indeed supported until August 21, 2021 . I've updated the description to make exception for 2.1 and as it's netstandard2.0 should be fine too.

RicoSuter:
[...]

PR #3417, opened 2021-04-15
Remove obsolete .NET Core targets and sample projects

lahma:
While working with code base it's a bit tedious when there are a lot of obsolete target configurations. This PR removes support for .NET Core targets that are no longer supported by Microsoft (netcore2.0, netcore2.2, netcore3.0), Basically keeps netcoreapp2.1, netcoreapp3.1 and net50 (+all netstandard and full framework versions). 

Contributes to #3384


RicoSuter:
Thanks for the PR.

PR #3437, opened 2021-04-26
Improve build and reduce warnings

lahma:
This should greatly reduce build warnings and also adds common `Directory.Build.props`. I see that version is updated to every csproj file, you just use Directory.Build.props in the future to get common configuration. Each csproj can always override the base configuration.

* add explicit JSON.NET reference to projects which have trouble resolving a version
* fix warnings (await usage, arrange/act/assert comments being XML doc type, etc)
* use common Directory.Build.props to remove repetition
  * remove duplicated values from individual csproj files
  * fix NuGet package icon to use new embedded version in projects


lahma:
Packing with new shared configuration looks good, still figuring out the errors.

image

RicoSuter:
Oh sh*t, many conflicts... how hard would it be to fix those?

lahma:
I'll look into it in the near future when I have time, all good.

lahma:
@RicoSuter now a green build and at least NSwagStudio installs and starts from the installer. Warnings about netcoreapp2.1 still present, but I would tackle them separately, maybe even by upgrading to .NET 6 all around, earlier versions won't be supported that long anymore (and they should still work quite OK while libs target netstandard).

lahma:
This was done on earlier PR work.

PR #3439, opened 2021-04-26
Generate async PrepareRequest and ProcessResponse methods async setting

### LINKED ISSUE #3428: GeneratePrepareRequestAndProcessResponseAsAsyncMethods generates invalid client code
Generating CSharp client code with generatePrepareRequestAndProcessResponseAsAsyncMethods true results in invalid code.

The generated client code does not have a definition for the partial methods PrepareRequestAsync and ProcessResponseAsync

This error was introduced by PR [#3248 ](https://github.com/RicoSuter/NSwag/pull/3248) that implements the async pattern

It introduced an error in  src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.liquid where there is no else for the statement
{% if GeneratePrepareRequestAndProcessResponseAsAsyncMethods == false -%}
on line 87

jclijmans:
- Generate async methods of PrepareRequest and ProcessResponse methods when GeneratePrepareRequestAndProcessResponseAsAsyncMethods setting is true
- also add cancellationToken parameter for these async method calls

This should fix #3428

### DIFF
--- a/src/NSwag.CodeGeneration.CSharp.Tests/CSharpClientSettingsTests.cs
+++ b/src/NSwag.CodeGeneration.CSharp.Tests/CSharpClientSettingsTests.cs
@@ -150,5 +150,55 @@ namespace NSwag.CodeGeneration.CSharp.Tests
             //// Assert
             Assert.Contains("public partial interface IFooClient : IClientBase", code);
         }
+
+        [Fact]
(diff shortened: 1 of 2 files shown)


RicoSuter:
Ah, no. This PR will not work as a partial method cannot have a return parameter. 
So the current code is as excepted (correct) - i.e you need to define a base class and implement the async method there...

jclijmans:
Oh, ok I see. I didn't know that.
I'll close the issue and the PR

PR #3444, opened 2021-04-30
Typescript named parameters for method invocation

danielwinkler:
Status: proposal

In order to achieve positional independence of the method parameters, this proposal adds a configuration flag that changes the current behavior

```typescript
    test(a: number, b: number | null | undefined): Promise<string | null> {
```

where `const a = 2; const b = 3; test(b,a);` is compiling just fine while not doing what would be expected.
This may in reality become a problem if the OAS changes, e.g. due to some refactoring of the backend code.

By changing to a structure like
```typescript
    test(parameters: {
            a: number; b: number | null | undefined
        }): Promise<string | null> {
        const { a, b } = parameters;
```
the above code will have to change to `test({a,b})`, but is now invariant to changes in the position.

If this is something that you would accept for the code generation, I would love to continue working on this approach.


danielwinkler:
@RicoSuter could you please give a short feedback if you'd accept the proposal?

BR Daniel

danielwinkler:
ping (@RicoSuter )

PR #3449, opened 2021-05-04
mitigate cve-2021-26701

jjoekoullas:
https://msrc.microsoft.com/update-guide/en-US/vulnerability/CVE-2021-26701

just bumping version for system.text.encodings.web to mitigate the above cve.

### DIFF
--- a/src/NSwag.AspNetCore.Launcher/NSwag.AspNetCore.Launcher.csproj
+++ b/src/NSwag.AspNetCore.Launcher/NSwag.AspNetCore.Launcher.csproj
@@ -18,5 +18,6 @@
 
   <ItemGroup Condition="'$(TargetFramework)'=='netcoreapp3.1'">
     <PackageReference Include="System.Runtime.Loader" Version="4.0.0" />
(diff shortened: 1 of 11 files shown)


wyatt-troia:
@jjoekoullas Thanks for opening this. I just encountered the same issue. Do you know if there's a way to fix this locally, while we wait for this PR to be merged?

jjoekoullas:
I ran into with nswag.msbuild.  Not really a good way around it for that - except maybe using a private nuget repo and publishing to it using this branch.

If it's just because you're referencing nswag in your project - just referencing an updated System.Text.Encodings.Web will get you past this.  That's actually what this fix is doing, just referencing an updating version since the problematic reference is in a dependency.

wyatt-troia:
@jjoekoullas Unfortunately I also ran into this with `nswag.msbuild` :(

I may try the private nuget repo approach

wyatt-troia:
Thank you @RicoSuter !

RicoSuter:
Thanks for the PR.

bradygaster:
Thank you @RicoSuter  - was just about to email you about this. @wyatt-troia - does this solve?

PR #3464, opened 2021-05-11
Add x-originalName handling in spec and code generator

RicoSuter:
If the C# controller parameter name does not match the HTTP parameter name then generate a custom name so that the client gens can generate the same parameter names.

### DIFF
--- a/src/NSwag.CodeGeneration/DefaultParameterNameGenerator.cs
+++ b/src/NSwag.CodeGeneration/DefaultParameterNameGenerator.cs
@@ -21,19 +21,22 @@ namespace NSwag.CodeGeneration
         /// <returns>The parameter name.</returns>
         public string Generate(OpenApiParameter parameter, IEnumerable<OpenApiParameter> allParameters)
         {
-            if (string.IsNullOrEmpty(parameter.Name))
+            var name = !string.IsNullOrEmpty(parameter.CustomName) ? 
+                parameter.CustomName : parameter.Name;
+
+            if (string.IsNullOrEmpty(name))
             {
                 return "unnamed";
             }
 
-            var variableName = ConversionUtilities.ConvertToLowerCamelCase(parameter.Name
+            var variableName = ConversionUtilities.ConvertToLowerCamelCase(name
                 .Replace("-", "_")
                 .Replace(".", "_")
                 .Replace("$", string.Empty)
                 .Replace("[", string.Empty)
                 .Replace("]", string.Empty), true);
 
-            if (allParameters.Count(p => p.Name == parameter.Name) > 1)
+            if (allParameters.Count(p => p.Name == name) > 1)
             {
                 return variableName + parameter.Kind;
             }
--- a/src/NSwag.CodeGeneration/DefaultParameterNameGenerator.cs
+++ b/src/NSwag.CodeGeneration/DefaultParameterNameGenerator.cs
@@ -21,8 +21,8 @@ namespace NSwag.CodeGeneration
         /// <returns>The parameter name.</returns>
         public string Generate(OpenApiParameter parameter, IEnumerable<OpenApiParameter> allParameters)
         {
-            var name = !string.IsNullOrEmpty(parameter.CustomName) ? 
-                parameter.CustomName : parameter.Name;
+            var name = !string.IsNullOrEmpty(parameter.OriginalName) ? 
+                parameter.OriginalName : parameter.Name;
 
             if (string.IsNullOrEmpty(name))
             {
--- a/src/NSwag.Core/OpenApiParameter.cs
+++ b/src/NSwag.Core/OpenApiParameter.cs
@@ -44,9 +44,9 @@ namespace NSwag
             }
         }
 
-        /// <summary>Gets or sets a custom name which is often used in code generation (default: null).</summary>
-        [JsonProperty(PropertyName = "x-name", DefaultValueHandling = DefaultValueHandling.IgnoreAndPopulate)]
-        public string CustomName { get; set; }
+        /// <summary>Gets or sets a original name property x-originalName which is often used in code generation (default: null).</summary>
+        [JsonProperty(PropertyName = "x-originalName", DefaultValueHandling = DefaultValueHandling.IgnoreAndPopulate)]
+        public string OriginalName { get; set; }
 
         /// <summary>Gets or sets the kind of the parameter.</summary>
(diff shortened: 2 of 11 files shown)


src/NSwag.Core/OpenApiParameter.cs
```diff
@@ -44,6 +44,10 @@ public string Name
             }
         }
 
+        /// <summary>Gets or sets a custom name which is often used in code generation (default: null).</summary>
+        [JsonProperty(PropertyName = "x-name", DefaultValueHandling = DefaultValueHandling.IgnoreAndPopulate)]
```
RicoSuter, line 48: Better use x-parameterName to avoid collisions with other tools.

jeremyVignelles: `x-parameter-name` or `x-original-parameter` would be better names IMHO

src/NSwag.Generation.AspNetCore/Processors/OperationParameterProcessor.cs
```diff
@@ -184,8 +184,8 @@ public bool Process(OperationProcessorContext operationProcessorContext)
 
                 if (operationParameter != null)
                 {
-                    operationParameter.Position = position;
                     position++;
+                    operationParameter.Position = position;
```
jeremyVignelles, line 188: What's the rationale for this change?

src/NSwag.Generation.WebApi/Processors/OperationParameterProcessor.cs
```diff
@@ -182,8 +182,18 @@ public bool Process(OperationProcessorContext context)
 
                 if (operationParameter != null)
                 {
-                    operationParameter.Position = position;
                     position++;
+                    operationParameter.Position = position;
+
+                    if (_settings.SchemaType == SchemaType.OpenApi3)
+                    {
+                        operationParameter.IsNullableRaw = null;
```
jeremyVignelles, line 190: What does this mean ?

src/NSwag.CodeGeneration/DefaultParameterNameGenerator.cs
```diff
@@ -21,19 +21,22 @@ public class DefaultParameterNameGenerator : IParameterNameGenerator
         /// <returns>The parameter name.</returns>
         public string Generate(OpenApiParameter parameter, IEnumerable<OpenApiParameter> allParameters)
         {
-            if (string.IsNullOrEmpty(parameter.Name))
+            var name = !string.IsNullOrEmpty(parameter.CustomName) ?
```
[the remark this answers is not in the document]
jeremyVignelles: Didn't understand the question...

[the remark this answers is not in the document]
jeremyVignelles: I get it now. I don't think it is necessary if the spec parameter is named anything else than `x-name` : it is explicit that it is generated by NSwag and I'd argue that the parameter should match on both side.

What's the breaking risk of that change?

src/NSwag.Generation.WebApi/Processors/OperationParameterProcessor.cs
RicoSuter: This is a fix to align web api impl with asp.net core - not really part of this PR.

src/NSwag.CodeGeneration/DefaultParameterNameGenerator.cs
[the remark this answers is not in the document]
RicoSuter: > What's the breaking risk of that change?

Not really a break, in the current PR state you can enable/disable in the spec gen but not in the client gen - just wondering whether we should have a config for both so it's aligned.

src/NSwag.Core/OpenApiParameter.cs
RicoSuter: All properties in JSON Schema/OpenAPI use lower camel case, e.g. "securityDefinitions" - so i would go with x-lowerCamelCase, no?

I like x-originalParameter better.

src/NSwag.Generation.AspNetCore/Processors/OperationParameterProcessor.cs
RicoSuter: Nothing, looks better to me :-)

src/NSwag.Core/OpenApiParameter.cs
jeremyVignelles: Then ok for `x-originalParameter`

src/NSwag.CodeGeneration/DefaultParameterNameGenerator.cs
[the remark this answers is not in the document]
jeremyVignelles: In the current release, what happens if I try to create an API with a different name ? Which name is taken, the param name or the HTTP one ?

The reason why we would introduce another parameter would be to make the new behavior opt-it to not break the current implementations.

src/NSwag.Generation.AspNetCore/Processors/OperationParameterProcessor.cs
jeremyVignelles: Is Position supposed to be 0-based or 1-based ?

jeremyVignelles: with `position = 1` above, that will make you start at 2, which I'm not sure is the wanted start number :) Did you mean to set `position = 0` instead ?

RicoSuter: Oh shit, yes, reverted.

src/NSwag.CodeGeneration/DefaultParameterNameGenerator.cs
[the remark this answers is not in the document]
RicoSuter: > Which name is taken, the param name or the HTTP one ?

Currently it always uses the HTTP name and the other is ignored (ie from an API perspective it doesnt matter). Currently there is no way to preserve the original parameter name.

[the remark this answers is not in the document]
jeremyVignelles: If I have http parameter A and C# parameter B, the name A was used in generated client code, and now it's named B.
It might be breaking in cases like these:
```cs
await client.CallMethod(A: 123);
```

I don't know if it's really a big deal though, that kind of cases should be very rare.

Do you see any other potential breaking change ?

Issue #2596, opened 2019-12-27
13.2.0.0 generating non-working C# client with ChangeType as a culprit


HolopaMir:
Upgrading from 13.1.6.0 to 13.2.0.0 creates client, that compiles, but does not work (Can't convert from System.String).

Problem is when client is generated, it tries to serialize json to types with ChangeType:

```
var responseData_ = response_.Content == null ? null : await response_.Content.ReadAsStringAsync().ConfigureAwait(false);
var result_ = (System.Collections.Generic.ICollection<string>)System.Convert.ChangeType(responseData_, typeof(System.Collections.Generic.ICollection<string>));
```

Old version does this correctly:
```
var objectResponse_ = await ReadObjectResponseAsync<System.Collections.Generic.ICollection<string>>(response_, headers_).ConfigureAwait(false);
return objectResponse_.Object;
```

Here is the bug repro for this issue: https://github.com/Enegia/NSwagGeneratorChangeTypeBugRepro

HolopaMir:
Releates to: https://github.com/RicoSuter/NSwag/issues/1866

pantonis:
Have the same issue here. Just upgrades and the latest upgrade causes this problem

scberr:
I was able to work around this by adding a [Produces("application/json")] to the top of my controller definition. e.g.
```
    [Produces("application/json")]
    [Route("api/[controller]")]
    [ApiController]
    public class MyController : ControllerBase
```

HolopaMir:
@scberr thanks, that fixed problem for us too.

We have tons of controllers, so we end up using global filter (startup.cs):

```
services.AddControllers(options =>
{
	options.Filters.Add(new ProducesAttribute("application/json"));
});
```

RicoSuter:
Strange, i think we need to look into this regression

aherrick:
this also broke my code. Don't have the ability to change controller endpoints. Rolled back 13.1.6.0

RicoSuter:
Must be a regression of this PR:
https://github.com/RicoSuter/NSwag/pull/2587

@Jehoel

Jehoel:
Erk!

I'm away from my desktop where my proper NSwag repo is - I'll be able to look at this after this coming weekend.

RicoSuter:
I hope I fixed this in this commit:
https://github.com/RicoSuter/NSwag/commit/4fbc0ccf65750f6e5d36abb0218da0862b6b96c7

v13.2.1

[...]

Issue #2384, opened 2019-09-02
nswag v13 can't handle text\plain and string response


antonGritsenko:
This is very old issue and it was fixed in v12, but appears again in v13: #701

Controller's attributes:
```
[Produces("text/plain")]
[ProducesResponseType(typeof(FileContentResult), (int)HttpStatusCode.OK)]
[ProducesResponseType(typeof(string), (int)HttpStatusCode.OK)]
[ProducesResponseType(typeof(ErrorResponseMessage), (int)HttpStatusCode.BadRequest)]
[ProducesResponseType(typeof(ErrorResponseMessage), (int)HttpStatusCode.NotFound)]
```

Swagger:
```
"get": {
        "tags": [ "Binary" ],
        "summary": "Download binary as base64 string.",
        "operationId": "DownloadBinaryAsBase64",
        "consumes": [],
        "produces": [ "text/plain" ],
        "parameters": [
          {
            "name": "binaryId",
            "in": "path",
            "description": "Id of binary to fetch.",
            "required": true,
            "type": "string",
            "format": "guid"
          }
        ],
        "responses": {
          "200": {
            "description": "Content downloaded.",
            "schema": { "type": "string" }
          },
          "400": {
            "description": "Request contains wrong data.",
            "schema": { "$ref": "#/definitions/ErrorResponseMessage" }
          },
          "404": {
            "description": "Not Found",
            "schema": { "$ref": "#/definitions/ErrorResponseMessage" }
          }
        }
```

Produced code (part):
```
var status_ = ((int)response_.StatusCode).ToString();
if (status_ == "200") 
{
    var objectResponse_ = await ReadObjectResponseAsync<string>(response_, headers_).ConfigureAwait(false);
    return objectResponse_.Object;
}
```

ReadObjectResponseAsync always try to deserialize the string and of course it will fail.

Revert to v12.2.5 fix the issue.

et1975:
https://github.com/RicoSuter/NSwag/pull/1976/files#diff-4cbdc640407fc6a649e91bcc43983584L17 :(

trucnguyen76:
https://docs.microsoft.com/en-us/aspnet/core/web-api/advanced/formatting?view=aspnetcore-3.1#format-specific-action-results
Seems like you can return plain text with ContentResult, but it seems like you can't wrap ContentResult object with ActionResult

RicoSuter:
So its a regression of this pr?
[...]

PR #3467, opened 2021-05-15
Properly generate client code for string and array of string responses

abborg:
See Github issues #2384 and #2596.

Attempt to resolve both issues handling plain text and JSON lists by
checking for the response's schema type. Treat as plaintext if schema
type is string or marked as produces text/plain content.

### DIFF
--- a/src/NSwag.CodeGeneration/Models/ResponseModelBase.cs
+++ b/src/NSwag.CodeGeneration/Models/ResponseModelBase.cs
@@ -86,8 +86,10 @@ namespace NSwag.CodeGeneration.Models
 
         /// <summary>Gets a value indicating whether the response requires a text/plain content.</summary>
         public bool IsPlainText =>
-            !_response.Content.ContainsKey("application/json") &&
-            (_response.Content.ContainsKey("text/plain") || _operationModel.Produces == "text/plain");
+            (_response.Content.ContainsKey("application/json") &&
+            _response.Schema != null && _response.Schema?.Type == JsonObjectType.String) ||
+            (!_response.Content.ContainsKey("application/json") &&
+            (_response.Content.ContainsKey("text/plain") || _operationModel.Produces == "text/plain"));
 
         /// <summary>Gets a value indicating whether this is a file response.</summary>
         public bool IsFile => IsSuccess && _response.IsBinary(_operation);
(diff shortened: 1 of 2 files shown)


src/NSwag.CodeGeneration/Models/ResponseModelBase.cs
```diff
@@ -86,8 +86,10 @@ public bool IsDate
 
         /// <summary>Gets a value indicating whether the response requires a text/plain content.</summary>
         public bool IsPlainText =>
-            !_response.Content.ContainsKey("application/json") &&
-            (_response.Content.ContainsKey("text/plain") || _operationModel.Produces == "text/plain");
+            (_response.Content.ContainsKey("application/json") &&
+            _response.Schema != null && _response.Schema?.Type == JsonObjectType.String) ||
```
RicoSuter, line 90: The problem is that this is actually a hack for ASP.NET Core but from the perspective of OAI not correct... i.e. a response of type JSON and string should return `"foo bar"` and not `foo bar` - the problem is that ASP.NET Core (at least in older versions) returned a string and not a JSON string changing the type of the response from JSON to text/plain.

RicoSuter: Ref: https://github.com/aspnet/Mvc/issues/4945

abborg: I was not aware of that specific hack for ASP.NET Core and I see how this change would not be a sufficient fix. It seems that in some locations throughout the codebase preference is given to "application/json" over "text/plain" for response types. Would it then be sufficient to select a response type of "application/json" when there are both "application/json" and "text/plain" are available? This would be a special case for strings.

src/NSwag.CodeGeneration/Models/ResponseModelBase.cs
```diff
@@ -85,9 +85,16 @@ public bool IsDate
...
-            !_response.Content.ContainsKey("application/json") &&
-            (_response.Content.ContainsKey("text/plain") || _operationModel.Produces == "text/plain");
+        public bool IsPlainText
+        {
+            get
+            {
+                JsonObjectType primitive = JsonObjectType.String | JsonObjectType.Integer | JsonObjectType.Number | JsonObjectType.Boolean;
+                return ActualResponseSchema != null &&
+                    (ActualResponseSchema.Type & primitive) == ActualResponseSchema.Type &&
+                    (ActualResponseSchema.Type | primitive) == primitive;
```
abborg, line 95: If the response schema only contains primitive data types, then the response can be considered plain text.

abborg:
Superseded by #3535

Issue #3476, opened 2021-05-20
Relax package references in NSwag.Generation.AspNetCore to work with v5 assemblies


jafin:
Please consider relaxing the dependency requirements in the AspNetCore project to build without warning with v5 assemblies on NET3.1.

```
 NSwag.Generation.AspNetCore 13.11.1 requires Microsoft.Extensions.DependencyInjection.Abstractions (>= 3.1.0 && < 4.0.0) but version Microsoft.Extensions.DependencyInjection.Abstractions 5.0.0 was resolved.
D:\a\1\s\proj\proj.csproj : warning NU1608: Detected package version outside of dependency constraint: NSwag.Generation.AspNetCore 13.11.1 requires Microsoft.Extensions.Options (>= 3.1.0 && < 4.0.0) but version Microsoft.Extensions.Options 5.0.0 was resolved.
  All projects are up-to-date for restore.
D:\a\1\s\proj\proj.csproj : warning NU1608: Detected package version outside of dependency constraint: NSwag.Generation.AspNetCore 13.11.1 requires Microsoft.Extensions.DependencyInjection.Abstractions (>= 3.1.0 && < 4.0.0) but version Microsoft.Extensions.DependencyInjection.Abstractions 5.0.0 was resolved.
D:\a\1\s\proj\proj.csproj : warning NU1608: Detected package version outside of dependency constraint: NSwag.Generation.AspNetCore 13.11.1 requires Microsoft.Extensions.Options (>= 3.1.0 && < 4.0.0) but version Microsoft.Extensions.Options 5.0.0 was resolved.
```

PR #3477, opened 2021-05-20
fix: (Dependencies): Bump dependencies for NSwag.Generation.AspNetCore to work with v5 assemblies.

jafin:
Bump depedencies for aspnetCore to allow Microsoft.Extensions.DependencyInjection.Abstractions and Extensions.Optios v5  (Both of which are compatible with netstandard2.0) #3476

The v5 assemblies have a base requirement of net standard 2.0.


RicoSuter:
Thanks for the PR.

Issue #3463, opened 2021-05-11
rxjs version 7 not working with Nswag


florianraffl:
Hi there,

we`re having a problem with the new rxjs version 7 and Nswag generating our typescript client for some C# rest crud operations.
It looks like in this auto generated "getAll" method:
```
protected processGetAll(response: HttpResponseBase): Observable<Some_DTO[]> {
        const status = response.status;
        const responseBlob =
            response instanceof HttpResponse ? response.body :
            (<any>response).error instanceof Blob ? (<any>response).error : undefined;

        let _headers: any = {}; if (response.headers) { for (let key of response.headers.keys()) { _headers[key] = response.headers.get(key); }}
        if (status === 200) {
            return blobToText(responseBlob).pipe(_observableMergeMap(_responseText => {
            let result200: any = null;
            let resultData200 = _responseText === "" ? null : JSON.parse(_responseText, this.jsonParseReviver);
            if (Array.isArray(resultData200)) {
                result200 = [] as any;
                for (let item of resultData200)
                    result200!.push(Some_DTO.fromJS(item));
            }
            else {
                result200 = <any>null;
            }
            return _observableOf(result200);
            }));
        } else if (status !== 200 && status !== 204) {
            return blobToText(responseBlob).pipe(_observableMergeMap((_responseText: string) => {
            return throwException("An unexpected server error occurred.", status, _responseText, _headers);
            }));
        }
        return _observableOf<Some_DTO[]>(<any>null);
    }
}
```
the `_observableMergeMap` method parameter has to be typed like that:
`_observableMergeMap((_responseText: string) => { ... }`

Otherwise it will throw that error:
```
Error: src/app/core/services/webapiclient.service.ts:95:74 - error TS2345: Argument of type 'unknown' is not assignable to parameter of type 'string'.

95             let resultData200 = _responseText === "" ? null : JSON.parse(_responseText, this.jsonParseReviver);
```

I also checked the nswag config and found a parameter:
```
...
"rxJsVersion": 6.0,
...
```
I also set that parameter to 7.0 but that didn't change a thing. 

So now my question is, is it possible to change the generation of the `_observableMergeMap` if the config parameter `"rxJsVersion"` is 7.0 to the typed one? 

Best regards,
Florian Raffl

Thomas-Lederer:
+1

RicoSuter:
Can someone create a PR with this fix?
Is this a breaking change for some users (e.g. users of older rxjs versions)?

[...]

PR #3482, opened 2021-05-26
typescript code generation fixes for rxjs 7

florianraffl:
#3463

### DIFF
--- a/src/NSwag.CodeGeneration.TypeScript/Models/TypeScriptFrameworkRxJsModel.cs
+++ b/src/NSwag.CodeGeneration.TypeScript/Models/TypeScriptFrameworkRxJsModel.cs
@@ -8,29 +8,32 @@
 
 namespace NSwag.CodeGeneration.TypeScript.Models
 {
-    /// <summary>RxJs specific information.</summary>
-    public class TypeScriptFrameworkRxJsModel
-    {
-        private readonly TypeScriptFrameworkModel _model;
+	/// <summary>RxJs specific information.</summary>
+	public class TypeScriptFrameworkRxJsModel
+	{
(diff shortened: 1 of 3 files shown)


src/NSwag.CodeGeneration.TypeScript/Models/TypeScriptFrameworkRxJsModel.cs
```diff
@@ -8,29 +8,32 @@
```
RicoSuter, line 8: Please use spaces instead of tabs

florianraffl:
@RicoSuter do you know why this CI build check is failing? I think it doesn't have anything to do with my changes or am I wrong?

RicoSuter:
Hmm, building the MSI installer times out in the CI pipeline... no clue how to fix it except excluding it :-)

RicoSuter:
Ok, maybe its fixed with this PR: https://github.com/RicoSuter/NSwag/pull/3493
Can you update your branch with master so we see whether it fixes CI here?

florianraffl:
We had nearly the same problem in our company ... but on an self hosted azure pipelines agent ... there we fixed it by adding the build argument `/p:RunWixToolsOutOfProc=true` to the build step according to:
https://stackoverflow.com/questions/42167960/building-wix-on-visual-studio-team-services-times-out
maybe this also helps you :-)

Ok thanks for the response, I will make a dummy commit to rerun the CI build :-)

RicoSuter:
Ok looks good, ready to merge i suppose?

florianraffl:
Yea that would be perfekt! :-)

RicoSuter:
Thanks for the PR.

PR #3487, opened 2021-05-27
Improve GenerateOperationIds

danielwinkler:
The current version of `GenerateOperationIds` handles the case when we overload a method on the Controller, and one method would be returning an collection.

I am running often into the problem that we annotate methods with `[HttpGet]` and `[HttpHead]`.
In this case one of them gets randomly assigned the the "All" suffix, which is prone to error.

This PR tries to add "All" if it makes sense (if all overloads return collections, better not do that at all).
Second, it will try to add the HTTP method in uppercase.
Third, it will fallback to appending numbers.

Instead of
```json
{
    "paths": {
      "/api/Values": {
        "get": {
          "operationId": "Values_GetAll",
        },
        "head": {
          "operationId": "Values_Get",
        },
      "/api/Values/{id}": {
        "get": {
          "operationId": "Values_Get2",
        },
        "head": {
          "operationId": "Values_Get3",

```

we will have
```json
{
    "paths": {
      "/api/Values": {
        "get": {
          "operationId": "Values_GetAll_GET",
        },
        "head": {
          "operationId": "Values_GetAll_HEAD",
        },
      "/api/Values/{id}": {
        "get": {
          "operationId": "Values_Get_GET",
        },
        "head": {
          "operationId": "Values_Get_HEAD",
        },

```

for the example `NSwag/src/NSwag.Sample.NET50/Controllers/ValuesController.cs`

```csharp
        [HttpGet]
        [HttpHead]
        public ActionResult<IEnumerable<Person>> Get()
        {
            return new Person[] { };
        }

        // GET api/values/5
        [HttpGet("{id}")]
        [HttpHead("{id}")]
        public ActionResult<TestEnum> Get(int id)
        {
            return TestEnum.Foo;
        }
```

### DIFF
--- a/src/NSwag.Core/OpenApiDocument.cs
+++ b/src/NSwag.Core/OpenApiDocument.cs
@@ -274,8 +274,6 @@ namespace NSwag
         /// <summary>Generates missing or non-unique operation IDs.</summary>
         public void GenerateOperationIds()
         {
-            // TODO: Improve this method
-
             // Generate missing IDs
             foreach (var operation in Operations.Where(o => string.IsNullOrEmpty(o.Operation.OperationId)))
             {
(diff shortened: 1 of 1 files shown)


danielwinkler:
Merged the change into #3488

PR #3488, opened 2021-05-27
Support for HttpMethodAttribute Name

danielwinkler:
This PR adds support for HttpAttribute names, e.g.

```csharp
        // GET api/values/5
        [HttpGet("{id}/foo", Name = "GET_AttributeName_TEST")]
        [HttpHead("{id}/foo", Name = "HEAD_AttributeName_TEST")]
        public ActionResult<string> GetFooBar(int id)
        {
            return "value";
        }
```

will create 
```json
    "/api/Values/{id}/foo": {
      "get": {
        "operationId": "GET_AttributeName_TEST",
        "parameters": [
          {
            "name": "id",
            "in": "path",
            "required": true,
            "schema": {
              "type": "integer",
              "format": "int32"
            },
            "x-position": 1
          }
        ],
        "responses": {        }
      },
      "head": {
        "operationId": "HEAD_AttributeName_TEST",
        "parameters": [
          {
            "name": "id",
            "in": "path",
            "required": true,
            "schema": {
              "type": "integer",
              "format": "int32"
            },
            "x-position": 1
          }
        ],
        "responses": {
        }
```

In addition it improves the behavior when no name is present and there are name collisions:
The current version of `GenerateOperationIds` handles the case when we overload a method on the Controller, and one method would be returning an collection.

I am running often into the problem that we annotate methods with `[HttpGet]` and `[HttpHead]`.
In this case one of them gets randomly assigned the the "All" suffix, which is prone to error.

This PR tries to add "All" if it makes sense (if all overloads return collections, better not do that at all).
Second, it will try to add the HTTP method in uppercase.
Third, it will fallback to appending numbers.

Instead of
```json
{
    "paths": {
      "/api/Values": {
        "get": {
          "operationId": "Values_GetAll",
        },
        "head": {
          "operationId": "Values_Get",
        },
      "/api/Values/{id}": {
        "get": {
          "operationId": "Values_Get2",
        },
        "head": {
          "operationId": "Values_Get3",

```

we will have
```json
{
    "paths": {
      "/api/Values": {
        "get": {
          "operationId": "Values_GetAll_GET",
        },
        "head": {
          "operationId": "Values_GetAll_HEAD",
        },
      "/api/Values/{id}": {
        "get": {
          "operationId": "Values_Get_GET",
        },
        "head": {
          "operationId": "Values_Get_HEAD",
        },

```

for the example `NSwag/src/NSwag.Sample.NET50/Controllers/ValuesController.cs`

```csharp
        [HttpGet]
        [HttpHead]
        public ActionResult<IEnumerable<Person>> Get()
        {
            return new Person[] { };
        }

        // GET api/values/5
        [HttpGet("{id}")]
        [HttpHead("{id}")]
        public ActionResult<TestEnum> Get(int id)
        {
            return TestEnum.Foo;
        }
```

### DIFF
--- a/src/NSwag.Generation.AspNetCore/AspNetCoreOpenApiDocumentGenerator.cs
+++ b/src/NSwag.Generation.AspNetCore/AspNetCoreOpenApiDocumentGenerator.cs
@@ -220,7 +220,7 @@ namespace NSwag.Generation.AspNetCore
                             Operation = new OpenApiOperation
                             {
                                 IsDeprecated = IsOperationDeprecated(item.Item1, controllerActionDescriptor, method),
-                                OperationId = GetOperationId(document, controllerActionDescriptor, method),
+                                OperationId = GetOperationId(document, controllerActionDescriptor, method, httpMethod),
                                 Consumes = apiDescription.SupportedRequestFormats
                                    .Select(f => f.MediaType)
                                    .Distinct()
@@ -404,7 +404,7 @@ namespace NSwag.Generation.AspNetCore
             return true;
         }
 
-        private string GetOperationId(OpenApiDocument document, ControllerActionDescriptor actionDescriptor, MethodInfo method)
+        private string GetOperationId(OpenApiDocument document, ControllerActionDescriptor actionDescriptor, MethodInfo method, string httpMethod)
         {
             string operationId;
 
@@ -412,9 +412,23 @@ namespace NSwag.Generation.AspNetCore
                 .GetCustomAttributes()
                 .FirstAssignableToTypeNameOrDefault("SwaggerOperationAttribute", TypeNameStyle.Name);
 
+            dynamic httpAttribute = null;
+            if (!string.IsNullOrWhiteSpace(httpMethod))
+            {
+                var attributeName = Char.ToUpperInvariant(httpMethod[0]) + httpMethod.Substring(1).ToLowerInvariant();
+                var typeName = string.Format("Microsoft.AspNetCore.Mvc.Http{0}Attribute", attributeName);
+                httpAttribute = method
+                    .GetCustomAttributes()
+                    .FirstAssignableToTypeNameOrDefault(typeName);
+            }
+
             if (swaggerOperationAttribute != null && !string.IsNullOrEmpty(swaggerOperationAttribute.OperationId))
             {
                 operationId = swaggerOperationAttribute.OperationId;
+            }
+            else if (httpAttribute != null && !string.IsNullOrWhiteSpace(httpAttribute.Name))
+            {
+                operationId = httpAttribute.Name;
             }
             else if (Settings.UseRouteNameAsOperationId && !string.IsNullOrEmpty(actionDescriptor.AttributeRouteInfo.Name))
             {
--- a/src/NSwag.Generation.AspNetCore/AspNetCoreOpenApiDocumentGenerator.cs
+++ b/src/NSwag.Generation.AspNetCore/AspNetCoreOpenApiDocumentGenerator.cs
@@ -426,7 +426,7 @@ namespace NSwag.Generation.AspNetCore
             {
                 operationId = swaggerOperationAttribute.OperationId;
             }
-            else if (httpAttribute != null && !string.IsNullOrWhiteSpace(httpAttribute.Name))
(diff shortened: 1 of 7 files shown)


RicoSuter:
Can you check with this PR so that we maybe can have both things in a single PR?
https://github.com/RicoSuter/NSwag/pull/3487

danielwinkler:
The other one is also from me, if you like both approaches i will add #3487 here and close the other.

src/NSwag.Generation.WebApi/WebApiOpenApiDocumentGenerator.cs
```diff
@@ -290,18 +290,32 @@ private static IEnumerable<MethodInfo> GetActionMethods(Type controllerType)
...
 
             dynamic swaggerOperationAttribute = method
                 .GetCustomAttributes()
                 .FirstAssignableToTypeNameOrDefault("SwaggerOperationAttribute", TypeNameStyle.Name);
 
+            dynamic httpAttribute = null;
+            if (!string.IsNullOrWhiteSpace(httpMethod))
+            {
+                var attributeName = Char.ToUpperInvariant(httpMethod[0]) + httpMethod.Substring(1).ToLowerInvariant();
+                var typeName = string.Format("Microsoft.AspNetCore.Mvc.Http{0}Attribute", attributeName);
```
RicoSuter, line 305: Sorry, but this is the legacy Web API generator which should not be used for ASP.NET Core... I think you're using the wrong generator in your project :-)

src/NSwag.Generation.AspNetCore/AspNetCoreOpenApiDocumentGenerator.cs
```diff
@@ -404,18 +404,32 @@ private bool RunOperationProcessors(OpenApiDocument document, ApiDescription api
...
+                httpAttribute = method
+                    .GetCustomAttributes()
+                    .FirstAssignableToTypeNameOrDefault(typeName);
+            }
+
             if (swaggerOperationAttribute != null && !string.IsNullOrEmpty(swaggerOperationAttribute.OperationId))
             {
                 operationId = swaggerOperationAttribute.OperationId;
             }
+            else if (httpAttribute != null && !string.IsNullOrWhiteSpace(httpAttribute.Name))
```
RicoSuter, line 429: Maybe we should add a setting, eg Settings.UseHttpAttributeNameAsOperationId (better name?, default false) to not break existing users and make it less opinionated?

RicoSuter:
"GenerateOperationIds" has only an effect if operation IDs are missing... ie if the spec was invalid in the first place :-)
Will this PR change this generation to a more stable version but otherwise will keep current behavior? 
Need to understand the impact on existing users (ie how big is the potential breaking change)?

src/NSwag.Generation.WebApi/WebApiOpenApiDocumentGenerator.cs
danielwinkler: I was doing it because the code was equal to ASP.NET Core, but just to have it for both running.
The method from which i am retrieving the name is already present in the code.

see https://docs.microsoft.com/en-us/aspnet/web-api/overview/web-api-routing-and-actions/attribute-routing-in-web-api-2#adding-route-attributes

src/NSwag.Generation.AspNetCore/AspNetCoreOpenApiDocumentGenerator.cs
danielwinkler: I added a setting, please see latest commits

danielwinkler:
> "GenerateOperationIds" has only an effect if operation IDs are missing... ie if the spec was invalid in the first place :-)
> Will this PR change this generation to a more stable version but otherwise will keep current behavior?
> Need to understand the impact on existing users (ie how big is the potential breaking change)?

Yes

So I first will try to resolve it with adding "All" , but only if it makes sense
e.g. it doesn't make sense if both overloads return a collection, because then it is becoming non-deterministic

If after applying this approach, there are still collisions then I will append the HTTP method, if it makes sense (it doesn't make sense if it is all GETs)

Then I fallback to the enumeration

RicoSuter:
Thanks for the PR.

RicoSuter:
Maybe it would be good to also add a unit test to cover this so we avoid regressions in the future

PR #3510, opened 2021-06-11
Automatically generate httpmethod, when action for controller does no…

zhenxin0603:
Automatically generate httpmethod, when active does not specify httpmethod.
When the parameter source of action is body, httmethod is set to POST, otherwise it is GET.
sorry,my bad english.


zhenxin0603:
controller and action
image

before:
image
after:
image

PR #3517, opened 2021-06-17
Add setting "UseDefaultValueOptionalParameter" to use default value for optional parameters instead of null

DanikRaikhlin:
This adds the abillity to use the default value provided in the open api spec in the generated optional parameters. Both GenerateOptionalParameters and UseDefaultValueOptionalParameter should be enabled and the parameter.DefaultValue should be present.

`public async System.Threading.Tasks.Task TestWithDefaultEnumAsync(MyEnum? myEnum = MyNamespace.MyEnum.Four, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken))`


DanikRaikhlin:
@RicoSuter could you please give feedback if this change is desired.

PR #3535, opened 2021-07-01
Fixed throwing exception for plain text responses

### LINKED ISSUE #3512: CSharp Client Generator - Not returning result nor throwing exception for non-success HTTP status codes if IsPlainText = true
Hi, 

I'm having an issue trying generate a C# client out of an OpenAPI spec which contains plain/text responses.

Specifically, I have this error telling me that **not all code paths return a value** if the operations have **non-success HTTP status codes** as you can see below:
image

I think that the problem happens here:
https://github.com/RicoSuter/NSwag/blob/1c836ee3cc766ce25f3170618a5d12e5e96fde61/src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.ProcessResponse.liquid#L15

Basically we will need to **add an else block** to deal with that non-success HTTP codes and throw an exception properly.

Does that make sense? If it does I can fix this and raise a PR.
Thanks!

icnocop:
Fixes #3512

### DIFF
--- a/src/NSwag.CodeGeneration.CSharp.Tests/PlainResponseTests.cs
+++ b/src/NSwag.CodeGeneration.CSharp.Tests/PlainResponseTests.cs
@@ -61,14 +61,14 @@ namespace NSwag.CodeGeneration.CSharp.Tests
 }";
             var document = await OpenApiDocument.FromJsonAsync(json);
 
-            //// Act
+            // Act
             var codeGenerator = new CSharpClientGenerator(document, new CSharpClientGeneratorSettings
             {
                 GenerateClientInterfaces = true
             });
             var code = codeGenerator.GenerateFile();
 
-            //// Assert
+            // Assert
             Assert.Contains("public async System.Threading.Tasks.Task<System.Collections.Generic.ICollection<string>> GetIenumerableAsync(", code);
             Assert.Contains("await ReadObjectResponseAsync<System.Collections.Generic.ICollection<string>>(", code);
         }
@@ -103,14 +103,14 @@ namespace NSwag.CodeGeneration.CSharp.Tests
 }";
             var document = await OpenApiDocument.FromJsonAsync(json);
 
-            //// Act
+            // Act
             var codeGenerator = new CSharpClientGenerator(document, new CSharpClientGeneratorSettings
             {
                 GenerateClientInterfaces = true
             });
             var code = codeGenerator.GenerateFile();
 
-            //// Assert
+            // Assert
             Assert.Contains("public async System.Threading.Tasks.Task<string> PlainAsync(", code);
             Assert.Contains("(string)System.Convert.ChangeType(responseData_, typeof(string));", code);
         }
@@ -146,14 +146,14 @@ namespace NSwag.CodeGeneration.CSharp.Tests
 }";
             var document = await OpenApiDocument.FromJsonAsync(swagger);
 
-            //// Act
+            // Act
             var codeGenerator = new CSharpClientGenerator(document, new CSharpClientGeneratorSettings
             {
                 GenerateClientInterfaces = true
             });
             var code = codeGenerator.GenerateFile();
 
-            //// Assert
+            // Assert
             Assert.Contains("public async System.Threading.Tasks.Task<System.Collections.Generic.ICollection<string>> GetIenumerableAsync(", code);
             Assert.Contains("await ReadObjectResponseAsync<System.Collections.Generic.ICollection<string>>(", code);
         }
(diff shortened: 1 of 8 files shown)


icnocop:
Supersedes #3467
Also fixes #2384 and #2596

src/NSwag.CodeGeneration/Models/ResponseModelBase.cs
```diff
@@ -85,9 +85,16 @@ public bool IsDate
         }
 
         /// <summary>Gets a value indicating whether the response requires a text/plain content.</summary>
-        public bool IsPlainText =>
-            !_response.Content.ContainsKey("application/json") &&
-            (_response.Content.ContainsKey("text/plain") || _operationModel.Produces == "text/plain");
+        public bool IsPlainText
```
RicoSuter, line 88: Is this necessary? 
Changing this is dangerous and a potential breaking change.

icnocop: Yes.
See https://github.com/RicoSuter/NSwag/issues/2596#issuecomment-872618108

RicoSuter: The problem is that this might be wrong... response type can be a JSON encoded integer which would lead to IsPlainText == true, no?
I think we need to check the produces/content.. why would the current implementation isPlainText evaluate to true for "application/problem+json"?

icnocop: > 
> 
> The problem is that this might be wrong... response type can be a JSON encoded integer which would lead to IsPlainText == true, no?
> I think we need to check the produces/content.. why would the current implementation isPlainText evaluate to true for "application/problem+json"?

I just added another test to illustrate the problem. The API endpoint in the test can produce multiple response types. If the response indicates success (200), then the response is "text/plain"; otherwise the response is "application/problem+json". However, without this change, when the c# client code is generated, non-success responses are still treated as "text/plain".

icnocop: > ... response type can be a JSON encoded integer

I'm not familiar with this response type. Can you create/describe a sample API endpoint method definition in C# for example?

Thank you.

icnocop: I have resolved this issue in my last commit.

Thank you.

RicoSuter: I wonder why you need to check ActualTypeSchema at all - ie when the response type is text/plain it's always a plain text response, ie also when there is no response schema or it is just any schema, no? If there is a schema with eg properties but the response type is text/plain then i'd say the spec is actually wrong..

RicoSuter: ... or do you want to go with the assumption that it's actually JSON if there is a spec, even if the response type indicates otherwise?

icnocop: Thank you for following up. I just pushed another commit which removes checking ActualTypeSchema. Thank you.

RicoSuter: Sorry, my question was not really about ActualTypeSchema (we should use that to resolve refs) but checking the .Type, ie changing the property in the first place?

icnocop: I figured out the underlying issue; `_operationModel.Produces` in [IsPlainText](https://github.com/RicoSuter/NSwag/blob/master/src/NSwag.CodeGeneration/Models/ResponseModelBase.cs#L90) shouldn't be used because [Produces](https://github.com/RicoSuter/NSwag/blob/master/src/NSwag.CodeGeneration/Models/OperationModelBase.cs#L265-L267) returns `SuccessResponse?.Produces`. The generated code for a non-success response, in the case where the success response (ex. text/plain) returns a different mime type than a non-success response (ex. application/problem+json), will erroneously assume the response is the same as the success response type (ex. text/plain), and fail when trying to convert the response.

I have pushed a commit with this fix.

Thank you.

icnocop:
Also fixes #3543

src/NSwag.CodeGeneration.CSharp.Tests/PlainResponseTests.cs
```diff
@@ -39,15 +103,59 @@ public async Task When_openapi3_reponse_contains_plain_text_then_Convert_is_gene
...
+        ""produces"": [
+          ""text/plain"",
+          ""application/json"",
+          ""text/json""
+        ],
+        ""responses"": {
+          ""200"": {
+            ""description"": ""Success"",
+            ""schema"": {
+              ""type"": ""array"",
```
GregaMohorko, line 136: Can you add a test for having type as string instead of array? Because it doesn't work when the type is directly a string. The assertion then being `Assert.Contains("var responseData_ = response_.Content == null ? null : await response_.Content.ReadAsStringAsync().ConfigureAwait(false);")` and `Assert.Contains("var result_ = (string)System.Convert.ChangeType(responseData_, typeof(string));")`.

RicoSuter: Can you or someone create a PR for this?

jochenwezel:
Deine Analyse war leider falsch in Bezug auf die Lösung im Client mithilfe der settings.
Im NSWAG Projekt gibt es derzeit ein offenes Ticket zu einem Fehlverhalten (Bug) der NSWAG Komponenten: https://github.com/RicoSuter/NSwag/issues/2384

Zur (vorläufigen) Behebung des Fehlers muss das Attribut `[Produces("text/plain", Type = typeof(string))]`gesetzt werden.

Issue #3506, opened 2021-06-08
_observableOf types misformed


wratho:
![image](https://user-images.githubusercontent.com/28812961/121263184-e183ab00-c87a-11eb-9c35-23bcb7553c00.png)

I recently updated from rxjs 6.6.7 to rxjs 7.1. I am also strictly a javascript developer and do not really understand how the NSwag product works (other coworkers set this up and helped me get this far in diagnosing the issues).

What i know:
The image posted above is the result of upgrading from rxjs 6.6.7 to 7.1. Upon research i found that NSwag was updated recently to fix rxjs 7. I updated the C# solution's package to the latest NSwag, and updated the rxjs version setting to 7.1. Regenerating the client did not fix this issue. If i remove everything in the `<>` it works.

A coworker sent me this file (https://github.com/RicoSuter/NSwag/blob/4f345e1adc9c49d8498d4018250760aa69dd8dd2/src/NSwag.CodeGeneration.TypeScript/Templates/Client.ProcessResponse.Return.liquid) with the claims that if we change `return {{ Framework.RxJs.ObservableOfMethod }}<{{ operation.ResultType }}>(<any>null);`
to: `return {{ Framework.RxJs.ObservableOfMethod }}(<any>null);` it would fix my issue.

as well 
![image](https://user-images.githubusercontent.com/28812961/121270283-054cee00-c887-11eb-9cda-8dd65ca2de89.png)
this appears to fix my issue as well. It appears that maybe the any cast here is losing the list concept (maybe)?

dolphinsd:
Issue with upgrade from rxjs 6.6.7 to rxjs 7.1

RicoSuter:
Can you create a PR? Would this break older versions of rxjs?

wratho:
> Can you create a PR? Would this break older versions of rxjs?

i really am not sure about breaking older versions. Like i said in my original post i'm very much just the front end guy and am not entirely sure everything that's going on with NSwag and the API. I can see if some of my colleagues would know more about it. 

Was also more curious if that suggested change was too much when it also looks like keeping the result type where it is now, but also adding more return type to the <any>null also seemed to fix the issue. I'm not sure which direction you'd want to go.

wratho:
With the release 13.12, and rebuilding the api with NSwag it looks like the change from #3544 may have made more changes than we intended? It doesn't appear to have any build issues with 7.X so far, but seems goofy?
[...]

PR #3544, opened 2021-07-06
update generated typescript to work with rxjs 7

dhwed:
@wratho has tested to confirm that this new generated code works with both the new rxjs 7 and with rxjs 6.

see #3506


RicoSuter:
Thanks for the PR.

PR #3573, opened 2021-07-28
Add support for HttpContext parameter in angular generation

bernard291:
Angular 12 provides HttpContext  that offer the ability to store and retrieve custom metadata for requests, especially in interceptors.

- Adds the httpContext parameter as the last optional parameter in the typescript generation for angular.


RicoSuter:
Fine for me...

But we also need to add it to the CLI command class (the settings property) and to the XAML/UI.

Issue #3568, opened 2021-07-22
Tons of errors after upgrading to the latest version of Nswag Studio


franconeilglovasa:
Hi!

I've got tons of errors after upgrading Nswag Studio. How can I fix it? How can I downgrade or if anyone has the link of the previous version of NSwag Studio?

this is one of the functions generated by NSWAG:

` return _observableOf<AddOrEditUserResponseVM>(<any[]>null);`

before it was:

`return _observableOf<boolean>(<any>null);`

Im trying to generate Typescript client code from c#

Error:

```
{
	"resource": "/c:/Users/admin/source/repos/TaskApp/ClientApp/src/app/swagger-generated.ts",
	"owner": "typescript",
	"code": "2769",
	"severity": 8,
	"message": "No overload matches this call.\n  Overload 1 of 20, '(...args: (SchedulerLike | AddOrEditUserResponseVM)[]): Observable<AddOrEditUserResponseVM>', gave the following error.\n    Argument of type 'any[]' is not assignable to parameter of type 'SchedulerLike | AddOrEditUserResponseVM'.\n      Type 'any[]' is missing the following properties from type 'AddOrEditUserResponseVM': init, toJSON, clone\n  Overload 2 of 20, '(a: AddOrEditUserResponseVM): Observable<AddOrEditUserResponseVM>', gave the following error.\n    Argument of type 'any[]' is not assignable to parameter of type 'AddOrEditUserResponseVM'.\n  Overload 3 of 20, '(...args: AddOrEditUserResponseVM[]): Observable<AddOrEditUserResponseVM>', gave the following error.\n    Argument of type 'any[]' is not assignable to parameter of type 'AddOrEditUserResponseVM'.",
	"source": "ts",
	"startLineNumber": 142,
	"startColumn": 55,
	"endLineNumber": 142,
	"endColumn": 66
}
```

![image](https://user-images.githubusercontent.com/68049032/126586520-6a9ec14f-897a-4594-ae7b-08722c49c3f9.png)

franconeilglovasa:
I think there's a problem when generating swagger. It returns an [].. see comparison between v13.11 vs v13.12

![image](https://user-images.githubusercontent.com/68049032/126590342-43e38370-202b-41e5-8cff-f8979b83b4e8.png)

jeremyVignelles:
Please report your issue properly :
- Don't use screenshots, but do paste the errors and coxe in a code block. That will help other people with the same error find your issue.
[...]

PR #3586, opened 2021-08-05
Patches/working with strict

didii:
Re-fix for #3568

* Add `null` as the return type
* Remove casting from `_observableOf(null)`

Note: I have no idea how to test this code, so I kept the changes as small as I could. Please doublecheck this code before completing this PR.

### DIFF
--- a/src/NSwag.CodeGeneration.TypeScript/Templates/AngularClient.liquid
+++ b/src/NSwag.CodeGeneration.TypeScript/Templates/AngularClient.liquid
@@ -39,7 +39,7 @@
 {% for operation in Operations -%}
 
     {% template Client.Method.Documentation %}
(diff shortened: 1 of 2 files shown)


RicoSuter:
These template changes can be easily tested on any project by overwriting the templates with the same changes in this PR: 
https://github.com/RicoSuter/NSwag/wiki/Templates

didii:
@RicoSuter ok, I managed to test it. It works in my medium sized application for both `strict: true` and `strict: false`. But as expected these are pretty breaking for `strict: true` since everything that uses the service suddenly needs to take into account that the return type can be `null`.

GeorgDangl:
I'm not sure if the project qualifies for a "bigger project", it's got around 15k lines in the generated client after formatting with Prettier.

I've just tested it, and everything seems to work for me.

SebastianBienert:
Hi, any updates on this? Checked it with my project and it works as expected

Issue #1613, opened 2018-09-17
Difference in OffsetMomentJS between parameters in the body and in the querystring


jochenjonc:
When generating TypeScript code with the dateTimeType set to "OffsetMomentJS" it only adjusts the offset for parameters that are in the body of a HTTP action.

`this.plannedStartDate.toISOString(true)`

But for parameters that are passed via the querystring the following is still used:

`url_ += "start=" + encodeURIComponent(start ? "" + start.toJSON() : "") + "&";`

Is it possible to change the toJSON() here to toISOString(true), keeping the offset?

RSuter:
We need to use the logic from here also for parameters: 
https://github.com/RSuter/NJsonSchema/blob/8e5bc578ee740384c0b468dcb787f4c802a51e3e/src/NJsonSchema.CodeGeneration.TypeScript/DataConversionGenerator.cs#L88

Tasks: 

- Extract this internal logic so that it can also be used for params
- Deploy new NJS package
- Update NSwag and use this

jochenjonc:
I recently encountered the same issue, but then with moment.Duration and the conversion to TimeSpan.

Will this be fixed with the same case?

PR #3660, opened 2021-10-13
Add logic for date and datetimes from the body to the request url #1613

bjarne-callewaert:
This PR resolves an issue where the dates and dateTimes in the request url are not formatted to the correct string value (e.g. when OffsetMomentJS is chosen in the settings, the offset was lost due to the .toJSON()) (https://github.com/RicoSuter/NSwag/issues/1613)

This logic was copied from NJS, where this was already implemented for dateTimes that live inside the body of a request


RicoSuter:
Thanks for the PR.

PR #3661, opened 2021-10-13
Parameter of type DateTime should only be nullable when it is not required

mhartigh:
Since #3346 parameters of type Datetime which are required are rendered as nullable in the header section while the parameter itself in the method isn't nullable. Because of this the generated client does not compile.


pinpoint-solutions:
We also have this particular issue on a project. DateTime should only be nullable when not required, otherwise not-nullable.

PR #3667, opened 2021-10-18
Make typings for `axios` more explicit on `AxiosResponse`

dubzzz:
Hello,

Following recent versions bumps of axios, the `AxiosResponse` moved from a default being `any` to one being `unknown`. When bumping to `0.23.0`, the code built by nswag breaks as `response.data` being an `unknown` instead of an `any` becomes incompatible with part of the generated code.

I had two solutions in mind but there are probably others and maybe some better ones:
- Option 1: Transform `AxiosResponse` (defaults to `any`) into `AxiosResponse<string>` (the one if this commit)
- Option 2: Adapt code downstream to do the `as string` whenever required

Here is the code I get with my current setup of nswag:

```ts
    addAccessRightMetric(applicationId: string, body: CreateApplicationArmRequest | undefined , cancelToken?: CancelToken | undefined): Promise<AccessRightMetric> {
        let url_ = this.baseUrl + "/api/workspace/accessrights/AddAccessRightMetric/{applicationId}";
        if (applicationId === undefined || applicationId === null)
            throw new Error("The parameter 'applicationId' must be defined.");
        url_ = url_.replace("{applicationId}", encodeURIComponent("" + applicationId));
        url_ = url_.replace(/[?&]$/, "");

        const content_ = JSON.stringify(body);

            data: content_,
            method: "POST",
            url: url_,
            headers: {
                "Content-Type": "application/json-patch+json",
                "Accept": "text/plain"
            },
            cancelToken
        };

        return this.instance.request(options_).catch((_error: any) => {
            if (isAxiosError(_error) && _error.response) {
                return _error.response;
            } else {
                throw _error;
            }
        }).then((_response: AxiosResponse) => {
            return this.processAddAccessRightMetric(_response);
        });
    }

    protected processAddAccessRightMetric(response: AxiosResponse): Promise<AccessRightMetric> {
        let _headers: any = {};
        if (response.headers && typeof response.headers === "object") {
            for (let k in response.headers) {
                if (response.headers.hasOwnProperty(k)) {
                    _headers[k] = response.headers[k];
                }
            }
        }
        if (status === 200) {
            let result200: any = null;
            result200 = JSON.parse(resultData200); // Does not compile anymore with 0.23.0
            return result200;
            return throwException("An unexpected server error occurred.", status, _responseText, _headers); // Does not compile anymore with 0.23.0
        }
        return Promise.resolve<AccessRightMetric>(<any>null);
    }
```

The signature of `AxiosResponse` in axios 0.23.0:
```ts
export interface AxiosResponse<T = unknown, D = any>  {
  data: T;
  status: number;
  statusText: string;
  headers: AxiosResponseHeaders;
  config: AxiosRequestConfig<D>;
  request?: any;
}
```

In previous releases:
```ts
export interface AxiosResponse<T = any>  {
  data: T;
  status: number;
  statusText: string;
  headers: any;
  config: AxiosRequestConfig;
  request?: any;
}
```

In the current PR, I implemented Option 1, but it is maybe not the one you planned to use. Please feel free to discard the PR if it does not make sense or if you have other options.

### DIFF
--- a/src/NSwag.CodeGeneration.TypeScript/Templates/AxiosClient.liquid
+++ b/src/NSwag.CodeGeneration.TypeScript/Templates/AxiosClient.liquid
@@ -74,16 +74,16 @@
             } else {
                 throw _error;
             }
-        }).then((_response: AxiosResponse) => {
+        }).then((_response: AxiosResponse<string>) => {
 {%     if UseTransformResultMethod -%}
-            return this.transformResult(url_, _response, (_response: AxiosResponse) => this.process{{ operation.ActualOperationNameUpper }}(_response));
+            return this.transformResult(url_, _response, (_response: AxiosResponse<string>) => this.process{{ operation.ActualOperationNameUpper }}(_response));
 {%     else -%}
             return this.process{{ operation.ActualOperationNameUpper }}(_response);
 {%     endif -%}
         });
     }
 
-    protected process{{ operation.ActualOperationNameUpper }}(response: AxiosResponse): Promise<{{ operation.ResultType }}> {
+    protected process{{ operation.ActualOperationNameUpper }}(response: AxiosResponse<string>): Promise<{{ operation.ResultType }}> {
         const status = response.status;
         {% template Client.ProcessResponse %}
     }


RicoSuter:
Is this breaking with older versions of axios?

dubzzz:
@RicoSuter I introduced a regression regarding one of the execution path: https://github.com/RicoSuter/NSwag/blob/2b0d6b173d2dd13bec23dca762cd8b9906c4c464/src/NSwag.CodeGeneration.TypeScript/Templates/Client.ProcessResponse.HandleStatusCode.liquid#L27

At this point, the code says `response.data` is a `Blob` which obviously does not compile now as `string` is not a `Blob`. But the issue is that (at least in my case) I really have a `string` and not a `Blob` at this point.

Before attempting a fix, I wanted to be sure that the fact I received a `string` and not a `Blob` at this point is normal, if so I could update this line to be: `new Blob([response.data])` in the case of axios. Otherwise, if a `Blob` is really expected I can update the type of `AxiosResponse` to be `Blob` for this precise case.

RicoSuter:
> By the way, axios version 0.24.0 comes back to any. But anyway, proposed change still works with 0.24.0 and will be way safer as it fully reflects the typings we are expecting to have.

So it's better to revert this PR?

dubzzz:
It sounds better to have right typings but at the moment, it is possibly better to revert until we finf what to expect in the case described above 🤔 

Nonetheless if you know why we have been doing an explicit as, we could probably work on a fix.

dubzzz:
@RicoSuter I'd need to dig the idea of doing:
````suggestion
--- return Promise.resolve({ fileName: fileName, status: status, data: response.data as Blob, headers: _headers });
+++ return Promise.resolve({ fileName: fileName, status: status, data: new Blob([response.data]), headers: _headers });
```

Because at least with the setup of my CSharp backend I get back data as a string, so the compile error looks expected and sign of a real bug. But I'll need to double-check on my side whether the data being a string is due to a setup issue on my side or not.

RicoSuter:
Can someone create a PR with the revert/fix?

dubzzz:
@RicoSuter I did open the two options:
- revert: https://github.com/RicoSuter/NSwag/pull/3739
- fix proposal: https://github.com/RicoSuter/NSwag/pull/3740

RicoSuter:
Thanks for the PRs.

PR #3678, opened 2021-10-26
Add support for .NET 6 minimal APIs

RicoSuter:
Closes https://github.com/RicoSuter/NSwag/issues/3560

Open NSwag.Min.sln then set the project **NSwag.Sample.NET60Minimal** as start project and run (/swagger). 

Not so sure about the operation ID generator...


src/NSwag.Generation.AspNetCore/AspNetCoreOpenApiDocumentGenerator.cs
```diff
@@ -440,19 +446,23 @@ private string GetOperationId(OpenApiDocument document, ControllerActionDescript
...
                     openApiControllerAttribute.Name :
-                    actionDescriptor.ControllerName;
+                    controllerActionDescriptor.ControllerName;
 
-                operationId = controllerName + "_" + GetActionName(actionDescriptor.ActionName);
+                operationId = controllerName + "_" + GetActionName(controllerActionDescriptor.ActionName);
+            }
+            else
+            {
+                operationId = Guid.NewGuid().ToString(); // TODO: What operation ID to use for minimal APIs?
```
RicoSuter, line 465: Here we need to construct an operation ID based on the HTTP method and path/route.

RicoSuter: Fixed.

PR #3679, opened 2021-10-26
Replace DotLiquid with Fluid

lahma:
Still work in progress, uses preview feed for NJsonSchema.

* update templates to follow strict liquid syntax
* net461 minimum full framework
* netstandard 1.3 removed, netstandard 2.0 is minimum when generating code using templates


RicoSuter:
> Let me know if you would like me to add GH actions workflows to NUKE PR.

It makes sense to improve the CI and GH actions are fine for me. 
We should ensure that you can run the full build easily locally and the CI build should produce all artifacts...
What about the CD pipeline? ie pushing the nuget and npm packages and releasing the NSwagStudio MSI (maybe we use a regular GitHub release for that)?

sebastienros:
There are nupkkg files in the PR

lahma:
> It makes sense to improve the CI and GH actions are fine for me. We should ensure that you can run the full build easily locally and the CI build should produce all artifacts... What about the CD pipeline? ie pushing the nuget and npm packages and releasing the NSwagStudio MSI (maybe we use a regular GitHub release for that)?

We can do these, but can you describe how it currently works, is AppVeyor doing something all is everything manual? I saw there's `04_Publish.bat` which only does NPM publish - is it used in automation or your personal tool?

I think NUKE could do following steps in `Publish` target:

* triggered by new GH actions decription for tag happening with prefix `v` on master, e.g. `v14.5.2`
* if there's a tag and we are on main/master - continue, otherwise skip publish
* build, test, pack 
* create GitHub release with name taken from the tag, e.g. `14.5.2`  via github API (requires API key)
* upload `.msi` and `.zip` to new release
* do dotnet nuget push to nuget.org (requires NuGet API key)
* do NPM publish (requires NPM API key)

So I could make base setup for you as a PR, but you basically are only one to complete the challenge with access to keys etc. Probably will make sense to do "beta" releases.

sebastienros:
> I think NUKE could do ...

Nothing too

sebastienros:
I just lost a valuable contributor ;)

lahma:
What a emotional moment, build passes, tests are green.

PR #3692, opened 2021-11-05
Add support for GitHub artifacts

lahma:
@RicoSuter here's a version that shows how artifacts are handled (only gathers for `master` builds/non-prs).  `build.yml` was automatically generated when running NUKE as it detects the c# has changed. 

Didn't see how `NSwag.zip` is generated so just created it by zipping the studio's bin folder.

As you can see ideally everything is written inside the `Build.cs` and can then generate minimal triggering logic and rules for providers, like Appveyor. Appveyor configuration doesn't seem to be part of version control so not sure what to expect there (NUKE can also generate to config if you can export a sanitized version from Appveyor for me to see and diff).


RicoSuter:
Thanks for the PR. Will merge and then maybe improve it if necessary. 

> Appveyor configuration doesn't seem to be part of version control so not sure what to expect there

Yeah it's configured in the UI... I think we do not need this after the migration to NUKE/GH actions.

PR #3693, opened 2021-11-05
Fix CapitalizeFirst method

adesam146:


RicoSuter:
Why do we even need ToLowerInvariant?

adesam146:
> Why do we even need ToLowerInvariant?

Yes true it's not needed? Updated

adesam146:
Also any reason for the duplication of the method? (left it alone for now)

RicoSuter:
> Also any reason for the duplication of the method? (left it alone for now)

Mainly because these are two independent implementations...

I think it's fine now... Thanks for the PR.

PR #3698, opened 2021-11-08
remove extra TFM

Sinhk:
non-preview version of .NET 6 nugets got released today and when I tried to install them I can no longer build projects that depend on NSwag.AspNetCore. The reason is dependency conflicts caused by the AspNetCore dependencies having restricted ranges (Version="[5, 6)")

I tried to find out if it's necessary to limit dependencies like this, but couldn't find any reason currently. (Expect maybe to guard against future breaking changes)
I also couldn't find any breaking changes between netcoreapp3.1 and net5.0 (or net6.0) that effects NSwag, so the net5.0 TFM should not be necessary either. 

Removing net5.0 from TargetFrameworks and removing the version ranges from dependencies does not seem to break anything. Though I have not tested very thoroughly.
.NET 5 projects still starts without errors and it now works with .NET 6 projects as well.


RicoSuter:
I think with the Net6 PR merged, this is not needed anymore, right?

Sinhk:
No, not this time. 

We'll get the same problem with .NET 7 next year, but it's not really a huge problem having to wait a few days for the new NSwag release.

RicoSuter:
With this PR, wouldnt it just fall back to using 3.1?

Sinhk:
It would use the 3.1 target of this library, but a .NET 6 project would then be able to use the .NET 6 versions of the Microsoft.AspNetCore.* libraries for instance.

I'm not sure if there are any benefits to targeting the newer frameworks when not using any new features.

lahma:
Closing as resolved in master

PR #3715, opened 2021-11-12
Release v13.14.2

RicoSuter:


lahma:
I hope that at least some of the recent release hurdles are not directly caused by the NUKE migration and we can see at least some benefits at some point 😞

mikeram1984:
[Function x wallet.pdf](https://github.com/RicoSuter/NSwag/files/7796955/Function.x.wallet.pdf)

PR #3718, opened 2021-11-12
Support minimal API via CLI

RicoSuter:


src/NSwag.Sample.NET60Minimal/Program.cs
```diff
@@ -2,16 +2,7 @@
 using Microsoft.Extensions.DependencyInjection;
 using System;
 
-var builder = WebApplication.CreateBuilder(args);
-
-builder.Services.AddEndpointsApiExplorer();
-builder.Services.AddOpenApiDocument(settings =>
-{
-    settings.Title = "Minimal API";
```
thestamp, line 10: The flag on this doesn't make sense, the new Startup-less program.cs is used for both controller-based and minimal API code templates.

src/NSwag.Sample.NET60Minimal/Program.cs
```diff
@@ -20,4 +11,21 @@
...
+public partial class Program
+{
+    public static WebApplication BuildHost(string[] args)
+    {
+        var builder = WebApplication.CreateBuilder(args);
+
+        builder.Services.AddEndpointsApiExplorer();
+        builder.Services.AddOpenApiDocument(settings =>
+        {
+            settings.Title = "Minimal API";
```
thestamp, line 25: The flag on this doesn't make sense, the new Startup-less program.cs is used for both controller-based and minimal API code templates. 

I don't have a name in mind, maybe this name fix can wait for a future release when the approach settles in with the community?

PR #3729, opened 2021-11-17
Modernize and cleanup csproj files

lahma:
I seem to always battle with binding redirect problems in Fluid branch, new csproj and SDK can handle these automatically. 

* converted projects that can be converted to new SDK format
* removed some defaults like `\bin\Debug` and `bin\Release`, `TRACE,DEBUG`, `OPTIMIZE`, they are just noise in csproj
* removed `AssemblyInfo.cs` files from test projects that are now generated
* setup project only had partial copying/handling of `net60 `directory, added the missing bits


lahma:
@RicoSuter PTAL, this should help quite a lot with dependency handling / version conflicts. I was able to install the studio locally using the installer and it started as expected.

PR #3735, opened 2021-11-18
error

asadsahi:


asadsahi:
For some reason it stuck at build line and doesn't proceed with NSWAG + MSBuild:

image

lKoTl:
Maybe you are missing the nswag.json for the run command:

https://github.com/RicoSuter/NSwag/wiki/CommandLine

run: Executes an nswag.json config file (recommended)

asadsahi:
@lKoTl I think it picks up default `nswag.json` if not specified. At least that is how my existing project is setup.

After waiting for a long time, this error reveals:

image

RicoSuter:
Usually time out happens if you run the CLI in the csproj and you didnt set "noBuild" to true...

asadsahi:
Sorry, not sure where that parameter needs to go. Can you suggest the change in this PR? Thanks

RicoSuter:
https://github.com/RicoSuter/NSwag/blob/master/src/NSwag.Sample.NET60Minimal/nswag.json#L11

the current nswag.json is created to be called from outside (eg the build script) and not within the csproj.

asadsahi:
@RicoSuter thanks, updating that has eliminated timeout. 

Now I have updated PR to include controllers and some typescript code generation. This brings us to the same issue that any controller code isn't generated either in `openapi.json` or `api-client.ts`.

PR #3739, opened 2021-11-19
Revert "Make typings for `axios` more explicit on `AxiosResponse`"

dubzzz:
Reverts RicoSuter/NSwag#3667

Related to discussion: https://github.com/RicoSuter/NSwag/pull/3667#issuecomment-970043087

### DIFF
--- a/src/NSwag.CodeGeneration.TypeScript/Templates/AxiosClient.liquid
+++ b/src/NSwag.CodeGeneration.TypeScript/Templates/AxiosClient.liquid
@@ -74,16 +74,16 @@
             } else {
                 throw _error;
             }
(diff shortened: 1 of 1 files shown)


dubzzz:
@RicoSuter We should probably re-revert this one as soon as axios team tries to re-apply the typing changes (if they do)

RicoSuter:
Can't we just use any as it's just internal in the method?

dubzzz:
> Can't we just use any as it's just internal in the method so that it works with every lib version?

Indeed we can. Not using `any` but a stricter type allowed to detect potential issues but at the end, the type stays internal to nswag so `any` looks to be a good option. I can open a PR to explicitly pass it to `any`.

PR #3744, opened 2021-11-21
Trim whitespace from operation summaries

esond:
Adds a `string.Trim()` call to operation summaries to remove leading and trailing whitespace.

I encountered this behaviour when adding XML documentation to some controller methods with a format like the [following](https://github.com/RicoSuter/NSwag/compare/master...esond:esond/trim-whitespace-from-summaries?expand=1#diff-d8e7dde10714933137a5f7426213c84428b2004fac57a0d15c5bcbbb081f30faR25) [^1]:

```csharp
/// <summary>
///     This method is documented.
/// </summary>
public void DocumentedMethodWithSummary()
{
}
```

This resulted in schema output with trailing whitespace at the end of the summary like this:

```jsonc
"/someController/documentedMethodWithSummary": {
  "get": {
    "summary": "This method is documented.\n            ",
    // ...
}
```

I figured if the beginning of the summary is going to be trimmed, the end probably should be trimmed as well. This is just a simple call to `string.Trim()`, so whitespace inside the summary itself will be maintained.

I also included the trimming behaviour on summaries set with `OpenApiOperationAttribute` and `DescriptionAttribute` as no downsides to this were immediately apparent to me.

Thank you for considering my contribution and for the great tool. NSwag has been perennially useful to me and teams I have been on.

[^1]: For what it's worth, this is the default way JetBrains tools (ReSharper, Rider) format XML doc.

### DIFF
--- a/src/NSwag.Generation.Tests/Processors/OperationSummaryAndDescriptionProcessorTests.cs
+++ b/src/NSwag.Generation.Tests/Processors/OperationSummaryAndDescriptionProcessorTests.cs
@@ -1,0 +1,96 @@
+﻿using System;
+using System.ComponentModel;
+using System.Reflection;
+using NSwag.Annotations;
+using NSwag.Generation.Processors;
+using NSwag.Generation.Processors.Contexts;
+using Xunit;
+
+namespace NSwag.Generation.Tests.Processors
+{
+    public class OperationSummaryAndDescriptionProcessorTests
+    {
+        public class DocumentedController
+        {
+            [OpenApiOperation("\r\n\t This method is documented. \r\n\t", "")]
+            public void DocumentedMethodWithOpenApiOperationAttribute()
+            {
+            }
+            
+            [Description("\r\n\t This method is documented. \r\n\t")]
+            public void DocumentedMethodWithDescriptionAttribute()
+            {
+            }
+
+            /// <summary>
+            ///     This method is documented.
+            /// </summary>
+            public void DocumentedMethodWithSummary()
+            {
+            }
+        }
+        
+        [Fact]
+        public void Process_TrimsWhitespaceFromOpenApiOperationSummary()
+        {
+            //// Arrange
+            var controllerType = typeof(DocumentedController);
+            var methodInfo = controllerType.GetMethod(nameof(DocumentedController.DocumentedMethodWithOpenApiOperationAttribute));
+
+            var context = GetContext(controllerType, methodInfo);
+            var processor = new OperationSummaryAndDescriptionProcessor();
+
+            //// Act
+            processor.Process(context);
+
+            //// Assert
+            var summary = context.OperationDescription.Operation.Summary;
+            Assert.Equal("This method is documented.", summary);
+        }
+        
+        [Fact]
+        public void Process_TrimsWhitespaceFromDescription()
+        {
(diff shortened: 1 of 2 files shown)


esond:
As I move back to the project where I observed this behavior, I'm noticing same thing happens for other elements as well, such as descriptions on parameters as well as responses. For example:

```csharp
/// <param name="some param">
///     Some long description that I want to line
///     wrap.
/// </param>
```
Will result in a `parameter` in the schema with a description like:

```jsonc
"description": "Some long description that I want to line\nwrap\n            "
```

Happy to proceed with fixes for those and other examples if I am on the right track here. Will hold on advice about my approach here before proceeding.

RicoSuter:
> I am also wondering if we care about doing anything about the line wrapping in the description, perhaps replacing a newline with a space or something to get a more pleasantly formatted description value? That could be overkill.

we should not do that as it might be markdown and the breaks are needed.

RicoSuter:
I wonder whether it's better to fix this directly in https://github.com/RicoSuter/Namotion.Reflection
(XmlDocs reader)

esond:
@RicoSuter Ten-four on the interior line-wrapping.

After my own glance into that code, it would seem reasonable to solve this in Namotion.Reflection instead. Happy to take this on if you are confident it's the right way to go.

RicoSuter:
@esond I cannot think of a scenario where this trimming would be a problem. We should only do it in the string-returning xmldocs method, this way a consumer can still use the xml node based one if he wants to read the "raw value".

esond:
@RicoSuter Ten-four, thanks for the advice.

Closing this PR. I will reference it in the subsequent PR to Namotion.Reflection. I should be able to get to this on the weekend.

esond:
@RicoSuter So, upon attempting to implement this trimming behaviour in Namotion.Reflection, I've arrived at the opinion that NSwag may be the correct place to implement this change for the sake of maintaining consistency in the reflection library behaviours.

Adding the trimming behaviours and tests in Namotion.Reflection for `<summary>` and `<remarks>` tags was straightforward, but when I moved onto trimming for generic tags, things became less clear. When a tag starts with indentation, the underlying `XElement` value contains the whitespace. Because of this, I hesitated to move forward as achieving a consistent behaviour across all tag types would have required changing the way `XElement.Value` gets handled for documentation generation across the board. I have a failing test for this [here](https://github.com/esond/Namotion.Reflection/blob/35e3a85696acae02922cc25edd26659cdfe3740f/src/Namotion.Reflection.Tests/XmlDocsExtensionsTests.cs#L178) to illustrate where this all started. I toyed around with a few solutions to this, but they felt a bit risky or too much like they could break things. I'll put it another way: if this code was an internal product of a team I was on, I'd sooner suggest we just change our XML doc style to not be indented than make a fundamental change to the XML doc reader/generator itself.

With all that said, it could still make sense for the purposes of generating OpenAPI `description` properties to trim these values, since they almost certainly shouldn't have leading or trailing whitespace in any circumstance.

RicoSuter:
So your proposal is to merge this PR?

esond:
Yes, that is my suggestion. I believe NSwag this is the most appropriate place to make this change.

RicoSuter:
Thx for the PR.

PR #3747, opened 2021-11-23
Prepare build publish pipeline

lahma:
* fix some warnings
* use common Directory.Build.props to remove repetition
* harmonize unit test libraries to single and latest version available
* catch tags and master/main build to do a publish like in NJsonSchema (requires removal of this branch before merge)

- [x] NSwag MyGet feed
    - https://www.myget.org/feed/Packages/nswag

Minimum requirements: 

- [x] Would be nice to patch the version in NSwag.NPM/package.json with the version from the .props file before build/publish
- [x] Publish nuget packages to nuget.org
- [x] Publish NSwagStudio.nupkg package to chocolatey.org (different nuget feed)
    - NuGet feed URL: https://push.chocolatey.org/ 
    - CHOCO_API_KEY
- [x] Publish NPM package


RicoSuter:
Minimum requirements: 

- [ ] Publish nuget packages to nuget.org
- [ ] Publish NSwagStudio.nupkg package to chocolatey.org (different nuget feed)
- [ ] Publish NPM package

src/Directory.Build.props
```diff
@@ -33,6 +33,9 @@
   <PropertyGroup Condition="$(MSBuildProjectName.Contains('.Test')) == true or $(MSBuildProjectName.Contains('.Sample')) == true">
     <IsPackable>false</IsPackable>
   </PropertyGroup>
+  <PropertyGroup Condition="$(MSBuildProjectName.Contains('.Test')) != true and $(MSBuildProjectName.Contains('.Sample')) != true">
+    <GeneratePackageOnBuild>True</GeneratePackageOnBuild>
+  </PropertyGroup>
```
lahma, line 38: Do we need this for every IDE build or does the `build.cmd pack` suffice?

RicoSuter: i think we should only do it with pack, i agree...
but i think the `.Where(x => x.ToString().Contains("NSwagStudio"))` is filtering out all projects which should be packed...

RicoSuter: what about adding a custom property to the projects which should be packed and then search them by this property?

lahma: The problem was with loading projects, before getting the info it crashed trying to load wix reference. Generally IsPackable=false would be ideal, or vice versa not having anything packable and then allowing with property per project

RicoSuter: changed.

Issue #2704, opened 2020-02-26
Is there a way to make client methods virtual?


jonlin:
I'm running into issues with Mocking the client because the methods aren't virtual and the generated interface isn't working for me because we have a partial class that handles authentication and disposing of the custom HttpClient, and the generated interface doesn't extend IDisposable.

Is there an option or any other way to make the generated client's methods virtual so it can be mocked?

RicoSuter:
Currently there is no option for that specifically.. you need to override the template and add the virtual modifier. The question is whether we should add this by default or if we should add an option. But maybe you can change the access modifier from “public” to “public virtual”?

jonlin:
I tried changing the "typeAccessModifier" config to "public virtual" but I got an error when generating that said: the modifier 'virtual' is not valid for this item.

It looks like "virtual" is getting added to the POCO objects and not the client methods themselves.

How do I override the template? Is that hard?

RicoSuter:
https://github.com/RicoSuter/NSwag/blob/master/src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.liquid#L92

https://github.com/RicoSuter/NSwag/wiki/Templates

TwoUnderscorez:
@RicoSuter Any updates?
Could you please provide a more detailed explanation?
Also, how does one provide a custom template?

Your help and effort are much appreciated.

jeremyVignelles:
I think that what he meant was this line from the wiki:

> To override templates, create a template directory, put the overridden templates into it and point to it with the TemplateDirectory setting.

Then, you can take the Client.Class.liquid from the repo, and add `virtual` after `{{ operation.MethodAccessModifier }}` at the line he pointed.

TwoUnderscorez:
Ah, thank you!
I didn't realize that that was the name of the setting in the nswag.json. (`templateDirectory`)
I think this issue can be closed now.

PR #3751, opened 2021-11-24
sdk methods generated as virtual methods

rw-ben:
@RicoSuter  This updates the Client.Class.liquid template to make SDK methods generated as virtual methods.

How it could be used from class that inherits generated sdk class:
```
        public override Task<SwaggerResponse<CompanyFeature>> GetCompanyFeatureByCompanyIdAsync(int companyId, CancellationToken cancellationToken = default)
        {
          //some code needed to extend goes here (ex. maybe some small cache code)
        }
```

Original output:
```
        public async System.Threading.Tasks.Task<SwaggerResponse<CompanyFeature>> GetCompanyFeatureByCompanyIdAsync(int companyId, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken))
```

New output:
```
        public virtual async System.Threading.Tasks.Task<SwaggerResponse<CompanyFeature>> GetCompanyFeatureByCompanyIdAsync(int companyId, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken))
```


rw-ben:
@jonlin noticed issue 2704 you previously posted

rw-ben:
@RicoSuter added VirtualMethods parameter options in C# generator and made it optional + default to false

rw-ben:
@RicoSuter 

I noticed your comments on issue #2704 "Currently there is no option for that specifically.. you need to override the template and add the virtual modifier. The question is whether we should add this by default or if we should add an option. But maybe you can change the access modifier from “public” to “public virtual”?"

I was hoping to see if there were any concerns with this feature addition in this pull request. It defaults to existing functionality and has an option to enable the access modifier to be "public virtual" instead of "public".

Again appreciate your time.

rw-ben:
@RicoSuter Would you kindly approve the workflow? I'm a first-time contributor.

RicoSuter:
I wonder whether we really need to add a new setting for that (more complexity) or just add virtual to the template - performance wise these methods are "slow" anyway as they are I/O based and adding virtual is not a breaking change (right?). Opinions? /cc @lahma @jeremyVignelles

lahma:
I don't see any harm for them being virtual by default. If the API would change it would anyway break the caller and also then the inheritor/override.

jeremyVignelles:
OK for virtual by default. You know my opinion about adding an option for each use case :)

RicoSuter:
@rw-ben ok I agree, let's remove the setting and just add virtual to the templates - then we have a quite simple PR.

rw-ben:
@RicoSuter @jeremyVignelles @lahma thank you all for taking a look at this. I will make it default and push the changes up.

rw-ben:
@RicoSuter @jeremyVignelles @lahma removed the virtual method C# generation param and it always generate virtual methods

rw-ben:
@RicoSuter thanks

RicoSuter:
Thanks for the PR.

rw-ben:
@RicoSuter How do I get a new release built? Is this something you have to kick off?

jkone27:
what about C# properties? they are not virtual is it possible to ovverride them too? (e.g. to give defaults to null/empty)

PR #3757, opened 2021-11-25
Improve publish diagnostics

lahma:
* print determined setup
* version prefix from tag and fall back to file
* don't include satellite resources (MSBuild NuGet now 12MB smaller)
* push to MyGet with try-catch
* now that `VersionSuffix ` is calculated at the beginning it shouldn't change the outcome even if the build takes more than a minute..


RicoSuter:
lgtm, you can merge if complete.

PR #3762, opened 2021-11-27
Use correct Microsoft.CodeAnalysis for target framework

trejjam:
Hi, I did not find the reason why the `Microsoft.CodeAnalysis` did appear in the [code base](https://github.com/RicoSuter/NSwag/commit/9ca4f4c572fbf988881f86e58ffd8ad19daa276f), but its version is tied with the target framework; [see Roslyn Wiki](https://github.com/dotnet/roslyn/blob/main/docs/wiki/NuGet-packages.md).
E.g. in net5.0 only environment it causes restore/build issues.

So I did bind `Microsoft.CodeAnalysis` to a specific targetFramework.

Maybe it is enough just to use `PrivateAssets="all"`. But that depends on why it is in the code base :) .

### DIFF
--- a/src/NSwag.Generation.AspNetCore/NSwag.Generation.AspNetCore.csproj
+++ b/src/NSwag.Generation.AspNetCore/NSwag.Generation.AspNetCore.csproj
@@ -7,11 +7,11 @@
   </PropertyGroup>
 
   <ItemGroup>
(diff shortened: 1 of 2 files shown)


RicoSuter:
Cannot remember why it's added, maybe it's better just to remove it.

trejjam:
Removed

RicoSuter:
Thanks for the Pr.

PR #3777, opened 2021-12-04
Get build running on Linux and MacOS

### LINKED ISSUE #4615: Can't restore and build on Linux
When I run build.sh on Linux (Ubuntu 23.10, .NET 8.0.100), the project fails to restore and build.

This is actually two issues:

1. Nuke fails to restore due to being unable to find `msbuild` in the path or relevant environment variables. On Linux, msbuild is supposed to be invoked by `dotnet build`. Adding a shell script named /usr/local/bin/msbuild that simply invokes `dotnet build` solved this issue for me. This is probably an issue with Nuke, I guess.
2. After getting past that, NSwag still fails to restore due to NSwagStudio being part of the solution. NSwagStudio requires the WindowsDesktop SDK, which isn't available on Linux. It would be nice to split the solution into two separate solutions, one for NSwag proper and another one for NSwagStudio.

By removing every reference to NSwagStudio, NSwagStudio.Installer, and NSwagStudio.Chocolatey I was able to build, although more work remains to be done in order to build with Nuke.

lahma:
Now can restore and compile solution on both macOS and Linux with NUKE, required some checks whether on Windows and skip some parts, also case-sensitiveness bite a bit too. Now publishing artifacts for PR too if people want to check how they work.

fixes #4615

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -36,7 +36,7 @@ nswag.js text eol=lf
 *.sass text
 *.scm text
 *.scss text
-*.sh text
+*.sh text eol=lf
 *.sql text
 *.styl text
 *.tpl text
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -68,3 +68,27 @@ jobs:
         with:
           name: NuGet Packages
           path: artifacts/*.nupkg
+  ubuntu-latest:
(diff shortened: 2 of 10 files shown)


RicoSuter:
Thanks for the PR.

Issue #3575, opened 2021-08-01
Typescript Generator Generating  return _observableOf(null); -- failing typescript build


mbrennan376:
ASP.NET Core 3.1
NSwag 13.12.1
Angular 11
Typescript 4.1
Rxjs 6.6

I am generating typescript only via MSBuild referring to the project file

All of my generated process methods end with
return _observableOf(null);

`     

        let _headers: any = {}; if (response.headers) { for (let key of response.headers.keys()) { _headers[key] = response.headers.get(key); }}
        let _mappings: { source: any, target: any }[] = [];
        if (status === 200) {
            return blobToText(responseBlob).pipe(_observableMergeMap(_responseText => {
            let result200: any = null;
            let resultData200 = _responseText === "" ? null : jsonParse(_responseText, this.jsonParseReviver);
            result200 = Foo.fromJS(resultData200, _mappings);
            return _observableOf(result200);
            }));
        } else if (status === 404) {
            return blobToText(responseBlob).pipe(_observableMergeMap(_responseText => {
            let result404: any = null;
            let resultData404 = _responseText === "" ? null : jsonParse(_responseText, this.jsonParseReviver);
            result404 = ProblemDetails.fromJS(resultData404, _mappings);
            return throwException("A server side error occurred.", status, _responseText, _headers, result404);
            }));
        } else if (status === 500) {
            return blobToText(responseBlob).pipe(_observableMergeMap(_responseText => {
            return throwException("A server side error occurred.", status, _responseText, _headers);
            }));
        } else if (status !== 200 && status !== 204) {
            return blobToText(responseBlob).pipe(_observableMergeMap(_responseText => {
            return throwException("An unexpected server error occurred.", status, _responseText, _headers);
            }));
        }
        return _observableOf(null);     **<--- THIS LINE HERE**
    }`

so my ng build fails

An example of how I defined my API method:

 `[ProducesResponseType(typeof(Foo), StatusCodes.Status200OK)]
        [ProducesResponseType(StatusCodes.Status404NotFound)]
        [ProducesResponseType(StatusCodes.Status500InternalServerError)]
        [HttpGet("{ID}")]
        public async Task<ActionResult<Foo>> GetFooAsync(int ID)
       {
            Do STUFF
        }`

IF I edit my generated file to be  return _observableOf(new Foo); 
ng build works. I would be happy to throw an exception if it gets to this part of the code, as the path will likely never be hit.

[...]

PR #3780, opened 2021-12-06
RxJS Version 7 causes problems on Angular Client

florianraffl:
We had a problem as mentioned by my colleague @Thomas-Lederer in the issue #3575. 
The problem was that if we try to generate an Angular Client with Nswag the process#ControllerMethodName# trys to return `_observableOf<ResultType>(<any>null);` which is not possible in RxJS 7. 
I changed that in condition to the config parameter `rxJsVersion` to `_observableOf(null);` which should be the correct syntax for RxJS 7.

I would be happy if we can get this change as soon as possible in the next release so that I can update to the newest version :)


RicoSuter:
Is this the only thing which is broken with RxJS 7?

florianraffl:
In our case, how we use NSwag, this the only thing we know, that is broken with RxJS7.

florianraffl:
Nice Thanks! :)

Issue #3794, opened 2021-12-15
Controllers not included in specification when using minimalAPI


Bleialf:
I'm trying to setup NSwagStudio (v13.15.4.0) with a net6 webapplication using the minimal API template.
It seems that the Legacy controllers are not included in the specification. I also tried the NSwag.Sample.NET60Minimal which shows the same behavior (doesn't map the controller included in the program.cs file). 

Is this behavior expected at this point or am I missing something? I want to force myself to use the new template but I still want to use the legacy Controllers.

jaumemilian:
We are facing the same issue trying to generate clients using MsBuild
But also checked that with the Npm Tool it neither works properly

RicoSuter:
It seems that the controllers are only added to the pipeline and API explorer when the middlewares are executed. However the CLI does not execute the middlewares but only the service registrations and thus they do not show up. I do not know whether this is a bug in ASP.NET or something NSwag can fix... does someone know whether this works with Swashbuckle and its CLI?

dylanvdmerwe:
Confirming that using a minimal API in net6, running nswag from command line generates a blank TS file. No controllers seem to be generated.

RicoSuter:
If you run the app and check Swagger UI and the spec is generated via the HTTP pipeline then controllers are picked up, so i conclude that controllers picked up by API explorer in the HTTP middlewares/pipeline and not in the services.

dylanvdmerwe:
If you browse to the swagger endpoint when running the site, all controllers and methods show up correctly. There must be a way for nswag to pick up the API explorer from the HTTP middleware?

RicoSuter:
> If you browse to the swagger endpoint when running the site, all controllers and methods show up correctly

This is also generated by NSwag (so it picks up in this case).

> nswag to pick up the API explorer from the HTTP middleware

The problem is that the CLI does not really start the app but only runs the DI setup (ConfigureServices) and somehow in this scenario the controllers are not added to API Explorer... no idea how to fix that.

dylanvdmerwe:
So are we saying that new .NET 6 projects (or migrated ones) that move to the recommended minimal API for Program.cs is not supported by NSwag?

RicoSuter:
> So are we saying that new .NET 6 projects (or migrated ones) that move to the minimal API for Program.cs is not supported by NSwag?

[...]

PR #3814, opened 2022-01-01
Net6.0 Minimal API projects without Startup, issue #3794

bsurprised:
Fixes missing controllers in the specification when generating Net6.0 Minimal API projects using CLI/MSBuild

### DIFF
--- a/src/NSwag.Commands/HostApplication.cs
+++ b/src/NSwag.Commands/HostApplication.cs
@@ -140,7 +140,13 @@ namespace NSwag.Commands
             try
             {
                 // Get the IServiceProvider from the host
+#if NET6_0
+                var assemblyName = assembly.GetName()?.FullName ?? string.Empty;
+                // We should set the application name to the startup assembly to avoid falling back to the entry assembly.
+                var services = ((IHost)factory(new[] { $"--{HostDefaults.ApplicationKey}={assemblyName}" })).Services;
+#else
                 var services = ((IHost)factory(Array.Empty<string>())).Services;
+#endif
 
                 // Wait for the application to start so that we know it's fully configured. This is important because
                 // we need the middleware pipeline to be configured before we access the ISwaggerProvider in


src/NSwag.Commands/HostApplication.cs
```diff
@@ -140,7 +140,13 @@ void OnEntryPointExit(Exception exception)
             try
             {
                 // Get the IServiceProvider from the host
+#if NET6_0
```
maxkoshevoi, line 143: `NET6_0_OR_GEATER` or else it will be broken in NET 7 again

bsurprised: Good idea. 
From my understanding, NSwag style has been code tailored to each dotnet version and I made it NET6 specific. DotNet has had many breaking changes in recent versions, so I do hope this works for NET7 as well. :-)

vpetrusevici:
Hello @RicoSuter 
Can you say please when this fix will be released?

lahma:
@vpetrusevici There's now a new release on NuGet

vip32:
just tried it, works fine :) However when also setting up API versioning (Microsoft.AspNetCore.Mvc.Versioning.ApiExplorer) with the following setup in Program.cs. Note that the versioning is not really used in the minimal endpoints or controller.

`builder.Services.AddApiVersioning(options => ConfigureApiVersioning(options));
builder.Services.AddVersionedApiExplorer(options => options.SubstituteApiVersionInUrl = true);`

the minimal endpoints are *not* present in the swagger, the controller operations are present.

====================================================================
this is the whole sample (NSwag.Sample.NET60Minimal) is used:

using Microsoft.AspNetCore.Builder;
using Microsoft.AspNetCore.Http;
using Microsoft.AspNetCore.Mvc;
using Microsoft.AspNetCore.Mvc.Versioning;
using Microsoft.Extensions.DependencyInjection;
using System;

var builder = WebApplication.CreateBuilder(args);

// Optional: Use controllers
builder.Services.AddControllers();

builder.Services.AddApiVersioning(options => ConfigureApiVersioning(options));
builder.Services.AddVersionedApiExplorer(options => options.SubstituteApiVersionInUrl = true);

builder.Services.AddEndpointsApiExplorer();
builder.Services.AddOpenApiDocument(settings =>
{
    settings.Title = "Minimal API";
    settings.Version = "v1";
});

var app = builder.Build();
app.UseDeveloperExceptionPage();

app.UseOpenApi();
app.UseSwaggerUi3();

app.MapGet("/", (Func<string>)(() => "Hello World!"))
    .WithTags("General123");

app.MapGet("/sum/{a}/{b}", (Func<int, int, int>)((a, b) => a + b))
    .WithName("CalculateSum")
    .WithTags("Calculator");

// Optional: Use controllers
app.UseRouting();
app.MapControllers();

app.Run();

void ConfigureApiVersioning(ApiVersioningOptions options)
{
    options.DefaultApiVersion = new ApiVersion(1, 0);
    options.AssumeDefaultVersionWhenUnspecified = true;
    options.ReportApiVersions = true;
}

// Optional: Use controllers
[ApiController]
[Route("api/examples")]
public class ExampleController : ControllerBase
{
	[HttpGet]
	public IActionResult Get()
	{
		return Ok("Examples");
	}
}

RicoSuter:
> However when also setting up API versioning (Microsoft.AspNetCore.Mvc.Versioning.ApiExplorer) with the following line in Program.cs. Note that the versioning is not really used in the minimal endpoints or controller.

Seems that for some reason the minimal operations are not reported anymore by API Explorer... needs investigating.

RicoSuter:
Is this still a problem with v13.15.10?

vip32:
yes, just tried again with 13.15.10. `AddVersionedApiExplorer` still causes minimal endpoints to be *not* present in the generated swagger json.

PR #3816, opened 2022-01-04
FileResponse: support for special characters in fileName  (+ suppress warning when nullable is enabled in C# clients) 

knoepdan:
**Warning suppressed**
When the Nullable feature is enabled and Warning are treated as errors, the generated C# client will not compile due to the fact, that deserialization might fail and newtonsoft will the return null. Therefore, this pull request will suppress the corresponding warning. 

Example code snippet that might cause an error depending on the C# projec settings: 
`      
  public static Path FromJson(string data)
        {
            return Newtonsoft.Json.JsonConvert.DeserializeObject<Path>(data, new Newtonsoft.Json.JsonSerializerSettings());
        }`

Remark: adding the ? to mark the return value as nullable would have been possible as well. However, this might not be compatible with older C# versions, so i guess its safer to just suppress the error.

### DIFF
--- a/src/NSwag.CodeGeneration.CSharp/Templates/File.liquid
+++ b/src/NSwag.CodeGeneration.CSharp/Templates/File.liquid
@@ -19,6 +19,7 @@ using {{ usage }};
 #pragma warning disable 1591 // Disable "CS1591 Missing XML comment for publicly visible type or member ..."
 #pragma warning disable 8073 // Disable "CS8073 The result of the expression is always 'false' since a value of type 'T' is never equal to 'null' of type 'T?'"
 #pragma warning disable 3016 // Disable "CS3016 Arrays as attribute arguments is not CLS-compliant"
+#pragma warning disable 8603 // Disable "CS8603 Possible null reference return"
 
 namespace {{ Namespace }}
 {
@@ -215,3 +216,4 @@ namespace {{ Namespace }}
 #pragma warning restore  114
 #pragma warning restore  108
 #pragma warning restore 3016
+#pragma warning restore 8603
--- a/src/NSwag.CodeGeneration.TypeScript/Templates/Client.ProcessResponse.HandleStatusCode.liquid
+++ b/src/NSwag.CodeGeneration.TypeScript/Templates/Client.ProcessResponse.HandleStatusCode.liquid
@@ -6,8 +6,8 @@ const contentDisposition = response.headers ? response.headers["content-disposit
 {%     else -%}
 const contentDisposition = response.headers ? response.headers.get("content-disposition") : undefined;
 {%     endif -%}
-const fileNameMatch = contentDisposition ? /filename="?([^"]*?)"?(;|$)/g.exec(contentDisposition) : undefined;
-const fileName = fileNameMatch && fileNameMatch.length > 1 ? fileNameMatch[1] : undefined;
+const fileNameMatch = contentDisposition ? /filename[^;=\n]*=(?:(\\?['"])(.*?)\1|(?:[^\s]+'.*?')?([^;\n]*))/g.exec(contentDisposition) : undefined;
+const fileName = fileNameMatch && fileNameMatch.length > 1 ? decodeURIComponent(fileNameMatch[fileNameMatch.length - 1]) : undefined;
 {%     if operation.WrapResponse -%}
 {%         if Framework.IsAngular -%}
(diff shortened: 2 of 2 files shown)


knoepdan:
:-(  oh no.. i didnt mean the second commit to be part of this pull request. That should have been a separate one.   (It's a long long time since i touched GitHub again.. nevertheless, i should have known). 

This commit fixes an issue with the fileResponse when the fileName contains special characters (such as Umlaute).


To support special characters in a filename in asp.net, one can use the ContentDisposition.FileNameStar property (instead of FileName). Example in the C# code below (which shows usage of both properties)

`    
            var response = new HttpResponseMessage(HttpStatusCode.OK);
            byte[] contents = GetFileResponse(out string filename);

            response.Content = new ByteArrayContent(contents);
            response.Content.Headers.ContentDisposition = new ContentDispositionHeaderValue("attachment");

            // fileresponse without support for special characters (e.g.: Umlaute) 
            //response.Content.Headers.ContentDisposition.FileName = filename; // filename="=?utf-8?B?w6TDpMOkMjAyMjAxMDZfMTc0ODUzLkNTVg==?="  (incorrect with special characters)

            // or with support for special characters (e.g. Umlaute)
            response.Content.Headers.ContentDisposition.FileNameStar = filename; // filename*=utf-8''%C3%A4%C3%A4%C3%A420220106_174853.CSV

            response.Content.Headers.ContentType = new MediaTypeHeaderValue("application/octet-stream");      
`

knoepdan:
Actually, i meant to do 2 commits, one for the warning and one for the fileResponse enhancement.     (It's a long long time since i touched GitHub again.. nevertheless, i should have known.). I hope its still ok like this since one commit is just about a warning.

RicoSuter:
Lgtm, thanks for the PR!

KostaMadorsky:
Looks like this PR broke working parsing in my case: #3865

RicoSuter:
@KostaMadorsky @knoepdan Can you provide a fix for this so that it's not introducing a regression and also supports this new scenario? Otherwise I have to rollback this change.

KostaMadorsky:
@RicoSuter I would not be able to push a well-tested fix for this any time soon, unfortunately.
I think the proper version should come with a bunch of unit tests (the new regex does not look straightforward).

For now, something like that would do (the regex is the same as in the new version):

```
const contentDisposition = `attachment; filename="Certificate - Test questions - 2022-02-07_00-10.pdf"; filename*=UTF-8''Certificate%20-%20Test%20questions%20-%202022-02-07_00-10.pdf`;
const fileNameRgx = /filename[^;=\n]*=(?:(\\?['"])(.*?)\1|(?:[^\s]+'.*?')?([^;\n]*))/g;

let fileName;
let matchArray;

while ((matchArray = fileNameRgx.exec(contentDisposition)) != null) {
  if (matchArray.length < 4) {
    continue;
  }

  fileName = matchArray[3] || matchArray[2] || fileName;
  console.log(fileName);
}

if (fileName) {
  fileName = decodeURIComponent(fileName);
}

console.log('final: ', fileName);
```

image


https://onecompiler.com/javascript/3xspf4sbx
image

knoepdan:
Sorry for the problems. That should not have happened.  (ohh i actually hate Regex, so hard to understand sometimes)
I also cannot provide a well tested fix so quickly. And yes unit tests would be nice. 

However, what I propose as a quick fix that certainly works is the following: 


`let fileNameMatch = contentDisposition ? /filename="?([^"]*?)"?(;|$)/g.exec(contentDisposition) : undefined;
let fileName = fileNameMatch && fileNameMatch.length > 1 ? fileNameMatch[1] : undefined;
if(!fileName){
	fileNameMatch = contentDisposition ? /filename[^;=\n]*=(?:(\\?['"])(.*?)\1|(?:[^\s]+'.*?')?([^;\n]*))/g.exec(contentDisposition) : undefined;
	fileName = fileNameMatch && fileNameMatch.length > 1 ? decodeURIComponent(fileNameMatch[fileNameMatch.length - 1]) : undefined;
}`

Basically, i use the old Regex which has been well tested in the field, and if it returns undefined, we try with the new one. 
However, i don't think this should be the final solution. (and even if it was, there should be unit tests etc.). Will still create a pull request for this as I think its a) better than the current solution and b) better than the previous state.

KostaMadorsky:
@knoepdan it will not use UTF8 filename if there is non-UTF8  filename in the string with your solution.
Could you try mine and see if it works?
Tried 4 different scenarios on the link above.

KostaMadorsky:
@knoepdan your regex always returns 4 array elements if matched: whole match string and 3 capturing groups.
So then you need to run it until there are matches (while loop in my code) and try to pull values from the 3rd capturing group
If no UTF8 filename, it will fall back to the second capturing group, which works for non-UTF8 filename.

RicoSuter:
Reverted this PR => v13.15.8
Please find a solution which works for everyone in the other PR...

PR #3822, opened 2022-01-10
Add ability to add a custom exception handler

johntyrrell:
In our solutions we already have existing exceptions that are handled in our front-ends.
These exceptions are custom and are now used when communicating with our legacy WCF services. 
When generating a WebAPI client with NSwag there is the option to re-use a single exception, if it has the correct constructor.

With this PR, I have added the option to specify an ExceptionFactory. If you do not specify a custom factory, the default is used:
```
    private static Exception DefaultExceptionFactory({{ExceptionClass}} ex)
    {
        return ex;
    }
```

But now we can specify a custom ExceptionFactory, for example: "My.Custom.NameSpace.MyCustomExceptionFactory":
```
public static Exception MyCustomExceptionFactory(ApiException ex)
{
  return ex.StatusCode switch
  {
    400 => ex.Result is ValidationProblemDetails ? new ValidationException(ex.Result.Errors) : new FunctionalException(ex.Result.Message)
    401 => new UnauthorizedException();
    403 => new ForbiddenException();
    _ => ex;
  };
}
```

Without the ExceptionFactory, we would need to catch the ApiException everywhere the client is used and convert it afterwards.

### DIFF
--- a/src/NSwag.CodeGeneration.CSharp.Tests/CSharpClientSettingsTests.cs
+++ b/src/NSwag.CodeGeneration.CSharp.Tests/CSharpClientSettingsTests.cs
@@ -150,5 +150,25 @@ namespace NSwag.CodeGeneration.CSharp.Tests
             // Assert
             Assert.Contains("public partial interface IFooClient : IClientBase", code);
         }
+
+        [Fact]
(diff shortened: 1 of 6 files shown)


buti1021:
This would be really helpful!

PR #3835, opened 2022-01-17
Improve processing performance

lahma:
* optimize couple hot paths where LINQ was allocating and making things slow
* only spawn Task.Run where forking makes sense (multiple generators), spawning of Task didn't produce performance benefit and only made profiling harder due the unnecessary extra thread, now spawning for generator handling.

Running against our code base and generating TS and C# clients from Swagger2.0 specification

master:        `Duration: 00:00:06.2780536`
this branch: `Duration: 00:00:05.6278623`

### DIFF
--- a/src/NSwag.Commands/Commands/CodeGeneration/OpenApiToCSharpClientCommand.cs
+++ b/src/NSwag.Commands/Commands/CodeGeneration/OpenApiToCSharpClientCommand.cs
@@ -169,7 +169,7 @@ namespace NSwag.Commands.CodeGeneration
             get { return Settings.ExposeJsonSerializerSettings; }
             set { Settings.ExposeJsonSerializerSettings = value; }
         }
-        
+
         [Argument(Name = "ClientClassAccessModifier", IsRequired = false, Description = "The client class access modifier (default: public).")]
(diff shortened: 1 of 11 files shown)


RicoSuter:
> only spawn Task.Run where forking makes sense (multiple generators), spawning of Task didn't produce performance benefit and only made profiling harder due the unnecessary extra thread, now spawning for generator handling.

I think the main reason for this was to avoid blocking in the WPF UI thread...

lahma:
If it's async it should already achieve that? Or should the UI spawn a top level helper task?

RicoSuter:
> If it's async it should already achieve that? Or should the UI spawn a top level helper task?

Just because a method is async doesnt mean that everything is running async, ie maybe the first part of the method is sync and only then it's calling await, with run() around all the whole method is async... however i think it's fine to merge as this was actually used at the time when the commands were called directly from the UI - now they are called via process, so doesnt really matter anyway...

RicoSuter:
Thanks again for the PR!

lahma:
Thanks for review and merge! Next release will again be faster when this and NJsonSchema improvements are added together 🚀 😉

PR #3842, opened 2022-01-21
Map blob type from response content type

sandves:
This is my first PR, so please help me out if I am not following your process.

#3740 actually turned out to be a breaking change for me.

Before this change, the blob would have the correct `type` set according to the `Content-Type` on the response. By initializing a new `Blob` without specifying the type, the `type` property is empty, resulting in potential bugs when processing the blob later.

I just made changes to the code where the `Framework.IsAxios` condition is met. This is because I am using `Axios` with the Typescript client, and I do not know if this could be handled in the same way for Angular JS.


RicoSuter:
Lgtm, thanks for the PR.

PR #3852, opened 2022-01-29
- extend settings

RosData:
`ObjectInQueryRenderOnlyProperties` used in `Client.Class.QueryParameter`

We need to render query parameters in url in different way, so I introduced new bool parameter: `ObjectInQueryRenderOnlyProperties`

```liquid
 {% if ObjectInQueryRenderOnlyProperties -%}  
        urlBuilder_.Append(System.Uri.EscapeDataString("{{property.Key}}") + "=").Append(System.Uri.EscapeDataString(ConvertToString({{parameter.Name}}.{{property.Name}}, System.Globalization.CultureInfo.InvariantCulture))).Append("&");  
{% else -%}  
        urlBuilder_.Append(System.Uri.EscapeDataString("{{parameter.Name}}[{{property.Key}}]") + "=").Append(System.Uri.EscapeDataString(ConvertToString({{parameter.Name}}.{{property.Name}}, System.Globalization.CultureInfo.InvariantCulture))).Append("&");
{% endif -%}
```


trejjam:
Hi, I can recommend fixing the original swagger.json (using processors), instead of introducing something similar.

RosData:
> Hi, I can recommend fixing the original swagger.json (using processors), instead of introducing something similar.

Maybe I don't quite understand. There is third party API exposing yaml. In yaml there is object Pageable (specified in Query) with properties page and Size.

RosData:
> Hi, I can recommend fixing the original swagger.json (using processors), instead of introducing something similar.

The api I generate client for exposes Yaml and comes from other company's department.  But you are right, after deeper study of the template "Client.Class.QueryParameter.liquid" I found  parameter combination that must be added to the query object in yaml: style: form and explode: true must be added, so final looks
Pageable:
in: query
style: form
      explode: true

No it generates the client correctly. Now I must convince the other department to change it.

trejjam:
That's always tough :) Good luck

PR #3858, opened 2022-02-02
Simplify `#if`'s

maxkoshevoi:
Here's list of conditional compilation symbols defined for each TFM: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/preprocessor-directives#conditional-compilation

- `NETCOREAPP` is also defined for `NET5` and `NET6`
- From what I see `!NETCOREAPP && !NETSTANDARD` is the same as `NETFRAMEWORK`
- Also `#if NET5_0_OR_GREATER` is the same as `#if NET`, but I didn't apply that here

Testing:
- Checked in Visual Studio, new conditions work as expected

Questions:
- There's a `#if NETCOREAPP1_0` here: https://github.com/RicoSuter/NSwag/blob/master/src/NSwag.AspNetCore.Launcher/Program.cs#L92, and looks like it will never be compiled. Should this code be removed?

### DIFF
--- a/src/NSwag.Commands/Commands/Generation/AspNetCore/AspNetCoreToOpenApiCommand.cs
+++ b/src/NSwag.Commands/Commands/Generation/AspNetCore/AspNetCoreToOpenApiCommand.cs
@@ -25,7 +25,7 @@ using NSwag.Generation;
 using NJsonSchema.Generation;
 using Namotion.Reflection;
 
-#if NET5_0_OR_GREATER || NETCOREAPP || NETSTANDARD
+#if NETCOREAPP || NETSTANDARD
 using System.Runtime.Loader;
 #endif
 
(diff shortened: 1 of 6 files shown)


src/NSwag.Commands/Commands/Generation/AspNetCore/AspNetCoreToOpenApiCommand.cs
```diff
@@ -25,7 +25,7 @@
 using NJsonSchema.Generation;
 using Namotion.Reflection;
 
-#if NET5_0_OR_GREATER || NETCOREAPP || NETSTANDARD
+#if NETCOREAPP || NETSTANDARD
```
RicoSuter, line 28: So NETCOREAPP includes .net 5 + 6?

RicoSuter:
> There's a #if NETCOREAPP1_0 here:

Yes, I think you can remove it (probably legacy which is not cleaned up)

RicoSuter:
Ok, now waiting for CI... but looks good. 
Thanks for the PR.

PR #3862, opened 2022-02-03
Fix npm resolving net5/6 incorrectly, change default version to net6

ovska:
Currently the npm package doesn't resolve NET6 version if `dotnet --version` returns `6.0.*`, most likely same is for NET5. User is met with an error for missing framework versions if core 2.1 is not also installed.

Core 2.1 is [out of support](https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core), replaced with NET6 (LTS release).

### DIFF
--- a/src/NSwag.Npm/bin/nswag.js
+++ b/src/NSwag.Npm/bin/nswag.js
@@ -2,7 +2,14 @@
 "use strict";
 
 var defaultCoreVersion = "Core21";
-var supportedCoreVersions = ["Core21", "Core22", "Core30", "Core31", "50", "60"];
+var supportedCoreVersions = [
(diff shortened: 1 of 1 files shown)


RicoSuter:
Thanks for the PR.

jvalkeejarvi:
Thanks for merging this PR. For some reason the binaries for version 13.15.8 were not released, assets contain only source code. Is there maybe a problem with some pipeline?

RicoSuter:
> Binaries seem to be available now. Previously there were only source code files listed under 13.15.8 release in GitHub and no .msi file for example.

Yes, i have to manually add the binaries to the release :-/

PR #3867, opened 2022-02-07
Fix for issue: 3865 content-disposition fileName parsing regex got broken recently

knoepdan:
Changes to how filename is retrieved from the response header: 

- Going back to the old regex (which we know works in the field). 
- if not successfull, try with the new Regex (which supports special characters, but was the cause for issue 3865)

Probably not to be the final solution (but in my opinion a better state then before)

### DIFF
--- a/src/NSwag.CodeGeneration.TypeScript/Templates/Client.ProcessResponse.HandleStatusCode.liquid
+++ b/src/NSwag.CodeGeneration.TypeScript/Templates/Client.ProcessResponse.HandleStatusCode.liquid
@@ -6,8 +6,8 @@ const contentDisposition = response.headers ? response.headers["content-disposit
 {%     else -%}
 const contentDisposition = response.headers ? response.headers.get("content-disposition") : undefined;
 {%     endif -%}
-const fileNameMatch = contentDisposition ? /filename="?([^"]*?)"?(;|$)/g.exec(contentDisposition) : undefined;
-const fileName = fileNameMatch && fileNameMatch.length > 1 ? fileNameMatch[1] : undefined;
+const fileNameMatch = contentDisposition ? /filename[^;=\n]*=(?:(\\?['"])(.*?)\1|(?:[^\s]+'.*?')?([^;\n]*))/g.exec(contentDisposition) : undefined;
+const fileName = fileNameMatch && fileNameMatch.length > 1 ? decodeURIComponent(fileNameMatch[fileNameMatch.length - 1]) : undefined;
 {%     if operation.WrapResponse -%}
 {%         if Framework.IsAngular -%}
 return {{ Framework.RxJs.ObservableOfMethod }}(new {{ operation.ResponseClass }}(status, _headers, { fileName: fileName, data: {% if Framework.Angular.UseHttpClient %}<any>responseBlob{% else %}<any>response.blob(){% endif %}, status: status, headers: _headers }));
--- a/src/NSwag.CodeGeneration.TypeScript/Templates/Client.ProcessResponse.HandleStatusCode.liquid
+++ b/src/NSwag.CodeGeneration.TypeScript/Templates/Client.ProcessResponse.HandleStatusCode.liquid
@@ -6,8 +6,12 @@ const contentDisposition = response.headers ? response.headers["content-disposit
 {%     else -%}
 const contentDisposition = response.headers ? response.headers.get("content-disposition") : undefined;
 {%     endif -%}
-const fileNameMatch = contentDisposition ? /filename[^;=\n]*=(?:(\\?['"])(.*?)\1|(?:[^\s]+'.*?')?([^;\n]*))/g.exec(contentDisposition) : undefined;
-const fileName = fileNameMatch && fileNameMatch.length > 1 ? decodeURIComponent(fileNameMatch[fileNameMatch.length - 1]) : undefined;
+let fileNameMatch = contentDisposition ? /filename="?([^"]*?)"?(;|$)/g.exec(contentDisposition) : undefined;
+let fileName = fileNameMatch && fileNameMatch.length > 1 ? fileNameMatch[1] : undefined;
+if(!fileName){
+	fileNameMatch = contentDisposition ? /filename[^;=\n]*=(?:(\\?['"])(.*?)\1|(?:[^\s]+'.*?')?([^;\n]*))/g.exec(contentDisposition) : undefined;
+	fileName = fileNameMatch && fileNameMatch.length > 1 ? decodeURIComponent(fileNameMatch[fileNameMatch.length - 1]) : undefined;
+}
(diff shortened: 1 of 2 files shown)


src/NSwag.CodeGeneration.TypeScript/Templates/Client.ProcessResponse.HandleStatusCode.liquid
```diff
@@ -6,8 +6,12 @@ const contentDisposition = response.headers ? response.headers["content-disposit
 {%     else -%}
 const contentDisposition = response.headers ? response.headers.get("content-disposition") : undefined;
 {%     endif -%}
-const fileNameMatch = contentDisposition ? /filename[^;=\n]*=(?:(\\?['"])(.*?)\1|(?:[^\s]+'.*?')?([^;\n]*))/g.exec(contentDisposition) : undefined;
-const fileName = fileNameMatch && fileNameMatch.length > 1 ? decodeURIComponent(fileNameMatch[fileNameMatch.length - 1]) : undefined;
+let fileNameMatch = contentDisposition ? /filename="?([^"]*?)"?(;|$)/g.exec(contentDisposition) : undefined;
+let fileName = fileNameMatch && fileNameMatch.length > 1 ? fileNameMatch[1] : undefined;
+if(!fileName){
+	fileNameMatch = contentDisposition ? /filename[^;=\n]*=(?:(\\?['"])(.*?)\1|(?:[^\s]+'.*?')?([^;\n]*))/g.exec(contentDisposition) : undefined;
+	fileName = fileNameMatch && fileNameMatch.length > 1 ? decodeURIComponent(fileNameMatch[fileNameMatch.length - 1]) : undefined;
```
KostaMadorsky, line 13: `decodeURIComponent(fileNameMatch[fileNameMatch.length - 1])` would return string 'undefined' if the group happen to be undefined.

You can make it work with the new regex and remove old code that you brought back with this:

`fileName = (fileNameMatch && fileNameMatch.length >= 4) ? fileNameMatch[3] || fileNameMatch[2]) : undefined;`
`fileName = fileName ? decodeURIComponent(fileName) : undefined;`

However, I suggest looping over the string as I explained here: https://github.com/RicoSuter/NSwag/pull/3816#issuecomment-1031502345

That would make it use UTF8 value in case when both non-UTF8 and UTF8 is provided.
`regex.exec` only gets the first match when run once.

KostaMadorsky: If you still want to run old code, move it to the end into `if (!fileName) { ... }`

zyofeng: Do we have any progress on this? had to revert to we have to revert to 13.15.5 in prod for this.

KostaMadorsky: @zyofeng I've proposed working solutions. Waiting for @knoepdan to check it out.
I will probably roll back to 13.15.5 as well for now and disable msbuild generation (I have minimal API).

KostaMadorsky: @RicoSuter could you roll back https://github.com/RicoSuter/NSwag/pull/3816 for now?
There are a few important features in the release that has it (e.g. support for minimal API).

knoepdan: Hi 
Yes I realize this PullRequest is a quick fix and should not be the final solution. Actually, I wanted to look at your proposal more closely so we that it could be the final solution. Unfortunatly, i didnt have time so far. (maybe this weekend). 

So i created this PR with the following in mind: 
- will always return filename when previously it returned one (no support for special characters)
- will always return a filename when the flawed solution returned one. (support for special character but might fail at some)
So no functionality will be broken. However, as I said, this should not be the finall solution.

KostaMadorsky: @knoepdan I think we should either fix it properly or rollback.

RicoSuter: Rolled back for now so we can find a proper solution here.

knoepdan:
Hi Kosta, Rico

I have done some tests and I believe your approach works well. Another option would be to use 2 regex. One for the special characters and one for the default one. Something like: 

1 test for a filename with special characters, minimally changed regex:  "/filename\*=(?:(\\?['"])(.*?)\1|(?:[^\s]+'.*?')?([^;\n]*))/g"  )
2. if(filename == undefined):{ try find filename without special characters

That approach should yield the same results but is - in my opinion slightly easier to understand and maintain. 
What do you think?

By the way, thanks for  your input and feedback Kosta

knoepdan:
I have adjusted the pull request, so the filename behavior is analogous to firefox. Firstly, we try to get the filename with special characters (minimally changed regex) and if that is not successful, we fall back to the regex that was in place before to read the filename. This should cover all the cases and also support responses where 2 filenames are returned.

KostaMadorsky:
@knoepdan I wonder if you could move `decodeURIComponent` into 
`if (fileName) { filename = decodeURIComponent(filename); } else { ___oldCode___ }`

Just in case the group you are capturing happens to be undefined. 
decodeURIComponent would make it a string 'undefined'.

KostaMadorsky:
@knoepdan What this part `(\\?['"])(.*?)\1|` of the regex is doing?
Could you provide some examples, where regex without it is not working?

image

KostaMadorsky:
Ah, actually, I think I got it:
https://regex101.com/r/aKss5U/1
image

KostaMadorsky:
@RicoSuter @knoepdan should we also check if `fileName && !fileName.trim()`?
Both regex expressions would match inputs with just spaces in them (e.g. filename*=UTF8''%20%20%20; or filename="    ";)
Not sure if it's the responsibility of NSwag or the consumer.

knoepdan:
Hi Kosta

oh so much input. Thanks a lot. Unfortunately, I've been really busy (and still am) but I will look at your input. Thanks a lot. 

Regarding the spaces, I think I personally, would just have NSWAG return what the server sends, even if it's just spaces, but I'm not sure and have no fixed opinion on this.  If you think a trim should be there, then I will happily implement it. I see it's advantages.

KostaMadorsky:
Seems like it's doing the right thing now.
Good work @knoepdan 

https://onecompiler.com/javascript/3xwtpr4zr
image

@RicoSuter could you have a look and merge if all is good from your side?

PR #3938, opened 2022-03-18
Used TryAddConsumes method to avoid duplicates in 'Consumes'

pavpavferg:
Previous implementation either re-created a list of 'content types or blindly appended a value from ConsumesAttribute attribute.
That blind adding caused content types duplication (in case if other preceding processes have already added a similar content type).
The fix uses a built-in TryAddConsumes method which makes sure that the content type is unique.


pavpavferg:
@RicoSuter, would you mind please checking if the PR can be merged? We have a multi-language/multi-framework app which highly depends on correct code generation.
Thanks!

RicoSuter:
Thanks for the PR.

PR #3951, opened 2022-03-29
Update to latest GA of M.E.ApiDescription.Server

captainsafia:
Update to the latest release of M.E.ApiDescription.Server package to:

- Address issues with resolving the ServiceProvider in the new host model like https://github.com/dotnet/aspnetcore/issues/40577
- Bring in bug fixes like https://github.com/dotnet/aspnetcore/pull/36355 and https://github.com/dotnet/aspnetcore/pull/39293

### DIFF
--- a/src/NSwag.ApiDescription.Client/NSwag.ApiDescription.Client.nuspec
+++ b/src/NSwag.ApiDescription.Client/NSwag.ApiDescription.Client.nuspec
@@ -15,7 +15,7 @@
     <repository type="git" url="https://github.com/RicoSuter/NSwag.git"/>
     <developmentDependency>true</developmentDependency>
     <dependencies>
-      <dependency id="Microsoft.Extensions.ApiDescription.Client" version="3.0.0" />
+      <dependency id="Microsoft.Extensions.ApiDescription.Client" version="6.0.3" />
       <dependency id="NSwag.MSBuild" version="13.14.5" />
     </dependencies>
     <references />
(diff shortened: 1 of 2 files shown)


RicoSuter:
Shouldnt we add a range here as well?
Is this backward compatible eg with .NET Core 3.1?

captainsafia:
> Shouldnt we add a range here as well? 

I figured that using a specific version would be better for dependency management here. We don't make a ton of changes to to the dotnet-openapi tool area at the moment, but if we do, it might be helpful to be intentional about validating fixes and integrating updates to the version.

> Is this backward compatible eg with .NET Core 3.1?

Yep, there's no changes that violate backwards compat here, AFAIK. cc: @dougbu

dougbu:
> Is this backward compatible eg with .NET Core 3.1?

Microsoft.Extensions.ApiDescription.Server should remain compatible w/ all (well, some will likely break something 😉) runtimes and apps back to `netcoreapp2.1`, not just 3.1.x.

> We don't make a ton of changes to to the dotnet-openapi tool area at the moment, but if we do, it might be helpful to be intentional about validating fixes and integrating updates to the version.

That said, we did make a change in Microsoft.Extensions.ApiDescription.**Client** back in December. It would be good for NSwag to pick that up if you haven't already @RicoSuter. Any &hellip;Client version from 6.0.2 on includes that fix. See dotnet/aspnetcore@e47af40def2584 (which was for 6.0.2) or the original change (dotnet/aspnetcore@e2707e1ed261e2) in 'main'.

nit: `dotnet-openapi` isn't relevant here @captainsafia 😃

dougbu:
> if you haven't already @RicoSuter

So, https://github.com/RicoSuter/NSwag/blob/bf8783461bf175212da910621fc5c486330ce960/src/NSwag.ApiDescription.Client/NSwag.ApiDescription.Client.nuspec#L18 is clearly using a very old (3.0.0) version of the Client package 😦

Sorry I didn't follow up w/ you before @RicoSuter.

captainsafia:
I went ahead and updated the Client package version in this PR as well.

captainsafia:
@RicoSuter Any other thoughts on this?

RicoSuter:
Sorry for my late reply, will this also still work with .NET Core 3.1?

captainsafia:
> Sorry for my late reply, will this also still work with .NET Core 3.1?

Yes, should be backwards-compatible as well.

RicoSuter:
Ok then, can we merge this PR or is anything missing?

captainsafia:
Yep, it can be merged. I don't have permissions to do that but this dependency bump should be sufficient.

PR #3959, opened 2022-04-04
Update OpenApiServerVariable.cs

philippemercure:
This will allow to populate the Enum from an external source, when the external source list can be dynamic in length.

```c#
    private static OpenApiServerVariable GetOpenApiServerVariable()
    {
        var environmentList = InformationHelper.GetEnvironmentList();
        
        var valid = new OpenApiServerVariable
        {
            Default = InformationHelper.GetEnvironmentDefault(),
            Enum = { "dev", "uat", "stage", "api" }
        };
        
        var invalid = new OpenApiServerVariable
        {
            Default = InformationHelper.GetEnvironmentDefault(),
            Enum = environmentList
        };

        return valid;
    }
```

### DIFF
--- a/src/NSwag.Core/OpenApiServerVariable.cs
+++ b/src/NSwag.Core/OpenApiServerVariable.cs
@@ -17,7 +17,7 @@ namespace NSwag
     {
         /// <summary>Gets or sets the enum of the server.</summary>
         [JsonProperty(PropertyName = "enum", DefaultValueHandling = DefaultValueHandling.IgnoreAndPopulate)]
-        public ICollection<string> Enum { get; } = new Collection<string>();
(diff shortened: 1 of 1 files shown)


RicoSuter:
Isnt this 

            Enum = { "dev", "uat", "stage", "api" }

already working as it will call Add() multiple times?

philippemercure:
Yes, exactly, this is working. What is not working is this

```c#
var invalid = new OpenApiServerVariable
        {
            Default = InformationHelper.GetEnvironmentDefault(),
            Enum = environmentList
        };
```

We can't use a list in the Enum because the "set" property is absent from the property declaration.

Maybe this was by design, but I can't see why this would cause problem.

By adding the "set" we can then use a list that would be populated from somewhere else, not only from the constructor with a list of const.

PR #3990, opened 2022-04-26
Update NSwag.ApiDescription.Client.nuspec

mohammadshahabrafiq:
To fix the issue mentioned https://github.com/RicoSuter/NSwag/issues/3987 . We want to use the latest tool chain version which has the supported bug fixes needed.


RicoSuter:
Created PR to fix this for all new versions: 
https://github.com/RicoSuter/NSwag/pull/4011

lahma:
I wonder if the whole nuspec thing would work better if there were a csproj file with deps and then building nupkg based on that (not sure if it works, but a theory), but I see that @RicoSuter has new automation to work around the current problem.

RicoSuter:
I think/hope that's good enough for now...
it's not really a dll based nuget so csproj might not work...

RicoSuter:
Fixed in the referenced PR. Thanks for the PR!

mohammadshahabrafiq:
No problem. Happy to contribute.

PR #4005, opened 2022-05-12
Added check on SupportsTypeOnlyImports when TypeScript version is >= 4.3 and updated File.liquid template for TypeScript

tomvdk:
- The TypeScript compiler type checking produces an error for the import of the axios types
- Not possible to change the preserveValueImports or isolatedModules compiler options because they are needed by vite-vue
- also see Issue 3964 - type-only imports for axios (https://github.com/RicoSuter/NSwag/issues/3964)
- Changed axios imports to support import type when typescript version >= 4.3 by adding SupportsTypeOnlyImports in TypeScriptFileTemplateModel.cs
- It is supported as from TypeScript version 3.8 but only versions 2.7, 4.3, so picked 4.3 (https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html)

### DIFF
--- a/src/NSwag.CodeGeneration.TypeScript/Models/TypeScriptFileTemplateModel.cs
+++ b/src/NSwag.CodeGeneration.TypeScript/Models/TypeScriptFileTemplateModel.cs
@@ -164,6 +164,9 @@ namespace NSwag.CodeGeneration.TypeScript.Models
         /// <summary>Gets a value indicating whether the target TypeScript version supports override keyword.</summary>
         public bool SupportsOverrideKeyword => _settings.TypeScriptGeneratorSettings.SupportsOverrideKeyword;
 
+        /// <summary>Gets a value indicating whether the target TypeScript version supports Type-Only imports</summary>
+        public bool SupportsTypeOnlyImports => _settings.TypeScriptGeneratorSettings.TypeScriptVersion >= 4.3m;
+
         private string GenerateExtensionCodeAfter()
         {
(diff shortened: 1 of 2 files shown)


src/NSwag.CodeGeneration.TypeScript/Models/TypeScriptFileTemplateModel.cs
```diff
@@ -164,6 +164,9 @@ public IEnumerable<string> ResponseClassNames
         /// <summary>Gets a value indicating whether the target TypeScript version supports override keyword.</summary>
         public bool SupportsOverrideKeyword => _settings.TypeScriptGeneratorSettings.SupportsOverrideKeyword;
 
+        /// <summary>Gets a value indicating whether the target TypeScript version supports Type-Only imports</summary>
+        public bool SupportsTypeOnlyImports => _settings.TypeScriptGeneratorSettings.TypeScriptVersion >= 4.3m;
```
lahma, line 168: I think you should check against 3.8 because it's the version that supports it. User can define free value in JSON configuration,  it's only config ui that has fixed options.

tomvdk: Thanks for your reply. Changed to 3.8. See commit: https://github.com/RicoSuter/NSwag/pull/4005/commits/a2c128053865b82113527abb75587c33b19b0080

lahma (approved):

RicoSuter:
Thanks for the PR.

PR #4011, opened 2022-05-18
Fix NSwag.ApiDescription.Client.nuspec gen

RicoSuter:


lahma (approved):
Nice NUKE usage 🚀

RicoSuter:
Doesnt look "best practice" but if it works i'm fine with it :-)

lahma:
I'm glad the code is concise and can be done using the automation and constructs present. Patching is always nasty, but the implementation is clean 👍🏻

PR #4031, opened 2022-06-02
fix axios tamplate JSON.serialize https://github.com/RicoSuter/NSwag/issues/3294

Yamaraj6:
remove json serialize for axiose because axios return object not json
https://github.com/RicoSuter/NSwag/issues/3294


jamshally:
@RicoSuter - looking forward to seeing this merged an in a release!

dmtx97:
It'd be great to get this merged in so we don't need to manually ignore transformRequest when creating the axios client.

DavisZ94:
@RicoSuter would it be possible to merge this one?

RicoSuter:
I think this one will affect not only axios but also the others... 
Should probably only be changed for axios, right?

Yamaraj6:
image
A few lines above, there's an {% if Framework.IsAxios -%} condition, so this particular line should only affect Axios in my opinion and not others frameworks.

PR #4038, opened 2022-06-07
Add ability to customize the type name for generated DTOs/models

archevel:
Models/Dto class names were picked from the spec provided when generating the code. This PR adds the ability yo customize the name of models with a template string, e.g. "My{model}IsNicer".

The main reason for me for adding this is for convinience when searching for types in an IDE where there also exist other model classes with similar names it is nicer to be able to add a prefix to the generated classes so that they don't show as the initial suggestions.

This could already be acomplished by setting the `TypeNameGenerator` property on the settings. However, that is a bit more complicated and not as accessible as having this property.

Note that if the template string does not contain "{model}" it will be ignored. That way the generated code doesn't become too messed up if someone just specified an explicit name (which all models would then be called, which is likely not what the user wants).


archevel:
Please let me know if something is missing or if this kind of change is not desireable (since it can be accomplished through the use of the `TypeNameGenerator `).

archevel:
@RicoSuter Would you be able to take a look at this PR and provide some feedback in case something is amiss?

Issue #4027, opened 2022-05-30
[Nswag] [NetCore] [MinimalApi] Wrong Signature


Magic73:
Hi guys,

I have a Netcore webApi built with Minimalapi.
I have an endpoint defined like this:

app.MapGet("api/details/data/{code}", (string code) => GetDataDetailsAsync(code));

this generate a swagger, like this:

GET /api/details/data/{code}

![image](https://user-images.githubusercontent.com/7645382/171037984-7f00afce-eabe-4f44-9b4d-9a31d07543ac.png)

and when I create a client with nswag (latest version), it create these signatures:

`        System.Threading.Tasks.Task<Utente> Conto{code}Async(string code);
        System.Threading.Tasks.Task<Utente> Conto{code}Async(string code, System.Threading.CancellationToken cancellationToken);
`

Issue #4002, opened 2022-05-10
NSwag Studio generates C# method with parentheses if operation id has parentheses (fails to compile)


jcoliz:
1. Create a simple OpenAPI.yaml with ()'s in the operation ID. Example below
2. Launch NSwag Studio (v13.15.10.0)
3. File > New
4. Runtime -> Net60
5. Specification URL -> the yaml from step 1
6. Outputs -> check Csharp Client 
7. Click "Generate Outputs"
8. Notice that the generated method is not going to compile (example below)

Note that the Open API spec only says "string" when describing operationID, so it seems that this operationID is allowed.

```yaml
openapi: 3.0.0
info:
  title: API
  version: 1.0.0
servers:
  - url: https://api.something.com
paths:
  /api/hello:
    get:
      description: ''
      operationId: GetSome(huge)data
      responses:
        200:
          description: Success

```

```c#
        public virtual System.Threading.Tasks.Task GetSome(huge)dataAsync()
```

PR #4041, opened 2022-06-09
Fix Wrong Method Name generation with invalid characters

Magic73:
when there is a route with parameters inside like this:

GET /api/details/data/{code}

the client is generated with wrong signatures:

System.Threading.Tasks.Task<Utente> Conto{code}Async(string code);
System.Threading.Tasks.Task<Utente> Conto{code}Async(string code, System.Threading.CancellationToken cancellationToken);

This fix will remove any invalid character from  the signature

### DIFF
--- a/src/NSwag.CodeGeneration/ClientGeneratorBase.cs
+++ b/src/NSwag.CodeGeneration/ClientGeneratorBase.cs
@@ -8,6 +8,7 @@
 
 using System.Collections.Generic;
 using System.Linq;
+using System.Text.RegularExpressions;
 using NJsonSchema;
 using NJsonSchema.CodeGeneration;
 using NSwag.CodeGeneration.Models;
@@ -159,6 +160,9 @@ namespace NSwag.CodeGeneration
                     var operation = p.Value;
 
                     var operationName = BaseSettings.OperationNameGenerator.GetOperationName(document, path, httpMethod, operation);
+
+                    // Strip any character that is not a word character or a period
+                    operationName = Regex.Replace(operationName, @"[^\w\.]", "", RegexOptions.None);
 
                     if (operationName.Contains("."))
                     {


Magic73:
Fix for bug #4027 and #4002

johannesmols:
It would be really great to merge this, I'm having the same issue with invalid clients due templated path names in my specification.

src/NSwag.CodeGeneration/ClientGeneratorBase.cs
```diff
@@ -160,6 +161,9 @@ private List<TOperationModel> GetOperations(OpenApiDocument document)
 
                     var operationName = BaseSettings.OperationNameGenerator.GetOperationName(document, path, httpMethod, operation);
 
+                    // Strip any character that is not a word character or a period
+                    operationName = Regex.Replace(operationName, @"[^\w\.-]", "", RegexOptions.None);
```
lahma, line 165: performance will be probably bad as your are calling dynamic regex inside a loop

johannesmols: @lahma I'm not sure if that really makes a difference, but the RegexOptions.Compiled could be used to speed it up. I can see that dynamic regexe's are used in a few other places though (like `SingleClientFromPathSegmentsOperationNameGenerator`).

I'm not facing this issue with `MultipleClientsFromPathSegments`, but I would really like to generate a single client. Maybe it just involves using similar logic in the other generators to avoid invalid names?

johannesmols: Okay I just found out that the curly braces are actually handled fine when using `SingleClientFromPathSegments`. I had a single area called `/upload?id={id}` that caused it to generate incorrect methods names.

PR #4049, opened 2022-06-16
Change static read-only properties of OpenApiOperationMethod to constants

PatrickHofman:
Change static read-only properties to constants to help optimizer and be able to use switches.


RicoSuter:
Thanks for the PR.

PR #4086, opened 2022-07-15
Adding in better handling for plain text body

### LINKED ISSUE #3824: Generating client off of text/plain request body serializes the string
I have a functioning api endpoint that accepts "text/plain" as the body. The user of this endpoint is a third-party that controls the content type so I can't change that. I'm using NSwag to generate a c# client based on a swagger.json.

My main problem is that the generated client's function for calling this endpoint still serializes the `string` body, which causes the endpoint to fail to parse. I'm not 100% certain if this is a bug in my swagger.json or if it's something unintentional, because the json is also generated.

Here's what the swagger.json looks like:
```json
{
  "openapi": "3.0.1",
  "info": {
    "title": "PlainTextParser",
    "version": "1.0"
  },
  "paths": {
    "/api/v1/text": {
      "post": {
        "tags": [
          "TextEvent"
        ],
        "summary": "Parses a plain text body",
        "operationId": "PostTextEvent",
        "requestBody": {
          "description": "Accepts a plain text body",
          "content": {
            "text/plain": {
              "schema": {
                "type": "string",
                "nullable": true
              }
            }
          }
        },
        "responses": {
          "202": {
            "description": "Success"
          }
        }
      }
    }
  },
  "components": { }
}
```

When I run nswag on it, I get this inside of the generated client:

```c#
public virtual async System.Threading.Tasks.Task PostTextEventAsync(string body, System.Threading.CancellationToken cancellationToken)
{
  //...

  var content_ = new System.Net.Http.StringContent(Newtonsoft.Json.JsonConvert.SerializeObject(body, _settings.Value));
  content_.Headers.ContentType = System.Net.Http.Headers.MediaTypeHeaderValue.Parse("text/plain");

  //...
}
```

The `SerializeObject` method call is what is causing the issue. When I removed it everything worked fine. You can even see on the second line there that it understands the content type, which is why I'm confused why it's trying to serialize it. After poking around the code a bit, I found this here: https://github.com/RicoSuter/NSwag/blob/f22bf3594a34073e67bfc6e84eec1040d9537990/src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.liquid#L196-L216. It doesn't seem to have anything specific to a text/plain type as far as I can tell. I looked into the [OpenApiBodyParameter](https://github.com/RicoSuter/NSwag/blob/a82c343baf166a118415df9b93023c16a9b43f0a/src/NSwag.Core/OpenApiParameter.cs), but I couldn't see any direct references to the text/plain content type.

So am I missing something or should this not be serialized?

### LINKED ISSUE #3860: C# client generation with requestbody as text/plain and string
When generating an endpoint that has a requestBody like this:
```yaml
      requestBody:       
        required: true
        content:
          text/plain:
            schema:
              type: string
              minLength: 18
              maxLength: 18
```
The generated C# method does this:
```C#
 var content_ = new System.Net.Http.StringContent(System.Text.Json.JsonSerializer.Serialize(body, _settings.Value));
```
However this adds some extra quotes around the body since we don't want to serialize this as json. This wil make the request invalid. Is there anyway to fix this or does someone have a workaround? (Other than manually altering the generated code)

JonruAlveus:
Removes the json serialising for plain text bodies.
Fixes #3824 
Fixes #3860 

Possibly others...

Apologies in advance as this is my first change here and I really don't know too many of the potential issues this change might raise!

### DIFF
--- a/src/NSwag.CodeGeneration.CSharp.Tests/PlainResponseTests.cs
+++ b/src/NSwag.CodeGeneration.CSharp.Tests/PlainResponseTests.cs
@@ -9,7 +9,7 @@ namespace NSwag.CodeGeneration.CSharp.Tests
 {
     public class PlainResponseTests
     {
-        [Fact]    
+        [Fact]
         public async Task When_openapi3_reponse_contains_plain_text_string_array_then_ReadObjectResponseAsync_is_generated()
         {
             // Arrange
--- a/src/NSwag.CodeGeneration.CSharp.Tests/PlainTextBodyTests.cs
+++ b/src/NSwag.CodeGeneration.CSharp.Tests/PlainTextBodyTests.cs
@@ -1,0 +1,82 @@
+﻿using System.Threading.Tasks;
+using Xunit;
+
+namespace NSwag.CodeGeneration.CSharp.Tests
+{
+    public class PlainTextBodyTests
+    {
+        [Fact]
+        public async Task When_request_contains_plain_text_string_body_then_ReadObjectResponseAsync_is_generated()
+        {
+            // Arrange
+            var json = @"{
+  ""swagger"":  ""2.0"",
+  ""paths"": {
+    ""/api/get-ienumerable"": {
+    ""put"":  {
+      ""tags"":  [
+        ""BuildType""
+      ],
+      ""summary"":  ""Update value of build parameter."",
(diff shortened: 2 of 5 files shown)


RicoSuter:
LGTM, thanks for the PR.

PR #4105, opened 2022-08-15
Allow both the summary and description from OpenAPI to be added as summary and remarks in c# generated code

codecooper:
The following changes will add the `description` property from an OpenAPI operation under the `<remarks>` xml comment in c# client code. Unit tests were updated to check for the annotations. My OpenAPI specs are typically quite detailed in the summary and description and and I wanted to add that documentation to the generated controllers and clients.


lahma (approved):
LGTM

RicoSuter:
Thanks for the PR.

PR #4129, opened 2022-09-01
WebApiOpenApiDocumentGenerator: Support multiple Route-attributes on Controller level

andreasbalzer:
This PR adds support for multiple Route-attributes on controller-level. For example, an API could respond both as api/beverages and api/fluids.

```
        [System.Web.Http.Route("api/{regionId:int}/beverages")]
        [System.Web.Http.Route("api/{regionId:int}/fluids")]
        public class BeveragesController : RegionalItemController
        {
            [System.Web.Http.HttpGet]
            public string[] Get(int regionId)
            {
                throw new NotImplementedException();
            }
        }
```

**Observed behavior:** WebApiOpenApiDocumentGenerator threw an exception, as it only allowed one Route-attribute on controller level

**Expected behavior:** Two paths get created.

Issue addressed partly: [1354 NSwag.MSBuild does not support API versioning or multiple routes](https://github.com/RicoSuter/NSwag/issues/1354)

### DIFF
--- a/src/NSwag.Generation.WebApi.Tests/Attributes/RouteTests.cs
+++ b/src/NSwag.Generation.WebApi.Tests/Attributes/RouteTests.cs
@@ -27,8 +27,8 @@ namespace NSwag.Generation.WebApi.Tests.Attributes
             }
         }
 
-        [System.Web.Http.Route("api1/{regionId:int}/food")]
-        [System.Web.Http.Route("api2/{regionId:int}/food")]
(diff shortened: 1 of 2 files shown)


andreasbalzer:
@RicoSuter please let me know if you like to see changes or if there is any way I can help.

RicoSuter:
As I do not actively support WebAPI anymore I'll just assume this is fine. 
Thanks for the PR.

PR #4136, opened 2022-09-06
Feature/#4131 store serialized json in separate variable for easy inspection

### LINKED ISSUE #4131: Please generate json in variable instead of StringContent constructor
In generated clients the json serialization of posted bodies is directly passed to the StringContent constructor. This removes a possibility to inspect the json that is sent. Please use a variable to hold the json before it is passed to StringContent. That would allow to see what is ultimately sent to the server. I.e.

```csharp
var json_ = Newtonsoft.Json.JsonConvert.SerializeObject(hotels, _settings.Value);
var content_ = new System.Net.Http.StringContent(json_);
content_.Headers.ContentType = System.Net.Http.Headers.MediaTypeHeaderValue.Parse("application/json");
request_.Content = content_;
request_.Method = new System.Net.Http.HttpMethod("POST");
request_.Headers.Accept.Add(System.Net.Http.Headers.MediaTypeWithQualityHeaderValue.Parse("application/json"));
```

dlidstrom:
This fixes #4131. The serialized content and parameters are saved in separate variables in order to make it possible to see what is sent. This adds an inspection point that was missing IMHO.

### DIFF
--- a/src/NSwag.CodeGeneration.CSharp.Tests/ObjectParameterTests.cs
+++ b/src/NSwag.CodeGeneration.CSharp.Tests/ObjectParameterTests.cs
@@ -1,4 +1,4 @@
-﻿using System.Threading.Tasks;
+using System.Threading.Tasks;
 using Xunit;
 
 namespace NSwag.CodeGeneration.CSharp.Tests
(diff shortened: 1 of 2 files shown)


lahma (approved):
LGTM

RicoSuter:
Thanks for the PR.

PR #4140, opened 2022-09-09
Client code generation in Console app fails with NullReferenceException

rpbeukes:
This affected the C# console app.
Issue: https://github.com/RicoSuter/NSwag/issues/2530


RicoSuter:
Thanks for the fix.

PR #4143, opened 2022-09-13
add new MultipleClientsFromFirstTagAndOperationNameGenerator

mw-survit:
Added a new OperationNameGenerator which takes first tag for clientName and the operationName (underscore separated from the operationId). Inherits from the already existing MultipleClientsFromOperationIdOperationNameGenerator.

This allows to have several small controllers that are tagged the same, to be generated into a single client.
```
[Tags("stuff")]
public class AController { public void A () {} }

[Tags("stuff")]
public class BController { public void B() {} }

public class CController { public void C() {} }
```

would then result in 

```
public class StuffClient
{
    public void A () {}
    public void B () {}
}

public class CClient 
{
    public void C() {}
}
```


RicoSuter:
Thanks for the PR.

PR #4166, opened 2022-09-29
export namespace

aouizerats:
when generate typescipt with namespace and export types, the export keyword is missing in namespace declaration


aouizerats:
@RicoSuter can you approve this PR. I need this change as soon  is possible. 
Thanks

PR #4172, opened 2022-10-07
Fix serialization of x-www-forms-urlencoded values with System.Text.Json or request and response serialization settings

Chryssie:


CNBoland:
The issue addressed in this PR affects me as well. I've tested the code change in a project of my own and concur.

@Chryssie Good job!

olegd-superoffice:
Fix for this was merged in #4402 so I guess this one can be deleted?

RicoSuter:
Fixed by #4402

PR #4175, opened 2022-10-10
Serialize to UTF-8 bytes with System.Text.Json.

olegd-superoffice:
Serializing to a UTF-8 byte array is about 5-10% faster than using the string-based methods.


ovska:
This can be further improved by using `JsonContent` so the payload is serialized directly to Http stream: `JsonContent.Create(payload, options: _settings.Value)`. The downside is that this api seems to be available only from .NET 5 onwards, so a `#if NET5_0_OR_GREATER` preprocessor symbol might be required, which make the method more complex.

olegd-superoffice:
> This can be further improved by using `JsonContent` so the payload is serialized directly to Http stream: `JsonContent.Create(payload, options: _settings.Value)`. The downside is that this api seems to be available only from .NET 5 onwards

@ovska  I think this API is available for any supported .Net version, but it adds dependency on [System.Net.Http.Json NuGet package](https://www.nuget.org/packages/System.Net.Http.Json). Could be worth another PR when/if this one is accepted.

olegd-superoffice:
I updated this PR to use UTF-8 byte array serialization when producing `application/x-www-form-urlencoded` in generated client.

olegd-superoffice:
@RicoSuter done.

PR #4184, opened 2022-10-17
Do not set clientSecret for CODE-PKCE

trejjam:
Hi, my oauth server is complaining that I send `client_secret` when using code flow with PKCE. This PR should resolve the issue

### DIFF
--- a/src/NSwag.AspNet.Owin/SwaggerUi3/index.html
+++ b/src/NSwag.AspNet.Owin/SwaggerUi3/index.html
@@ -85,16 +85,21 @@ window.onload = function() {
   });
 
   if ("{ClientId}") {
-    ui.initOAuth({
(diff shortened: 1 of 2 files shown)


RicoSuter:
I need to revert this PR as it breaks our application. 
With the improved revert just dont specify clientSecret and it should not be an issue in your case.

trejjam:
Ok, no problem.

I am curious, how it can break the application

RicoSuter:
In our case we need the secret and usePkceWithAuthorizationCodeGrant = true, dont ask me why...
With the changes it should support your case (usePkceWithAuthorizationCodeGrant  = true and no secret) and also the (usePkceWithAuthorizationCodeGrant = true and a secret) case.

trejjam:
Got it, thanks :)

PR #4195, opened 2022-10-21
Enable support for generating DTOs as C# 9.0 records

ccopsey:
This is a duplicate of 3655 to see if I can get it over the line.


lahma:
Record support is in master already.

Issue #2456, opened 2019-10-10
Code generation does not support wildcards for status codes in responses


cmann-andagon:
According to the OpenApi 3 specification, the Responses object can contain a wildcard like `4xx`. I have the following in my spec:
```
"responses": {
    "4xx": {
        "description": "",
        "content": {
            "application/json": {
                "schema": {
                    "$ref": "#/components/schemas/ApiError"
                }
            }
        }
    },
    ...
}
```
The CSharpClientGenerator generates the following code for the error handling:
```
ProcessResponse(client_, response_);

var status_ = ((int)response_.StatusCode).ToString();
if (status_ == "4xx") 
{
    var objectResponse_ = await ReadObjectResponseAsync<ApiError>(response_, headers_).ConfigureAwait(false);
    throw new ApiException<ApiError>("A server side error occurred.", (int)response_.StatusCode, objectResponse_.Text, headers_, objectResponse_.Object, null);
}
```

The check for `4xx` does not work and the client falls back to the default error handling. The check should be generated as follows when a wildcard is present:
```
ProcessResponse(client_, response_);

var status_ = (int)response_.StatusCode;
if (status_ >= 0 && status_ <= 499) 
{
    var objectResponse_ = await ReadObjectResponseAsync<ApiError>(response_, headers_).ConfigureAwait(false);
    throw new ApiException<ApiError>("A server side error occurred.", (int)response_.StatusCode, objectResponse_.Text, headers_, objectResponse_.Object, null);
}
```

This same issue seems to be present in the TypeScriptClientGenerator.

cpirtea-bun:
relates to https://github.com/RicoSuter/NSwag/issues/4141

peterdrier:
Not sure how this was classified an enhancement.  It's straight up broken when using valid elements from the OpenAPI 3.0 spec. 

Is there something holding up the PR?

PR #4204, opened 2022-10-31
fixes #4141 and #2456 OpenAPI 3 support response ranges like "5XX" meaning 500..599

### LINKED ISSUE #4141: Support OpenAPI 3 response Status code ranges
OpenAPI 3 support response ranges line "5XX" meaning 500..599.

Nswag is unable to generate valid client if such ranges are used.

If that is too much work, it would be great if checking the status code was done via an overridable property or event so we could override it ourselves if nswag will not natively support the ranges.

E.G.:
Instead of 

        else
        if (status_ == 5XX)

it would be something like

        else
        if (InternalMatchStatus(status_, "5XX"))

where InternalMatchStatus would be something like

        private System.Func<int, string,bool> MatchStatus { get; set; }
        private bool InternalMatchStatus(int value, string match)
        {
            return MatchStatus?.Invoke(value, match) ?? value == int.Parse(match); 
        }

cpirtea-bun:
fixes https://github.com/RicoSuter/NSwag/issues/4141 for both C# and TypeScript
also fixes https://github.com/RicoSuter/NSwag/issues/2456

### DIFF
--- a/src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.liquid
+++ b/src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.liquid
@@ -322,7 +322,7 @@
 
                     var status_ = (int)response_.StatusCode;
 {%     for response in operation.Responses -%}
-                    {% if response.StatusCodeIsInt %}if (status_ == {{ response.StatusCode }}{% if response.CheckChunkedStatusCode %} || status_ == 206{% endif %}){% else %}if (StatusInRange(status_, {{ response.CreateRegex }})){% endif %}
+                    {% if response.StatusCodeIsInt %}if (status_ == {{ response.StatusCode }}{% if response.CheckChunkedStatusCode %} || status_ == 206{% endif %}){% else %}if (StatusInRange(status_, {{ response.CreateRegexFoxCS }})){% endif %}
                     {
                         {% template Client.Class.ProcessResponse %}
                     }
--- a/src/NSwag.CodeGeneration.TypeScript/Templates/Client.ProcessResponse.liquid
+++ b/src/NSwag.CodeGeneration.TypeScript/Templates/Client.ProcessResponse.liquid
@@ -3,7 +3,7 @@
 let _mappings: { source: any, target: any }[] = [];
 {% endif -%}
 {% for response in operation.Responses -%}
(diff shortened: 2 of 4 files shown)


emmanuelmathot:
How can this PR merge be moved up?

cpirtea-bun:
Hi folks, any word on when could this be merged? I am stuck using 13.6.2 because no other version is able to generate buildable code.

src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.liquid
```diff
@@ -405,6 +405,14 @@
         public string Text { get; }
     }
 
+    private bool StatusInRange(int value, string match)
+    {
+        var regexBody = match.Replace("x", "\\d", System.StringComparison.OrdinalIgnoreCase);
```
minichma, line 410: Building a regex from external input should usually be avoided. Consider parsing the first digit using a pattern like `@"^(?<class>\d)xx$`. This would also allow statically holding a pre-compiled regex instead of creating a new one on every request, thus avoiding multiple compilations. You could also consider not using a regex altogether and instead parsing completely old fashioned (match.length == 3) && (match[1] == ...).

src/NSwag.CodeGeneration.TypeScript/Templates/Client.ProcessResponse.liquid
```diff
@@ -3,7 +3,7 @@
 let _mappings: { source: any, target: any }[] = [];
 {% endif -%}
 {% for response in operation.Responses -%}
-if (status === {{ response.StatusCode }}{% if response.CheckChunkedStatusCode %} || status === 206{% endif %}) {
+{% if response.StatusCodeIsInt %}if (status === {{ response.StatusCode }}{% if response.CheckChunkedStatusCode %} || status === 206{% endif %}){% else %}if (`${status}`.match('^{{ response.StatusCode }}$'.replace(/x/gi, '\\d'))){% endif %} {
```
minichma, line 6: Same as in CS case, using external input as regex should probably avoided. Maybe use something like `(`${status/100}` + 'xx') === response.StatusCode.toLower()`. (pseudo code)

minichma (approved):
Would be great, if we'd get this merged soon. Aside from a possible minor improvement, TLGTM.

src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.liquid
cpirtea-bun: Please see update. I made the regex part of the generated compile time code.

cpirtea-bun:
@RicoSuter, it's been a year since I raised this. Would it be possible to merge it, please?

PR #4213, opened 2022-11-05
Expose tags in operation model base

chivandikwa:
Expose operation tags in the OperationModelBase. These will come very handy on overriding template and to make it easy to work with from liquid I have exposed these as a comma delimited string.

Please kindly accept this PR if this change is reasonable and adheres to standards, thanks.

### DIFF
--- a/src/NSwag.CodeGeneration/Models/OperationModelBase.cs
+++ b/src/NSwag.CodeGeneration/Models/OperationModelBase.cs
@@ -289,6 +289,9 @@ namespace NSwag.CodeGeneration.Models
         /// <summary>Gets the operation extension data.</summary>
         public IDictionary<string, object> ExtensionData => _operation.ExtensionData;
 
+        /// <summary>Gets the operation tags </summary>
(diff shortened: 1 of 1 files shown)


src/NSwag.CodeGeneration/Models/OperationModelBase.cs
```diff
@@ -289,6 +289,9 @@ public string Produces
         /// <summary>Gets the operation extension data.</summary>
         public IDictionary<string, object> ExtensionData => _operation.ExtensionData;
 
+        /// <summary>Gets the operation tags </summary>
+        public string Tags => string.Join(",", _operation.Tags);
```
RicoSuter, line 293: I think we should expose the string array, allows more custom scenarios in templates and you can still join within the liquid templates if needed.

fauzanzaid: Addressed this PR comment in pr #4807
Kindly review that PR

PR #4220, opened 2022-11-08
Add .Net 7 target

### LINKED ISSUE #4194: .NET 7 support
We are currently onboarding .NET 7 (preview) in our project and ran into problems when using `NSwagExe_Net60`:
`System.TypeLoadException: Could not load type 'System.Numerics.INumber1'` (interface added in .NET 7).

The reason for the failure is that NSwag generator runs using .NET 6 runtime and tries use reflection to load .NET 7 dll.

**Is there a plan to add .NET 7 support?**
I guess the PR would be very similar as: https://github.com/RicoSuter/NSwag/pull/3632/

Our hack to test the functionality meanwhile is to change `Microsoft.NETCore.App` and `Microsoft.AspNetCore.App` properties to current preview .NET 7 version (e.g. 7.0.0-rc.2.22472.3) in `tools\Net60\dotnet-nswag.runtimeconfig.json` in the .nuget folder.

### LINKED ISSUE #4219: .Net 7.0.0 now on nuget
Hi - due to the constraint on the .net 7 packages, we can't install NSwag.AspNetCore in a project using .net 7.0.0 packages now.

image

trejjam:
fixes #4219
fixes #4194

### DIFF
--- a/src/NSwag.CodeGeneration/NSwag.CodeGeneration.csproj
+++ b/src/NSwag.CodeGeneration/NSwag.CodeGeneration.csproj
@@ -6,7 +6,7 @@
     <DocumentationFile>bin\$(Configuration)\$(TargetFramework)\$(AssemblyName).xml</DocumentationFile>
   </PropertyGroup>
   <ItemGroup>
-    <PackageReference Include="Newtonsoft.Json" Version="9.0.1" />
+    <PackageReference Include="Newtonsoft.Json" Version="10.0.1" />
     <PackageReference Include="NJsonSchema" Version="10.8.0" />
     <PackageReference Include="NJsonSchema.CodeGeneration" Version="10.8.0" />
   </ItemGroup>
--- a/src/NSwag.Core/NSwag.Core.csproj
+++ b/src/NSwag.Core/NSwag.Core.csproj
@@ -7,7 +7,7 @@
     <DocumentationFile>bin\$(Configuration)\$(TargetFramework)\$(AssemblyName).xml</DocumentationFile>
   </PropertyGroup>
   <ItemGroup>
-    <PackageReference Include="Newtonsoft.Json" Version="9.0.1" />
+    <PackageReference Include="Newtonsoft.Json" Version="10.0.1" />
     <PackageReference Include="NJsonSchema" Version="10.8.0" />
   </ItemGroup>
   <ItemGroup Condition="'$(TargetFramework)' == 'net45'">
--- a/src/NSwag.Generation.WebApi/NSwag.Generation.WebApi.csproj
+++ b/src/NSwag.Generation.WebApi/NSwag.Generation.WebApi.csproj
@@ -6,7 +6,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Newtonsoft.Json" Version="9.0.1" />
+    <PackageReference Include="Newtonsoft.Json" Version="10.0.1" />
   </ItemGroup>
 
   <ItemGroup Condition="'$(TargetFramework)' == 'net45'">
--- a/src/NSwag.Generation/NSwag.Generation.csproj
+++ b/src/NSwag.Generation/NSwag.Generation.csproj
@@ -7,7 +7,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Newtonsoft.Json" Version="9.0.1" />
+    <PackageReference Include="Newtonsoft.Json" Version="10.0.1" />
     <PackageReference Include="NJsonSchema" Version="10.8.0" />
   </ItemGroup>
 
--- a/src/NSwag.Integration.Console/NSwag.Integration.Console.csproj
+++ b/src/NSwag.Integration.Console/NSwag.Integration.Console.csproj
@@ -38,6 +38,6 @@
     <PackageReference Include="Microsoft.AspNet.WebApi.Client" Version="5.2.3" />
     <PackageReference Include="Microsoft.AspNet.WebApi.Core" Version="5.2.3" />
     <PackageReference Include="Microsoft.AspNet.WebApi.WebHost" Version="5.2.3" />
-    <PackageReference Include="Newtonsoft.Json" Version="9.0.1" />
+    <PackageReference Include="Newtonsoft.Json" Version="10.0.1" />
   </ItemGroup>
 </Project>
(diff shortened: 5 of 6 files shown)


sergey-tihon:
global.json probably should be updated as well
https://github.com/RicoSuter/NSwag/blob/master/global.json#L3

.github/workflows/build.yml
```diff
@@ -32,19 +32,25 @@ jobs:
...
+      - uses: actions/setup-dotnet@v3
         with:
           dotnet-version: 5.0.*
-      - uses: actions/setup-dotnet@v1
+      - uses: actions/setup-dotnet@v3
         with:
           dotnet-version: 6.0.*
+      - uses: actions/setup-dotnet@v3
+        with:
+          dotnet-version: 7.0.*
```
prom3theu5, line 52: You can remove preview annotation now. 7.0.100 sdk is out

lahma (approved):
Splendid work, looks good to me.

RicoSuter:
Also need to update the npm js file: 
https://github.com/RicoSuter/NSwag/blob/master/src/NSwag.Npm/bin/nswag.js#L11

TonyMoonStar:
Hi,

Some of the 7.0.0 libraries support Net6 and Net7, as example EF Core libraries.
NSwag overlaps on Net6 as
NSwag.AspNetCore -> NSwag.Generation.AspNetCore -> Microsoft.Extensions.DependencyInjection.Abstractions (>= 6.0.0 && < 7.0.0)

Where EF Core dependencies are 7.0.0.

Also with Microsoft.Extensions.Options.

for the Generation should be next may be?
```cs
  <ItemGroup Condition="'$(TargetFramework)' == 'net6.0'">
    <FrameworkReference Include="Microsoft.AspNetCore.App" />
    <PackageReference Include="Microsoft.Extensions.DependencyInjection.Abstractions" Version="[6, 8)" />
    <PackageReference Include="Microsoft.Extensions.Options" Version="[6, 8)" />
  </ItemGroup>
```

---
We tend to use Net6, but since it's a Target 7 updated, could uptake also these changes.

trejjam:
What can possibly go wrong :D
The constraints relaxed

Mahdi-Joveyni (approved):

GeorgDangl (approved):

src/NSwag.Sample.NET70Minimal/nswag.json
```diff
@@ -0,0 +1,58 @@
+﻿{
+  "runtime": "Net60",
```
patrickklaeren, line 2: This seems wrong?

trejjam: Sure, it is. Thanks

Dzivo:
Do you have an idea of release schedule

r-work:
This fails if your project is net 7 and have `<PlatformTarget>x64</PlatformTarget>` in your csproj with the following message:

```
 Detected package downgrade: System.Collections from 4.3.0 to 4.0.11. Reference the package directly from the project to select a different version. 
 MiSystems.GeoServices -> NSwag.AspNetCore 13.17.1 -> NSwag.Core 13.17.1 -> Newtonsoft.Json 9.0.1 -> System.Xml.ReaderWriter 4.0.11 -> System.IO.FileSystem 4.0.1 -> runtime.win.System.IO.FileSystem 4.3.0 -> System.Collections (>= 4.3.0) 
 MiSystems.GeoServices -> NSwag.AspNetCore 13.17.1 -> NSwag.Core 13.17.1 -> Newtonsoft.Json 9.0.1 -> System.Collections (>= 4.0.11)	MiSystems.GeoServices	C:\Work\MiSystems\MiSystems.GeoServices\MiSystems.GeoServices.csproj	
```

trejjam:
Hello @r-work and thanks.
It should be solvable in your code by declaring a dependency on `System.Xml.ReaderWriter=4.3.0` and `System.Collection=4.3.0`, or better on a higher `Newtonsoft.Json`. Which you should do anyway, there are some security issues.

At the same time, I bumped the minimal version in this library to 10.0.1, which is still vulnerable to the security issues mentioned above, but it should resolve without help.

prom3theu5:
Aye that’s right. Anything below 13.0.1 synk or another Analyzers will raise CWE-755

r-work:
Hi @trejjam, sorry had to delete my previous comment as I realized it contained internal info.

Our project does not reference `Newtonsoft.Json` as we've migrated to `System.Text.Json`.

So the issue might just be NSwag referencing an old `Newtonsoft.Json` version.

trejjam:
Original message:

> This fails if your project is net 7 and have <PlatformTarget>x64</PlatformTarget> in your csproj with the following message:
> 
>  Detected package downgrade: System.Collections from 4.3.0 to 4.0.11. Reference the package directly from the project to select a different version. 
>  `A.Project -> NSwag.AspNetCore 13.17.1 -> NSwag.Core 13.17.1 -> Newtonsoft.Json 9.0.1 -> System.Xml.ReaderWriter 4.0.11 ->  System.IO.FileSystem 4.0.1 -> runtime.win.System.IO.FileSystem 4.3.0 -> System.Collections (>= 4.3.0)`
>  `A.Project -> NSwag.AspNetCore 13.17.1 -> NSwag.Core 13.17.1 -> Newtonsoft.Json 9.0.1 -> System.Collections (>= 4.0.11)	A.Project	C:\A.Project\A.Project.csproj`

> Our project does not reference Newtonsoft.Json as we've migrated to System.Text.Json.

Since you are using NSwag you are using Newtonsoft.Json, so you should upgrade it. Lower-bound dependencies are kinda a habit here :) (They permit you to upgrade when you want, instead of forcing you)

r-work:
I meant it's not referencing it directly.
Got it, thanks.

ilias-hamdaoui:
Thank you @RicoSuter  for the release !

RicoSuter:
v13.18.0 is now released with .NET 7... please report if it works as expected

NoahStolk:
I'm able to upgrade to .NET 7 now using NSwag 13.18.0. All seems to work as expected. Thank you for the release. 🚀

RicoSuter:
@trejjam can you tell me why the BodyParametersTests are invalid in .NET 7 (Skip = "Wrong in .Net 7")?

trejjam:
Hi @RicoSuter,

The test says `When_primitive_body_parameter_has_default_value_then_it_is_optional` but it's tested to be required in .Net6. In .Net 7 `IsRequired` result is false for `BodyParameters_RequiredPrimitiveWithDefault` (as it should be, based on the test description)

trejjam:
"Correct" solution, in my opinion, is probably to remove `*before_net7` with conditional `Skip` definition `!NET7_0_OR_GREATER` and then fix the parser for .Net6. The question is if it's worthy/possible.

RicoSuter:
Ok, will think about it. Now working on the bigger v14 release itself...

PR #4221, opened 2022-11-08
Add reference assemblies metapackage to projects targeting net45

### LINKED ISSUE #4191: Building requires .NET 4.5 reference assemblies, which are no longer available from Microsoft
Cloning this repo and trying to build results in over dozens of errors if you don't have the .NET Framework 4.5 reference assemblies installed. They can be resolved by installing them, but they are [no longer provided by Microsoft](https://support.microsoft.com/en-us/topic/-net-framework-retiring-sha-1-content-9750f20d-a9ef-4d43-853f-2075f0a9d7da).

Given that .NET Framework 4.5 is out of support, is it time to drop the link to 4.5 (and potentially other out-of-support frameworks) here as well?

Pilchard123:
I believe this should allow building on machines lacking the .NET 4.5 SDK or reference assemblies.

There are plenty of warnings, but they appear to be unrelated to any of the changes I have made and would have been there if I was able to build without the changes in this PR.

Fixes #4191


savornicesei:
Hi @Pilchard123 
I think it's better to use the [meta package](https://www.nuget.org/packages/Microsoft.NETFramework.ReferenceAssemblies) which covers all versions of .NET Framework - under the hood it knows to import the right 'package' for the targeted framework.
It will also keep the .csproj file simple.

Pilchard123:
Y'know, I'd considered that and decided against it, but I can't remember why. But you're right, that's what the metapackages are *for*. I'll update.

Pilchard123:
Some projects target `net461`. I didn't have any problems building those ones even without adding the reference assemblies, but on the other hand it could be useful to add them there as well. I have the targeting packs all the way back as far as is provided in the VS installer, so presumably anyone else could get them too - for now - but I expect at some point MS will remove those too.

To future-proof the project against losing Framework references, would it be worth adding the reference assemblies to every project targeting anything that is explicitly Framework rather than Standard or Core? I felt reasonably justified in adding the 4.5 references because building was impossible without them, but I don't want to go around adding more dependencies than are necessary without the go-ahead from @RicoSuter .

savornicesei:
Actually you can add it to [Directory.Build.props](https://github.com/RicoSuter/NSwag/blob/master/src/Directory.Build.props) and it will be included automatically in all projects.

Pilchard123:
Ah, indeed you can! You learn something new every day.

Pilchard123:
Moving the PackageReferences into Directory.Build.props has made building rather flaky. I think I'll leave the PR as it is for now, with the references in each project, and let others decide whether to move them out.

RicoSuter:
v14 (current master) will only support .NET 4.6.2 going forward. Closing PR

PR #4241, opened 2022-11-21
Multiple version installation for `actions/checkout@v3`

Saibamen:
Addendum to https://github.com/RicoSuter/NSwag/pull/4220

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -37,19 +37,12 @@ jobs:
       - uses: actions/checkout@v1
       - uses: actions/setup-dotnet@v3
         with:
-          dotnet-version: 2.1.*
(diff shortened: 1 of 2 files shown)


lahma:
These files are automatically generated. Why do you want these changes applied?

Saibamen:
One action should be a little bit faster than 5

lahma:
If you run build.cmd you'll probably see the file being overwritten, so this approach won't work.

Saibamen:
73 vs 133 seconds
image

Saibamen:
You should also update `actions/checkout` to `v3` and `actions/cache` to `v3`

lahma:
This has been resolved in master

PR #4256, opened 2022-12-02
Make serializer settings static

ErikPilsits-RJW:
For System.Text.Json it is imperative to use static serializer options, see https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/configure-options?pivots=dotnet-6-0 . Aside from the performance penalty, the memory and GC pressure is enormous for a high volume client.

I believe Newtonsoft uses a static cache internally so it is not affected by a non-static settings instance, but there is no harm in making the Newtonsoft settings static as well.

This is a breaking change to users with partials that configure these settings. It would be reasonable to wrap the Newtonsoft case in a new UI settings flag, but it is catastrophic for STJ without it and should not be an option.


RicoSuter:
could you update the branch and fix conflicts?

RicoSuter:
As this will be released with v14, I'd say it's ok to make a breaking change...

ErikPilsits-RJW:
Updated to current master.

PR #4257, opened 2022-12-02
Do not throw on 204, return default response

ErikPilsits-RJW:
A 204 response is successful and it is expected that there is no body. Therefore the client should return the `default(T)` response instead of throwing a null response exception.

### DIFF
--- a/src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.liquid
+++ b/src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.liquid
@@ -322,11 +322,29 @@
 
                     var status_ = (int)response_.StatusCode;
 {%     for response in operation.Responses -%}
+{%          if response.StatusCode != "204" -%}
                     if (status_ == {{ response.StatusCode }}{% if response.CheckChunkedStatusCode %} || status_ == 206{% endif %})
                     {
                         {% template Client.Class.ProcessResponse %}
                     }
                     else
+{%          else -%}
+                    if (status_ == 204)
+                    {
+{%             if operation.HasResultType -%}
+{%                 if operation.WrapResponse and operation.UnwrappedResultType != "FileResponse" %}
+                        return new {{ ResponseClass }}<{{ operation.UnwrappedResultType }}>(status_, headers_, {{ operation.UnwrappedResultDefaultValue }});
+{%                 else -%}
+                        return {{ operation.UnwrappedResultDefaultValue }};
+{%                 endif -%}
+{%             elsif operation.WrapResponse -%}
+                        return new {{ ResponseClass }}(status_, headers_);
+{%             else -%}{% comment %} This method isn't expected to return a value. Just return. {% endcomment %}
+                        return;
+{%             endif -%}
+                    }
+                    else
+{%          endif -%}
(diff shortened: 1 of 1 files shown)


Shaddix:
Since 204 is handled in Typescript generator, it seems to make sense to handle it in C# as well

bstordrup:
Any estimate on merging this in? Could use this as well.

ErikPilsits-RJW:
@RicoSuter any chance this can make it into the next release? It's the last manual template update I maintain for our org.

ErikPilsits-RJW:
I know there are a few other open issues and PRs regarding 204 responses, but mine seems to be simple and correct for now. Future enhancements are still possible, but this has been working in our org for quite some time. I'm just getting tired of maintaining the template override with every new nswag release.

Any chance I could at least get some feedback if this is being considered?

@RicoSuter @arshrp @lahma

svartis:
I am also very keen to get this one in place

RicoSuter:
Please also add some tests covering this change and fix the build errors. Then it's fine to merge.

ErikPilsits-RJW:
> Please also add some tests covering this change and fix the build errors. Then it's fine to merge.

Can you give me some guidance on what tests you would like added and where? If there's an example to go off that should be sufficient.

RicoSuter:
We have some snapshot based tests here: 
https://github.com/RicoSuter/NSwag/tree/master/src/NSwag.CodeGeneration.TypeScript.Tests

I think we should add a sample minimal openapi with a 204 and 200 in it (etc) and generate a typescript from it and snapshot so that we have this feature under test..

ErikPilsits-RJW:
Finally getting around to looking at this. I can't build the project unfortunately, we don't use .NET 10, and your latest master is using an RC version of it. I'll do my best, hopefully the pipeline here will validate for me.

ErikPilsits-RJW:
@RicoSuter I think my test looks good. The failing builds there are other tests. Let me know if you'd like anything else. Thank you!

lahma:
`master` is green and as this is based on that, the PR has problems. Make sure you generate the output and include the snapshots as part of the PR.

ErikPilsits-RJW:
Hmm, I didn't expect that I'd have to regenerate other test's outputs. I guess that makes sense since I'm modifying a template, however that's definitely a weakness in snapshot based tests.

lahma:
> Hmm, I didn't expect that I'd have to regenerate other test's outputs. I guess that makes sense since I'm modifying a template, however that's definitely a weakness in snapshot based tests.

I believe that's the idea behind snapshot testing.

ErikPilsits-RJW:
Not being able to build this locally made that quite an effort, esp for some of those really large snapshots. Let me know if you need anything else.

@lahma @RicoSuter

chriscameron-vertexinc:
Polite bump. I've been looking forward to this enhancement for a few years now. Is there anything we can do to get it merged?

MarcoSennHaag:
@RicoSuter this would also solve a bug for me with 204-responses and SystemTextJson serialization which seems to be slightly stricter then NewtonsoftJson.There is no need trying to parse an object from an empty response. Looking forward for a merge and release!

PR #4258, opened 2022-12-02
Use all text MIME types to determine if body is a text

1taki1:
There are more MIME types than text/plain that should be handled as plain text


molinch:
It might be better to check if the mime type starts with text/ then it works for all cases and you won't need to change it ever again.

PR #4282, opened 2022-12-13
add axios abort signal

fmg-lydonchandra:


fmg-lydonchandra:
Hi, other than unit tests (adding them tomorrow), is there anything that needs to be done as part of this PR ?

fmg-lydonchandra:
hi @RicoSuter , review please :-)

BenJenkinson (changes requested):
I was going to create my own PR, but this is great, thank you!

---

The `CancelToken` type is imported from `axios` in the `File.liquid` template. 

`eslint` flags the unused import if `CancelToken` isn't used in the client.

https://github.com/RicoSuter/NSwag/blob/3ee148afbe36223db2d6c2ba65f0f5bbb67af855/src/NSwag.CodeGeneration.TypeScript/Templates/File.liquid#L47-L49

Could you apply the `{% if UseAbortSignal %}` logic there as well?

jontcarroll:
Would love to see this PR make it in soon, we are currently dealing with a lot of unwanted behavior when requests are aborted.

tomvdk:
Please check out https://github.com/RicoSuter/NSwag/pull/5202 as it was not added to the generated interfaces for both Axios and Fetch

PR #4310, opened 2023-01-11
Fix unwanted query parameter explosion in generated c# client

burgmanna:
Per default NSwag ignores if array parameters should be exploded or not.

So as of now a request containing multiple IDs generates the following:
`/users?id=3&id=4&id=5`

As per specification we do not want to specify the parameter name every time in case `explode` is set to false.
Instead we expect `/users?id=3,4,5`.


paulcurren:
Note: this doesn't fix the OpenApi 2.0 variant:
            "collectionFormat": "csv",

src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.QueryParameter.liquid
```diff
@@ -6,8 +6,13 @@ foreach (var item_ in {{ parameter.VariableName }}) { urlBuilder_.Append(System.
 urlBuilder_.Append(System.Uri.EscapeDataString("{{ parameter.Name }}") + "=").Append(System.Uri.EscapeDataString({% if parameter.IsNullable and parameter.IsRequired %}{{ parameter.VariableName }} != null ? {% endif %}{{ parameter.VariableName }}{% if parameter.IsSystemNullable %}.Value{% endif %}.ToString("{{ ParameterDateTimeFormat }}", System.Globalization.CultureInfo.InvariantCulture){% if parameter.IsNullable and parameter.IsRequired %} : "{{ QueryNullValue }}"{% endif %})).Append("&");
 {% elsif parameter.IsDate -%}
 urlBuilder_.Append(System.Uri.EscapeDataString("{{ parameter.Name }}") + "=").Append(System.Uri.EscapeDataString({% if parameter.IsNullable and parameter.IsRequired %}{{ parameter.VariableName }} != null ? {% endif %}{{ parameter.VariableName }}{% if parameter.IsSystemNullable %}.Value{% endif %}.ToString("{{ ParameterDateFormat }}", System.Globalization.CultureInfo.InvariantCulture){% if parameter.IsNullable and parameter.IsRequired %} : "{{ QueryNullValue }}"{% endif %})).Append("&");
-{% elsif parameter.IsArray -%}
+{% elsif parameter.IsArray and parameter.Explode == true -%}
```
paulcurren, line 9: Should also test for  "collectionFormat": "csv",

Mapel:
Hey, can this pull request gets considered in near future? This bug is already quite old and it is an easy fix.

Eldriann:
Hey, are any changes needed for this PR?
I also encountered this case today, and this code seems to resolve the issue. I'm using the `templateDirectory` parameter to temporarily override the file, but it would be a great bug fix to merge.

lahma:
Handled in #5016

PR #4314, opened 2023-01-16
Remove description from OpenApiResponse when serializing for Swagger v2

PatrickHofman:
Remove description from OpenApiResponse when serializing for Swagger v2 to prevent error "Structural error at paths.somepath.get.responses.304 should NOT have additional properties additionalProperty: description".

Because the C# specification of `OpenApiResponse` is adapted to comply with OpenAPI v3, the generation of Swagger v2 yields an invalid result. Specifically the "description" property of `OpenApiResponse` is not allowed with Swagger v2, but is always generated into the Swagger when calling `ToJson`.

This commit fixes this bug by ignoring the description property at all, since it is invalid anyways.

### DIFF
--- a/src/NSwag.Core.Tests/Serialization/MediaTypesSerializationTests.cs
+++ b/src/NSwag.Core.Tests/Serialization/MediaTypesSerializationTests.cs
@@ -22,7 +22,6 @@ namespace NSwag.Core.Tests.Serialization
         ""operationId"": ""foo"",
         ""responses"": {
           ""200"": {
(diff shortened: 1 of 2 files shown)


PatrickHofman:
@RicoSuter can you review the remaining failing test case? Not sure what would be needed to fix it.

RicoSuter:
Thx for the PR

RicoSuter:
Had to revert as it breaks tests...
Can you post the generated operation which is invalid?

PR #4338, opened 2023-02-13
Add support for MSBuild transitive dependency flow

### LINKED ISSUE #4026: NSwag.MSBuild support for transitive dependency flow (/buildTransitive)
### Scenario
It would be great if `NSwag.MSBuild` would support MSBuild transitive dependency flow, example:
- `Project A` has a PackageReference to `Package B`
- `Package B` has a PackageReference on `Package C`
- `Package C` has a build assets (C.targets) which is added under `/build` as well as `/buildTransitive` folders
- `Project A` should be able to consume `C.targets` from `Package C`

**Project legend in this Scenario**
- `Project A` is an `asp.net application`.
- `Project B` is a `nuget package`, containing `shared NSwag configuration files` with a PackageReference on `Project C`
- `Project C` is `NSwag.MSBuild`.

[Source of this example: `1. PackageReference with transitive dependencies to allow flowing build assets`](https://github.com/NuGet/Home/wiki/Allow-package--authors-to-define-build-assets-transitive-behavior#1-packagereference-with-transitive-dependencies-to-allow-flowing-build-assets)

**Official docs about /buildTransitive**
- [Nuget Spec](https://github.com/NuGet/Home/wiki/Allow-package--authors-to-define-build-assets-transitive-behavior)
- [Nuget package docs](https://docs.microsoft.com/en-us/nuget/create-packages/creating-a-package#from-a-convention-based-working-directory)

### What doesn't work?
- `Project B` contains a custom MSBuild file which uses MSBuild properties from [NSwag.MSBuild](https://github.com/RicoSuter/NSwag/blob/master/src/NSwag.MSBuild/NSwag.MSBuild.props)
- `NSwag.MSBuild` properties are working in `Project B` because it is imported as expected by nuget.
- `NSwag.MSBuild` properties are not working in `Project C` because MSBuild files are not imported by nuget because `buildTransitive` is missing

### What is required to get it working?
- [NSwag.MSBuild](https://github.com/RicoSuter/NSwag/blob/master/src/NSwag.MSBuild/NSwag.MSBuild.props) MSBuild files are included in the nuget package by folder convention `/buildTransitive` folders.

DanielTheCoder:
Fixes #4026

### DIFF
--- a/src/NSwag.MSBuild/NSwag.MSBuild.nuspec
+++ b/src/NSwag.MSBuild/NSwag.MSBuild.nuspec
@@ -20,6 +20,7 @@
   <files>
     <file src="NSwag.MSBuild.props" target="build" />
     <file src="NSwag.MSBuild.props" target="buildCrossTargeting" />
+    <file src="NSwag.MSBuild.props" target="buildTransitive" />
 
     <file src="..\NSwag.Console\bin\Release\net461\*" target="tools/Win" />
     <file src="..\NSwag.Console.x86\bin\Release\net461\NSwag.AspNetCore.Launcher.x86.exe" target="tools/Win" />


DanielTheCoder:
Check failed because of last commit to master
=> https://github.com/RicoSuter/NSwag/commit/3ee148afbe36223db2d6c2ba65f0f5bbb67af855

Issue #4373, opened 2023-03-22
Unwrap ValueTask<T> return types, same as Task<T>, unwrap (Value)Task<ActionResult<T>> and unify unwrapping for consistency


alasdaircs:
NSwag generates incorrect Swagger/OpenAPI doc for (an async) controller action that returns ValueTask<T>. It should be unwrapped as returning T, same as Task<T>. 

(Value)Task<ActionResult<T>> is also a valid return type for an async controller action, and requires double unwrapping which is not currently implemented.

I have forked the repo and made the required code changes, including new tests. I'll incorporate this Issue Id in the commit.

PR #4374, opened 2023-03-22
Unwrap ValueTask<T> return types

alasdaircs:
Unwrap ValueTask<T> return types, same as Task<T>
Unwrap (Value)Task<ActionResult<T>>
Unify unwrapping for consistency
Tests included
See #4373

### DIFF
--- a/src/NSwag.Generation/GenericResultWrapperTypes.cs
+++ b/src/NSwag.Generation/GenericResultWrapperTypes.cs
@@ -3,27 +3,23 @@ using System.Linq;
 
 namespace NSwag.Generation
 {
+    internal static class GenericResultWrapperTypes
+    {
+        internal static bool IsGenericWrapperType( string typeName )
+            =>
+                typeName == "Task`1" ||
+                typeName == "ValueTask`1" ||
+                typeName == "JsonResult`1" ||
+                typeName == "ActionResult`1"
+            ;
 
-	internal static class GenericResultWrapperTypes
-	{
-		private readonly static String[] WrapperTypeNames =
-		{
-			"Task`1",
-			"ValueTask`1",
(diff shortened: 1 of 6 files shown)


src/NSwag.Generation/GenericResultWrapperTypes.cs
```diff
@@ -0,0 +1,29 @@
...
+		private readonly static String[] WrapperTypeNames =
+		{
+			"Task`1",
+			"ValueTask`1",
+			"JsonResult`1",
+			"ActionResult`1"
+		};
+
+		internal static bool IsGenericWrapperType(string typeName)
+			=> WrapperTypeNames.Contains( typeName );
```
lahma, line 18: this is quite slow, probably better to just do 
``` 
=> typeName == "Task`1" || typeName == "ValueTask`1" || typeName== "JsonResult`1" || typeName == "ActionResult`1" 
```

alasdaircs: I have no idea what the difference is in practice - I tend to write simple stuff like this more for readability than marginal gains in execution time but I guess it all adds up. I did some searching and found this [Performance insight for Enumerable.Contains](https://learn.microsoft.com/en-us/visualstudio/profiling/perf-insights-contains) which points out the doing this with a HashSet<T> would make it O(1), however for this short list of only fours items I bet your code smaller and possibly faster, so I'm happy to make your suggested change.

alasdaircs:
I believe the failing checks are failing away from my changes - I think master I based on at 3ee148af had the same failing tests.

lahma:
Yes I think we need to wait for some PRs to land into master before many of the builds succeed...

lahma (approved):

Issue #4379, opened 2023-03-29
Required form parameters are not generated properly in OpenAPI 3.0 and C# client


ddaspit:
Parameters in a controller action that are marked with the attributes `[FromForm]` and `[BindRequired]` are not properly generated in OpenAPI 3.0. The parameters should be added to the `required` section of the body schema. This works with Swagger 2.0.

In addition, even if the parameters are marked correctly as required in OpenAPI 3.0, the C# client is not generated correctly when `generateOptionalParameters` is set to `true`. Required form parameters are treated as optional and assigned a default value.

Steps to reproduce:

Setup NSwag to use OpenAPI 3.0 in Startup.cs:
```cs
public class Startup
{
    public Startup(IConfiguration configuration)
    {
        Configuration = configuration;
    }

    public IConfiguration Configuration { get; }

    // This method gets called by the runtime. Use this method to add services to the container.
    public void ConfigureServices(IServiceCollection services)
    {
        services.AddControllers();
        services.AddOpenApiDocument(document =>
        {
            document.SchemaType = SchemaType.OpenApi3;
            document.DocumentName = "test";
            document.Title = "testing";
        });
    }

    // This method gets called by the runtime. Use this method to configure the HTTP request pipeline.
    public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
    {
        if (env.IsDevelopment())
        {
            app.UseDeveloperExceptionPage();
        }
        
        app.UseHttpsRedirection();

        app.UseRouting();
        app.UseOpenApi();
        app.UseSwaggerUi3();
        app.UseAuthorization();

        app.UseEndpoints(endpoints =>
        {
            endpoints.MapControllers();
        });
    }
}
```

Create a controller class with a required form parameter:
```cs
[ApiController]
[Route("api/[controller]")]
public class FormController : Controller
{
    [HttpPost("Action")]
    public ActionResult Action([FromForm, BindRequired] string test)
    {
        return Ok();
    }
}
```

The generated OpenAPI spec is:
```json
"paths": {
  "/Action": {
    "post": {
      "operationId": "Form_Action",
      "requestBody": {
        "content": {
          "multipart/form-data": {
            "schema": {
              "type": "object",
              "properties": {
                "test": {
                  "type": "string",
                  "nullable": true
                }
              }
            }
          }
        }
      }
    }
  }
```

The expected spec is:
```json
"paths": {
  "/Action": {
    "post": {
      "operationId": "Form_Action",
[...]

PR #4380, opened 2023-03-29
Correctly preserve IsRequired for form parameters in OAPI 3.0 (#4379)

ddaspit:
- Set `IsRequired`, `Description`, `Default`, and `Example` when creating OAPI3.0 form parameters in `OperationParameterProcessor`
- Set `IsRequired` when getting OAPI 3.0 form data parameters in `OperationModelBase`
- See #4379


lahma:
@ddaspit can you please rebase this?

ddaspit:
It has been a while since I've looked at this. I will see if I can rebase it.

Issue #2863, opened 2020-05-31
Duplicate return code declaration causes nasty exception 


alalloo:
A controller might have several reasons to return a bad request status and in our project had them declared like below which is handled OK by Swagger. But switching to NSwag was problematic since it caused an exception and it was really hard knowing what NSwag was unhappy about. I had to clone nswag repo and debug it to find what was the problem (which then could trivially be corrected). 

A proper error message to user would have saved a couple of hours. 

```
	/// <returns>Updated insured object.</returns>
	/// <response code="400">Bad request, missing information or similar.</response>
	/// <response code="400">Unknown category.</response>
	/// <response code="409">Attempt to update read-only properties.</response>
```
	
The exception
```
System.InvalidOperationException: Sequence contains more than one matching element

Runtime: NetCore31
   at System.Linq.ThrowHelper.ThrowMoreThanOneMatchException()
   at System.Linq.Enumerable.SingleOrDefault[TSource](IEnumerable`1 source, Func`2 predicate)
   at NSwag.Generation.Processors.OperationResponseProcessorBase.GetResponseXmlDocsElement(MethodInfo methodInfo, String responseCode) in C:\projects\nswag\src\NSwag.Generation\Processors\OperationResponseProcessorBase.cs:line 92
   at NSwag.Generation.Processors.OperationResponseProcessorBase.UpdateResponseDescription(OperationProcessorContext operationProcessorContext) in C:\projects\nswag\src\NSwag.Generation\Processors\OperationResponseProcessorBase.cs:line 59
   at NSwag.Generation.WebApi.Processors.OperationResponseProcessor.Process(OperationProcessorContext context) in C:\projects\nswag\src\NSwag.Generation.WebApi\Processors\OperationResponseProcessor.cs:line 52
   at NSwag.Generation.WebApi.WebApiOpenApiDocumentGenerator.RunOperationProcessors(OpenApiDocument document, Type controllerType, MethodInfo methodInfo, OpenApiOperationDescription operationDescription, List`1 allOperations, OpenApiDocumentGenerator swaggerGenerator, OpenApiSchemaResolver schemaResolver) in C:\projects\nswag\src\NSwag.Generation.WebApi\WebApiOpenApiDocumentGenerator.cs:line 241
[...]

PR #4388, opened 2023-04-04
Descriptive exception for duplicate XML doc line (#2863)

aviita:
Duplicate response code in XML documentation response xml tag causes somewhat hard to debug exception, which is now made more descriptive.

Did not yet figure out how to make an unit test for this one or if some other other exception type would be more suitable than InvalidOperationException.

Bug: Duplicate return code declaration causes nasty exception
id: #2863


aviita:
Thought I'll just make a quick PR on this although I do not yet have time to check e.g. unit testing practices or how to test this even manually. It anyways seems it is quite easy to create a very descriptive error message for this specific exception.

PR #4422, opened 2023-05-11
Disable warning CS8604 "Possible null reference argument for parameter" in generated code

NoahStolk:
We'd like to use `<WarningsAsErrors>nullable</WarningsAsErrors>` in our code base to enforce strict NRT checks. However, a lot of `CS8604` warnings are emitted in generated NSwag code, causing build errors when enabling this build property.

image

Seeing that `CS8603` is already disabled, it seems reasonable to disable `CS8604` as well.

An alternative would be to disable NRT warnings in generated code altogether, using `#nullable disable warnings`, which disables [these warnings](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-messages/nullable-warnings) if I'm not mistaken.


RicoSuter:
Thanks for the PR.

PR #4471, opened 2023-07-02
Update Client.Class.liquid

### LINKED ISSUE #4466: System.InvalidOperationException: ValueFactory attempted to access the Value property of this instance.
nswag in system,text,json mode is not thread safety!
because in constructor of generated source use
 _settings = new System.Lazy<System.Text.Json.JsonSerializerOptions>(CreateSerializerSettings);
instead of
 _settings = new System.Lazy<System.Text.Json.JsonSerializerOptions>(CreateSerializerSettings, true);

alireza2125:
fix #4466


RicoSuter:
Thanks for the fix

PR #4227, opened 2022-11-13
Net462update

geckogeckogecko:
As .NET 4.6.1 is no longer supported, the changes made update the project to .NET 4.6.2, and remove other references to outdated .NETCore versions.
Further reading here: https://devblogs.microsoft.com/dotnet/net-framework-4-5-2-4-6-4-6-1-will-reach-end-of-support-on-april-26-2022/

PR #4472, opened 2023-07-03
Add multi-file output to NSwagStudio

TheCakeMonster:
This PR adds support for splitting the generated output into multiple, separate files. Tested (and in use) for C#; whether this additionally supports other languages has not been tested.

This is built upon PR #4227, which was required so that we were not required to install unsupported runtimes on our machines. Therefore, it includes dropping support for .NET Core 2.1 and 3.0, and .NET 5.0, as well as bumping the minimum supported version of .NET Framework to .NET 4.6.2

Credit for this change goes to @geckogeckogecko who did all of the detailed work while at PDL.

Addresses the epic #1398


soroshsabz (changes requested):
I think this PR could not merge before #4227

Bykiev:
Any chance it'll merged?

Minecraftschurli:
@Bykiev I think it needs a rebase

Issue #4392, opened 2023-04-07
Generates different clients based on ref variable in multipart 


fedotovsi:
I noticed that last version generates different client depending whether or not multipart request defined as $ref or it is directly.
So two identical from functional perspective api definition produce different client code. see below the code and api description.
I wonder if this is some settings missing or is it intended behavior ? 
so rest api looks like : 
```
{
  "openapi": "3.0.1",
  "info": {
    "title": "Api",
    "version": "1.0"
  },
  "paths": {    
    "/api/v1": {
      "post": {        
        "requestBody": {
          "content": {
            "multipart/form-data": {
              "schema": {
                "$ref": "#/components/schemas/uploadBody"
              }
            }
          }
        },
        "responses": {
          "200": {
            "description": "Success"            
            }
          }
        }
      }
    },
  "components": {
    "schemas": {
      "uploadBody": {        
        "type": "object",
        "properties": {
          "file": {
            "type": "string",
            "format": "binary"
          }
        }
      }      
    }
  }
}
```
If instead of "$ref": "#/components/schemas/uploadBody" to put directly schema from components it produces 2 different version of the code client : 
```
public virtual async System.Threading.Tasks.Task V1Async(System.IO.Stream body) // with ref
public virtual async System.Threading.Tasks.Task V1Async(FileParameter file) //directly  
```
I am using nswag to generate c# code so .NET60 
Let me know if you need my nswag.json fully

PR #4476, opened 2023-07-06
Fix for multipart/formdata schema bug.

### LINKED ISSUE #3524: "multipart/form-data" Request Body Schema Inconsistency With Inline Definition Versus Reference
If I have the following definition for the request body of an operation:

      requestBody:
        description: Contact data
        content:
          multipart/form-data:
            schema:
              type: object
              properties:
                emailAddress:
                  type: string
                firstName:
                  type: string
                lastName:
                  type: string

... then the OperationModelBase will contain a parameter for each of the properties.

However, if I have the following (functionally equivalent) definition for the request body:

      requestBody:
        description: Contact data
        content:
          multipart/form-data:
            schema:
              $ref: '#/components/schemas/Contact'

... then the OperationModelBase will contain a body parameter, of a type which has a property for each of the properties.

I am using versions 13.11.3.

Is it correct that these two scenarios are treated differently?
(they behave differently because line 342 in OperationModelBase uses .Schema rather than .ActualSchema.

May I also please ask why "multipart/form-data" is treated this way? As opposed to other form request content types, like "application/x-www-form-urlencoded" or "multipart/mixed"?

bobend:
Fixes #3524 and #4392

### DIFF
--- a/src/NSwag.CodeGeneration.CSharp.Tests/BinaryTests.cs
+++ b/src/NSwag.CodeGeneration.CSharp.Tests/BinaryTests.cs
@@ -6,6 +6,7 @@ namespace NSwag.CodeGeneration.CSharp.Tests
 {
     public class BinaryTests
     {
+
         [Fact]
         public async Task When_body_is_binary_then_stream_is_used_as_parameter_in_CSharp()
         {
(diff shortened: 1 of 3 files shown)


vpetrusevici:
@RicoSuter looks like a simple fix. can you merge it please?

PR #4502, opened 2023-08-30
Add net8.0 target framework

cristipufu:
// Will update as soon as .net 8 gets a stable release

// Need to update Nuke.Common to support net8 https://github.com/nuke-build/nuke/pull/1247


RicoSuter:
Would be great if you could update it against latest master

RicoSuter:
I'd say the build project can stay on .net 7 for now, just the needed libraries etc need to be updated for now.

cristipufu:
@RicoSuter @lahma done

lahma:
nit: if you look at the GitHub diff, you can see that there are multiple odd indents present, maybe tabs vs spaces or something?

cristipufu:
@RicoSuter @lahma can you please approve the latest workflow?

cristipufu:
@lahma 
I don't think that the net8-rc1 sdk is installed in the base image, do I need to add these back or we just wait for the final release?
```
build.yaml/pr.yaml
- uses: actions/setup-dotnet@v3
        with:
          dotnet-version: 8.0.100-rc.1.23463.5
```
```
Build.CI.GitHubActions.cs
// only need to list the ones that are missing from default image
newSteps.Insert(0, new GitHubActionsSetupDotNetStep(new[] 
{
    "8.0.100-rc.1.23463.5"
}));
```

Any other suggestions?

lahma:
IMO it's a good idea to add them, NUKE setup was cleaned earlier to not include obsolete ones bur for NET 8 is a different story.

Getting builds out against RC would help to validate and find possible problems.

cristipufu:
@lahma done, we can give it another try

lahma (approved):
LGTM

lahma:
I found one place that might still need alteration:

https://github.com/RicoSuter/NSwag/blob/c24ed12d91fb0d717ccdd455be972d88d0ced534/src/NSwag.Commands/Commands/Generation/AspNetCore/AspNetCoreToOpenApiCommand.cs#L146-L148

Because it's under pre-processor directive which should only hit modern .NET I wonder whether the if is needed at all.

RicoSuter:
I think we can remove this if completely

src/NSwag.AspNetCore/Middlewares/RedirectToIndexMiddleware.cs
```diff
@@ -40,7 +40,7 @@ public async Task Invoke(HttpContext context)
 
                 var suffix = !string.IsNullOrWhiteSpace(_swaggerRoute) ? "?url=" + _transformToExternal(_swaggerRoute, context.Request) : "";
                 var path = _transformToExternal(_swaggerUiRoute, context.Request);
-                context.Response.Headers.Add("Location", (path != "/" ? path : "") + "/index.html" + suffix);
+                context.Response.Headers.Append("Location", (path != "/" ? path : "") + "/index.html" + suffix);
```
RicoSuter, line 43: Did you have a problem with the Add()?

RicoSuter:
> I think we can remove this if completely

I think the whole reasoning behind the two ifs (.netframework, .netcore) is that you get a "nice" error when you try to load a .netfw project in .net core and vice versa, which you would not get anymore (probably it fails differently now :))

RicoSuter:
... updated with my proposal (keep ifs to be more safe but make it more generic in .net "core" case so we do not have to touch it again).

src/NSwag.AspNetCore/Middlewares/RedirectToIndexMiddleware.cs
cristipufu: VS preview complains with https://learn.microsoft.com/en-us/aspnet/core/diagnostics/asp0019?view=aspnetcore-8.0

Issue #4507, opened 2023-09-05
Multipart/Form-Data parameters not populated correctly when using $ref


drinehim:
When using a $ref for the body of a multipart/form-data request the parameters are not correct. An `IFormFile body` is added instead of any of the actual parameters.

[test.yaml.txt](https://github.com/RicoSuter/NSwag/files/12527710/test.yaml.txt)
[TestController.cs.txt](https://github.com/RicoSuter/NSwag/files/12527712/TestController.cs.txt)

PR #4508, opened 2023-09-05
Populate multiforms properties from reference #4507

drinehim:


src/NSwag.CodeGeneration/Models/OperationModelBase.cs
```diff
@@ -350,9 +350,14 @@ protected IList<OpenApiParameter> GetActualParameters()
                 _operation?.RequestBody?.Content?.ContainsKey("multipart/form-data") == true ?
                 _operation.RequestBody.Content["multipart/form-data"]?.Schema : null;
 
-            if (formDataSchema != null && formDataSchema.ActualProperties.Count > 0)
+            var formDataProperties = formDataSchema?.ActualProperties.Count > 0
+                ? formDataSchema.ActualProperties
+                : formDataSchema?.Reference?.ActualProperties.Count > 0
```
RicoSuter, line 355: Do you need reference check here? I'd expect that ActualProperties resolves this automatically, no?

drinehim: It does not. When running the `When_body_is_binary_with_additional_data_via_component_ref_then_all_parameters_supplied_in_CSharp_ASPNETCore` test I added `formDataSchema?.ActualProperties.Count` is 0 but `formDataSchema?.Reference?.ActualProperties.Count` is 2.

drinehim: ActualSchema.ActualProperties works though. I will push an update to do that.

Issue #2839, opened 2020-05-11
'Missing ApiException' compiler error


dgmagno:
Currently I have a project that uses multiple openapi.json v3 files. We used the official openapi tools described in this link: [https://docs.microsoft.com/en-us/aspnet/core/web-api/microsoft.dotnet-openapi?view=aspnetcore-3.1](url)

Because of the following code the generated c# client do not compile:

```
<ItemGroup>
  <!-- @(CurrentOpenApiReference) item group will never contain more than one item. -->
  <CurrentOpenApiReference>
    <Command>$(_NSwagCommand) openapi2csclient /className:%(ClassName) /namespace:%(Namespace)</Command>
  </CurrentOpenApiReference>
  <CurrentOpenApiReference>
    <Command Condition="! %(FirstForGenerator)">%(Command) /GenerateExceptionClasses:false</Command>
  </CurrentOpenApiReference>
  <CurrentOpenApiReference>
    <Command>%(Command) /input:"%(FullPath)" /output:"%(OutputPath)" %(Options)</Command>
  </CurrentOpenApiReference>
</ItemGroup>
```

Why did you make this design decision of not generating the Exception code for the remaining clients? The commented affirmation of only one item is not true.

I managed it to work by adding a `AdditionalNamespaceUsages` options but for me it feels like a hack.

Below I added a sample of a csproj to reproduce the issue:

```
<Project Sdk="Microsoft.NET.Sdk">
  
  <PropertyGroup>
    <TargetFramework>netcoreapp3.1</TargetFramework>
  </PropertyGroup>
      
  <ItemGroup> 
    <PackageReference Include="Newtonsoft.Json" Version="12.0.2" /> 
    <PackageReference Include="NSwag.ApiDescription.Client" Version="13.5.0" />
  </ItemGroup>
  
  <ItemGroup>
    <OpenApiReference Include=".\OpenApi1.json">
      <Namespace>Integration.ExternalApi1</Namespace>
    </OpenApiReference>
    
    <OpenApiReference Include=".\OpenApi2.json">
      <Namespace>Integration.ExternalApi2</Namespace>
      <!-- Workaround to make the generated code compile -->
      <!--<Options>/AdditionalNamespaceUsages:Integration.ExternalApi1</Options>-->
    </OpenApiReference>
  </ItemGroup>
  
</Project>
```

KevinOneSixDotNet:
[...]

PR #4510, opened 2023-09-06
updated ExceptionClass name

vixero-dev:
Lazy fix for #2839

### DIFF
--- a/src/NSwag.CodeGeneration.CSharp/CSharpClientGeneratorSettings.cs
+++ b/src/NSwag.CodeGeneration.CSharp/CSharpClientGeneratorSettings.cs
@@ -17,7 +17,7 @@ namespace NSwag.CodeGeneration.CSharp
             ClassName = "{controller}Client";
 
             GenerateExceptionClasses = true;
-            ExceptionClass = "ApiException";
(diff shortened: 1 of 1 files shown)


RicoSuter:
Cant you just change this in your project settings instead of changing the default?

vixero-dev:
I don't fully understand what project settings you mean. Could you please be more specific?

RicoSuter:
image

vixero-dev:
Oh, I see. Didn't know that was an option.

I was using Visual Studio to add a Connected Service and it doesn't have that option.

image

RicoSuter:
Try "/ExceptionClass:{ClassName}Exception"

RicoSuter:
options can be found here
https://github.com/RicoSuter/NSwag/blob/master/src/NSwag.Commands/Commands/CodeGeneration/OpenApiToCSharpClientCommand.cs

vixero-dev:
Thank you, I will have a look at it.

Would you like me to update the PR?

vixero-dev:
@RicoSuter I tried the proposed solution (`/ExceptionClass:{ClassName}Exception`) and now the Exception class is not generated at all.

PR #4534, opened 2023-10-03
headers dictionary with IgnoreCase comparer

ju5tas:
I'm facing a problem when a generated nswag client has a Headers field that is case sensitive.
I suggest adding StringComparer.OrdinalIgnoreCase since HTTP2 requires lower case headers (https://datatracker.ietf.org/doc/html/rfc7540#section-8.1.2).

### DIFF
--- a/src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.liquid
+++ b/src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.liquid
@@ -301,7 +301,7 @@
                 var disposeResponse_ = true;
                 try
                 {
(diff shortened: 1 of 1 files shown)


olegd-superoffice:
Correct me if I'm wrong but from what I can see, generated client is building a dictionary by combining response headers and response context headers, and then this dictionary is only used as a parameter when throwing exception. If case insensitive comparison is used in `ToDictionary()` call it is just increases chances of `ArgumentException` on duplicate keys. So I'm not sure which problem this would solve.

Can you provide more details about a problem you are facing?

lahma:
I don't see this fix as effective, maybe propose alternative solution with some test cases.

Issue #2824, opened 2020-04-27
NSwag.MSBuild has security vulnerabilities


AndrewKnieriem:
nswag.msbuild.13.4.2.nupkg has vulnerabilities CVE-2019-1075, CVE-2019-0548, CVE-2019-0564, and CVE-2017-0247. 

Related to [Microsoft Security Advisory 4021279: Vulnerabilities in .NET Core, ASP.NET Core Could Allow Elevation of Privilege](https://github.com/aspnet/Announcements/issues/239).

These vulnerabilities were identified through my company's license of WhiteSource, and as such we can't use this library in production with these issues in place.

RicoSuter:
Its probably because it still supports older/outdated .net versions

Numpsy:
Hi,

I work on some projects that use NSwag.MSBuild (13.9.4 at present), and that has recently started to be run through Whitesource, and has similar warnings.

I'm not convinced there is an actual issue here (given that the tool is being run at build time to generate C# clients for use in applications that have no ASP code or hosting at all at runtime, so I don't how the CVEs about runtime denial of service issues can occur?)

However:

Looking at the Microsoft.AspNetCore.* binaries installed by the MSBuild package, it looks to me like the 2.2 binaries installed with the .NET version of the tool are older versions than the 2.2 binaries installed with the .NETCore2.2 version of the tool?

Looking at the change history, the netcoreapp2.2 build was updated in https://github.com/RicoSuter/NSwag/commit/6b35c3e5f5685108673726e0cb687c9892b36e84 and that mentions CVE-2019-1075 (which is one of the issues that Whitesource mentions), so I wonder if it would be possible to update the .net461 reference at

https://github.com/RicoSuter/NSwag/blob/1fb4a1af8785d375684b7c16be2ad76d40b422e5/src/NSwag.Commands/NSwag.Commands.csproj#L59

in a similar way?

These are the suggestions from Whitesource for reference

![image](https://user-images.githubusercontent.com/1178570/105518719-850b5f00-5cd0-11eb-9bbc-b5c2158e7690.png)

gbenthomas:
Hi @AndrewKnieriem and @Numpsy, I ran into the same issue with WhiteSource. May I ask what you ended up doing to work around this?

@RicoSuter Any plans on updating the dependencies as @Numpsy mentioned?

Thanks!

SafeNuts:
Hello!

Any updates?

Thanks!

Numpsy:
[...]

PR #4537, opened 2023-10-08
Remove support for ASP.NET Core on full framework in AspNetCoreToOpenApiGenerator

lahma:
ASP.NET Core hasn't been supported with full framework for a long time and currently causes dependency to ancient and vulnerable ASP.NET Core 2.2 hosting bundle. `HostApplication` type can now conditionally compiled when `!NETFRAMEWORK`.

### DIFF
--- a/src/NSwag.Commands/Commands/Generation/AspNetCore/AspNetCoreToOpenApiCommand.cs
+++ b/src/NSwag.Commands/Commands/Generation/AspNetCore/AspNetCoreToOpenApiCommand.cs
@@ -17,7 +17,7 @@ using NConsole;
 using Newtonsoft.Json;
 using NSwag.Generation;
 
-#if NETCOREAPP || NETSTANDARD
+#if !NETFRAMEWORK
 using System.Runtime.Loader;
 #endif
 
--- a/src/NSwag.Commands/Commands/Generation/AspNetCore/AspNetCoreToOpenApiGeneratorCommandEntryPoint.cs
+++ b/src/NSwag.Commands/Commands/Generation/AspNetCore/AspNetCoreToOpenApiGeneratorCommandEntryPoint.cs
@@ -6,6 +6,7 @@
 // <author>Rico Suter, mail@rsuter.com</author>
 //-----------------------------------------------------------------------
 
+using System;
 using System.IO;
 using System.Reflection;
 using Newtonsoft.Json;
@@ -19,6 +20,9 @@ namespace NSwag.Commands.Generation.AspNetCore
     {
         public static void Process(string commandContent, string outputFile, string applicationName)
         {
+#if NETFRAMEWORK
+            throw new NotSupportedException("ASP.NET Core running on Full Framework is not supported");
+#else
             var command = JsonConvert.DeserializeObject<AspNetCoreToOpenApiCommand>(commandContent);
             var previousWorkingDirectory = command.ChangeWorkingDirectoryAndSetAspNetCoreEnvironment();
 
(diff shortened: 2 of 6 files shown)


Numpsy:
Refs #2824 I think (Mend thinks the current v14 preview suffers from https://www.mend.io/vulnerability-database/CVE-2019-1075, whicn alas creates noise whether it actually does or not)

paulomorgado:
From [.NET and .NET Core Support Policy](https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core):

> ### ASP.NET Core 2.1 on .NET Framework
> Support for ASP.NET Core 2.1 on .NET Framework matches the [ASP.NET Support](https://dotnet.microsoft.com/en-us/platform/support/policy/aspnet) policy for other package-based ASP.NET frameworks. The complete list of packages covered by this policy can be seen in [ASP.NET Core 2.1 Supported Packages](https://dotnet.microsoft.com/en-us/platform/support/policy/aspnet/2.1-packages).

ASP.NET Core 2.1 on supported versions of .NET Framework is supported, but ASP.NET Core 2.2 isn't.

Neither is .NET Core 2.1 runtime.

olegd-superoffice:
I actually use NSwag with Asp.Net Core 2.1 application on .Net Framework 4 which is a combination supported by Microsoft and will be supported for long time. This problem needs to be solved in a different way, by referencing supported versions of packages instead of deprecated ones.

Numpsy:
> I actually use NSwag with Asp.Net Core 2.1 application on .Net Framework 4 which is a combination supported by Microsoft and will be supported for long time. This problem needs to be solved in a different way, by referencing supported versions of packages instead of deprecated ones.

Are we talking about the server side or a client side?

The security complaints we see are about NSwag.MSuild being used to generate clients (which have nothing to do with ASP of any flavour) from an openapi spec, *not* from NSwag.AspNetCore which is used on the server to generate said specifications.

olegd-superoffice:
> Are we talking about the server side or a client side?

Both. And both are supported by Microsoft. This pull request just removes support for .Net Framework instead of fixing it.
And NSwag.MSBuild does depend on some of Microsoft.AspNetCore packages. Look, for example, into [tools/Net80 folder](https://nuget.info/packages/NSwag.MSBuild/14.0.0-preview008).

paulomorgado:
> > I actually use NSwag with Asp.Net Core 2.1 application on .Net Framework 4 which is a combination supported by Microsoft and will be supported for long time. This problem needs to be solved in a different way, by referencing supported versions of packages instead of deprecated ones.
> 
> Are we talking about the server side or a client side?
> 
> The security complaints we see are about NSwag.MSBuild being used to generate clients (which have nothing to do with ASP of any flavour) from an openapi spec, _not_ from NSwag.AspNetCore which is used on the server to generate said specifications.

@Numpsy, what security concerns are those? Are there any issues open on that?

olegd-superoffice:
#4561 should fix [CVE-2019-1075](https://github.com/aspnet/Announcements/issues/373) without removing support for .Net Framework.

Numpsy:
> @Numpsy, what security concerns are those? Are there any issues open on that?

https://github.com/RicoSuter/NSwag/issues/2824

Numpsy:
> #4561 should fix [CVE-2019-1075](https://github.com/aspnet/Announcements/issues/373) without removing support for .Net Framework.

yeah, but those were only updated in https://github.com/RicoSuter/NSwag/commit/1eae91f56b1bd7de8719ccc6d10bb925ab402814 and that fixed a couple of other security warnings.

There are newer versions of some of the lower level libs about though, so maybe updating those as well would avoid the complaints.

lahma:
I'm cool with that 👍🏻

PR #4541, opened 2023-10-17
Sanitize BaseUrl on write.

paulomorgado:
Currently, `BaseUrl` is sanitized every time it is used:

```
urlBuilder_.Append({% if UseBaseUrl %}BaseUrl != null ? BaseUrl.TrimEnd('/') : "").Append("/{% else %}"{% endif %}{{ operation.Path }}{% if operation.HasQueryParameters %}?{% endif %}");
```

By sanitizing its value when set:

```
{% if UseBaseUrl and GenerateBaseUrlProperty -%}
    public string BaseUrl
    {
        get { return _baseUrl; }
        set
        {
            _baseUrl = value;
            if (!string.IsNullOrEmpty(_baseUrl) && !_baseUrl.EndsWith("/"))
                baseUrl += '/';
        }
    }

{% endif -%}
```

it can be used without changes:

```
        {% if UseBaseUrl %}if (!string.IsNullOrEmpty(BaseUrl)) urlBuilder_.Append(BaseUrl);{% endif %}
        urlBuilder_.Append("{{ operation.Path }}{% if operation.HasQueryParameters %}?{% endif %}");
```

This is a runtime performance improvement for those using base URL.

### DIFF
--- a/src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.liquid
+++ b/src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.liquid
@@ -3,7 +3,7 @@
 {{ ClientClassAccessModifier }} partial class {{ Class }} {% if HasBaseType %}: {% endif %}{% if HasBaseClass %}{{ BaseClass }}{% if GenerateClientInterfaces %}, {% endif %}{% endif %}{% if GenerateClientInterfaces %}I{{ Class }}{% endif %}
 {
 {% if UseBaseUrl and GenerateBaseUrlProperty -%}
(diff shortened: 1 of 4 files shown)


paulomorgado:
The null or empty test is useless.

It could be:

```
        urlBuilder_{% if UseBaseUrl %}.Append(BaseUrl);{% endif %}.Append("{{ operation.Path }}{% if operation.HasQueryParameters %}?{% endif %}");
```

paulomorgado:
Sorry about the bugs on this PR, everyone.

@MaximeMorin-Devolutions, I've created #4570.  Can you check the changes and propose changes for anything else you found?

PR #4553, opened 2023-10-24
Remove unnecessary string creations from OpenApiDocumentMiddleware

paulomorgado:
Every time methods like `string.Trim()` or `string.ToLowerInvariant()` are invoked on an instance of a `string`, an new `string` instance is created.

Being a middleware, `OpenApiDocumentMiddleware` is on the hot path of every request.

This PR removes those string creations from `OpenApiDocumentMiddleware`.

To keep the same logic as before, `string.AsSpan()´ could be used, but that doesn't work as well on .NET Framework as in .NET, performance-wise.

### DIFF
--- a/src/NSwag.AspNet.Owin/Middlewares/OpenApiDocumentMiddleware.cs
+++ b/src/NSwag.AspNet.Owin/Middlewares/OpenApiDocumentMiddleware.cs
@@ -34,7 +34,7 @@ namespace NSwag.AspNet.Owin.Middlewares
         public OpenApiDocumentMiddleware(OwinMiddleware next, string path, IEnumerable<Type> controllerTypes, SwaggerSettings<WebApiOpenApiDocumentGeneratorSettings> settings)
             : base(next)
         {
-            _path = path;
+            _path = path.StartsWith("/") ? path : '/' + path;
             _controllerTypes = controllerTypes;
             _settings = settings;
         }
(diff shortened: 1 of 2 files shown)


src/NSwag.AspNet.Owin/Middlewares/OpenApiDocumentMiddleware.cs
```diff
@@ -44,7 +44,7 @@ public OpenApiDocumentMiddleware(OwinMiddleware next, string path, IEnumerable<T
         /// <returns>The task.</returns>
         public override async Task Invoke(IOwinContext context)
         {
-            if (context.Request.Path.HasValue && string.Equals(context.Request.Path.Value.Trim('/'), _path.Trim('/'), StringComparison.OrdinalIgnoreCase))
+            if (context.Request.Path.HasValue && string.Equals(context.Request.Path.Value, _path, StringComparison.OrdinalIgnoreCase))
```
olegd-superoffice, line 47: What if `context.Request.Path.Value` doesn't start with `'/'` here?

paulomorgado: Is that possible?

Either way, creating new strings is not the solution.

olegd-superoffice: I think it is possible in general (AFAIR it can be empty string for example). I do agree with you, low allocation code is good, I'm just saying that this seem to be changing comparison logic.

RicoSuter: I think this looks fine (did you test it locally)?
I'd say we merge this... ok?

paulomorgado: I only tested on ASP.NET Core.

But the path is supposed to always start with `7`, as far as I recall.

RicoSuter: Ok then..

RicoSuter:
Thanks for the pr

PR #4565, opened 2023-11-01
Add NSwag specific metadata to OpenApiReference OpenApiProjectReference items and use it to build the command line.

paulomorgado:
This PR allows to specify the arguments of the NSwag command on an item-by-item basis or as properties (sourced from any MSBuild file, command line invocation of MSBuild or environment variables.


paulomorgado:
> We probably also need to add
> 
> ```
>     /// <summary>Gets a value indicating whether to generate the output of client interfaces.</summary>
>     public bool SuppressClientInterfacesOutput => _settings.SuppressClientInterfacesOutput;
> 
>     /// <summary>Gets a value indicating whether to generate the output of client classes.</summary>
>     public bool SuppressClientClassesOutput => _settings.SuppressClientClassesOutput;
> ```

Where? I'll do that.

> my problem with all this is that the maintenance gets quite hard, a new config needs to be added to
> 
> * settings
> * command
> * wpf ui
> * props file

Point me in the right direction.

This is a very useful feature for those that want to split DTOs, interfaces and clients in different projects.

You wouldn't believe what I have to do in the templates to implement this without having my own fork for the tools. 😄

paulomorgado:
> We probably also need to add
> 
> ```
>     /// <summary>Gets a value indicating whether to generate the output of client interfaces.</summary>
>     public bool SuppressClientInterfacesOutput => _settings.SuppressClientInterfacesOutput;
> 
>     /// <summary>Gets a value indicating whether to generate the output of client classes.</summary>
>     public bool SuppressClientClassesOutput => _settings.SuppressClientClassesOutput;
> ```
> 
> my problem with all this is that the maintenance gets quite hard, a new config needs to be added to
> 
> * settings
> * command
> * wpf ui
> * props file

That's the cost of backwards compatibility.

Just submitted the addition of those 2.

This is a very mechanical process. Maybe we should automate this.

RicoSuter:
my plan was to eg serialize settings directly and use that instead of command indirection and also CLI always is configured via config JSON file... but this would be quite the breaking change.

RicoSuter:
... but i'd say let's keep it as is for now.

paulomorgado:
From those MSBuild properties, a JSON file can easily be generated,

This just creates another "interface" that can be used to isolate what and how the generation is done.

PR #4570, opened 2023-11-05
Fix template and test bugs from PR #4541

paulomorgado:
Sorry about that everyone.

@MaximeMorin-Devolutions, can you check the changes and propose changes for anything else you found?

### DIFF
--- a/src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.liquid
+++ b/src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.liquid
@@ -35,11 +35,11 @@
 {% else -%}
     public {{ Class }}()
     {
-{% endif -%}
-{% if UseBaseUrl and GenerateBaseUrlProperty -%}
+{%     if UseBaseUrl and GenerateBaseUrlProperty -%}
         BaseUrl = "{{ BaseUrl }}";
+{%     endif -%}
 {% endif -%}
-        {% template Client.Class.Constructor %}
+    {% template Client.Class.Constructor %}
     }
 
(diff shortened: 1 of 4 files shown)


MaximeMorin-Devolutions:
@paulomorgado LGTM, thanks for the quick fixes!

MaximeMorin-Devolutions (approved):

src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.liquid
```diff
@@ -35,11 +35,11 @@
 {% else -%}
     public {{ Class }}()
     {
-{% endif -%}
-{% if UseBaseUrl and GenerateBaseUrlProperty -%}
+{%     if UseBaseUrl and GenerateBaseUrlProperty -%}
         BaseUrl = "{{ BaseUrl }}";
+{%     endif -%}
```
ramonsmits, line 40: Something seems of with BaseUrl, the property isn't assigned due to this when run with NSwagStudio 13.20.0.0.

paulomorgado: Are you replacing NSwagStudio 13.20.0.0 templates with this one?

Can you share a simple repro?

RicoSuter:
Is this ready to merge and will fix the problem? What about the open comment?

paulomorgado:
> Is this ready to merge and will fix the problem? What about the open comment?

As far as I could assert from the tests, yes!

I have more coming up, but I want to keep them bite sized.

Issue #633, opened 2017-02-27
Generate client interfaces indipendently from client classes


Liero:
In case I would like to have client interfaces in separate file (or project) from client classes, this should generate the interfaces:

```
 var settings = new SwaggerToCSharpClientGeneratorSettings
{
     GenerateClientClasses = false,
     GenerateClientInterfaces = true,
};
```

PR #4571, opened 2023-11-05
Added feature to suppress the generation of the output of client classes and interfaces.

paulomorgado:
This PR addresses #633 by adding a command-line argument to `OpenApiToCsharpCommand` `/SuppressClientClassesOutput` to suppress the output of the generated classes and a command-line argument `/SuppressClientInterfacesOutput` to suppress the output of the generated interfaces.

The objective here is for the usage of the tooling to be retro-compatible and to be just an additional feature.

If a user wants to generate a client class that implements an interface but not have the interface output as part of the generation, then `/GenerateClientClasses:true /GenerateClientInterfaces:true /SuppressClientInterfacesOutput:true`.

The same for interfaces. If a user wants to generate a client interface but not have the client class output as part of the generation, then `/GenerateClientClasses:true /SuppressClientClassesOutput:true /GenerateClientInterfaces:true /SuppressClientClassesOutput:true`.

### DIFF
--- a/docs/tutorials/GenerateProxyClientWithCLI/generate-proxy-client.md
+++ b/docs/tutorials/GenerateProxyClientWithCLI/generate-proxy-client.md
@@ -77,7 +77,9 @@ nswag run sample.nswag /runtime:Net50
     "codeGenerators": {
         "openApiToCSharpClient": {
             "generateClientClasses": true,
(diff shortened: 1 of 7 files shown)


diogopoco (approved):

RicoSuter:
Is this ready to merge and will fix the problem?

PR #4575, opened 2023-11-07
Use sorted properties when saving configuration in NSwagStudio

lahma:
It's quite annoying to see all JSON properties jumping around when saving new version of `nswag.json` (migrating from v13 to v14). We could introduce a one time breaking change which will make Studio save file following ordered JSON property names so that later VCS diffs etc would look nicer.

Included change for `NSwag.Sample.NET80/nswag.json` which shows how obsolete properties are removed and they are now sorted correctly after save via NSwagStudio.


src/NSwag.Commands/NSwagDocumentBase.cs
```diff
@@ -515,12 +525,6 @@ private static string TransformLegacyDocument(string data, out bool saveFile)
                 saveFile = true;
             }
 
-            if (data.Contains("\"noBuild\":") && !data.ToLowerInvariant().Contains("UseDocumentProvider".ToLowerInvariant()))
```
lahma, line 518: Couldn't find a property `UseDocumentProvider` anywhere so this seemed unnecessary

lahma:
NET 8 sample's `nswag.json` before:

```json
{
  "runtime": "Net80",
  "defaultVariables": null,
  "documentGenerator": {
    "aspNetCoreToOpenApi": {
      "project": "NSwag.Sample.NET80.csproj",
      "msBuildProjectExtensionsPath": null,
      "configuration": null,
      "runtime": null,
      "targetFramework": null,
      "noBuild": false,
      "verbose": true,
      "workingDirectory": null,
      "requireParametersWithoutDefault": false,
      "apiGroupNames": null,
      "defaultPropertyNameHandling": "Default",
      "defaultReferenceTypeNullHandling": "Null",
      "defaultDictionaryValueReferenceTypeNullHandling": "NotNull",
      "defaultResponseReferenceTypeNullHandling": "NotNull",
      "defaultEnumHandling": "Integer",
      "flattenInheritanceHierarchy": false,
      "generateKnownTypes": true,
      "generateEnumMappingDescription": false,
      "generateXmlObjects": false,
      "generateAbstractProperties": false,
      "generateAbstractSchemas": true,
      "ignoreObsoleteProperties": false,
      "allowReferencesWithProperties": false,
      "excludedTypeNames": [],
      "serviceHost": null,
      "serviceBasePath": null,
      "serviceSchemes": [],
      "infoTitle": "My Title",
      "infoDescription": null,
      "infoVersion": "1.0.0",
      "documentTemplate": null,
      "documentProcessorTypes": [],
      "operationProcessorTypes": [],
      "typeNameGeneratorType": null,
      "schemaNameGeneratorType": null,
      "contractResolverType": null,
      "serializerSettingsType": null,
      "useDocumentProvider": true,
      "documentName": "v1",
      "aspNetCoreEnvironment": null,
      "createWebHostBuilderMethod": null,
      "startupType": null,
      "allowNullableBodyParameters": true,
      "output": "openapi.json",
      "outputType": "Swagger2",
      "assemblyPaths": [],
      "assemblyConfig": null,
      "referencePaths": [],
      "useNuGetCache": false
    }
  },
  "codeGenerators": {}
}
```

After save

```` json

````

src/NSwag.Commands/NSwagDocumentBase.cs
RicoSuter: I think this is gone, correct

RicoSuter:
Can't you just update the nswag.json with the new version and then the props should be stable as the order in c# stays? Do we really need reorder? I think alpha order is not optimal as the props which belong together are shuffled...

PR #4577, opened 2023-11-10
Don't bundle unused Swagger UI ES scripts

lahma:
Swagger UI only needs non-es versions:

```html
<script src="./swagger-ui-bundle.js" charset="UTF-8"> </script>
<script src="./swagger-ui-standalone-preset.js" charset="UTF-8"> </script>
```

And the es scripts (`swagger-ui-es-bundle.js`, `swagger-ui-es-bundle-core.js`) are almost 2MB together and thus unnecessarily bloat the DLL size.


RicoSuter:
Can you do the same for the webapi/legacy pkg?

lahma:
done

PR #4580, opened 2023-11-13
Use .NET 8.0.100

RicoSuter:


lahma:
Remember to regenerate yaml by running build.cmd, 8 RTM is also tomorrow I believe 😉

RicoSuter:
Just running ./build.cmd does not change any files? Yeah I think we can also wait till tomorrow with that :-)

PR #4586, opened 2023-11-16
Replaced LINQ with iteration for performance improvements.

paulomorgado:
**LINQ To Objects** might look good, but comes with a performance penalty.

Using **LINQ To Objects** to create enumerables that have single and immediate iteration should be replaced with the equivalent traditional control flow instructions.

### DIFF
--- a/src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.ConvertToString.liquid
+++ b/src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.ConvertToString.liquid
@@ -37,10 +37,19 @@ private string ConvertToString(object value, System.Globalization.CultureInfo cu
     {
         return System.Convert.ToBase64String((byte[]) value);
     }
+    else if (value is string[])
+    {
+        return string.Join(",", (string[])value);
+    }
     else if (value.GetType().IsArray)
     {
-        var array = System.Linq.Enumerable.OfType<object>((System.Array) value);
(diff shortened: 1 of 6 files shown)


src/NuGet.Config
```diff
@@ -2,8 +2,8 @@
 <configuration>
```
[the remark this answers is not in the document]
paulomorgado: No. My mistake. I'll remove i.

src/NSwag.sln
```diff
@@ -134,6 +135,11 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "NSwag.Sample.NET80", "NSwag
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "NSwag.Sample.NET80Minimal", "NSwag.Sample.NET80Minimal\NSwag.Sample.NET80Minimal.csproj", "{F0569608-BD55-4316-94F0-E85A14D7FE14}"
 EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution Items", "{FBF5AB11-68C3-41B8-9A08-5D25D6E814E4}"
```
RicoSuter, line 138: remove here as well?

paulomorgado: I like to see all the files in my solutions.

But I'll remove it, because it's not the subject of this PR.

PR #4589, opened 2023-11-19
Use and support NET 8 SDK artifacts output

lahma:
Use the new [NET 8 SDK artifacts output layout](https://learn.microsoft.com/en-us/dotnet/core/sdk/artifacts-output) feature which consolidates `bin` and `obj` folders under artifacts directory. Need to move `Directory.Build.props` to root level for proper configuration.

To remove old cruft, might be good idea to run locally in solution root directory:

```powershell
Get-ChildItem .\ -include bin,obj -Recurse | foreach ($_) { remove-item $_.fullname -Force -Recurse }
```

This also improves the project metadata querying by trying to use the new [CLI-based metadata querying](https://learn.microsoft.com/en-us/dotnet/core/whats-new/dotnet-8#cli-based-project-evaluation) , basically `obj `folder can now be anywhere and no need to give extensions path like before (for artifacts output enabled projects it would always fail). New retrieval should be faster too. If NET 8 SDK is not present, old logic will be used.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -31,6 +31,8 @@
     <!-- reduce package size by only including english resources -->
     <SatelliteResourceLanguages>en-US</SatelliteResourceLanguages>
 
+    <UseArtifactsOutput>true</UseArtifactsOutput>
+
   </PropertyGroup>
 
   <ItemGroup>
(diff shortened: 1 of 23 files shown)


RicoSuter:
lgtm. thanks

Issue #3684, opened 2021-11-01
Nswag not sending list of objects correctly


zubairkhakwani:
![image](https://user-images.githubusercontent.com/38009549/139663783-6bd27f02-a686-4515-8514-214db32696d1.png)

when I'm using (**UploadDocument** class which have **`List<Documents>`** class as a member) as a FromForm parameter
nswag will try to convert it into string 

![image](https://user-images.githubusercontent.com/38009549/139664526-f8e521b7-4844-4164-a8a0-06bef5f5654a.png)

and it'll hit the controller with empty data like this

![image](https://user-images.githubusercontent.com/38009549/139664964-3822daca-c0ce-4acf-aaf3-845d7457d4eb.png)

but if I make it Documents from `List<Documents>`

![image](https://user-images.githubusercontent.com/38009549/139665620-3201a5d8-a9e0-4e3d-99a2-725f5c32c2da.png)

Nswag generate this code and work just fine

![image](https://user-images.githubusercontent.com/38009549/139665548-78942dfc-2fb9-4dbd-af0f-2a1adea318de.png)

eholman:
I have the exact same issue.

I've removed nswag because it's not usable for these case, at least not that I'm aware of. 

Pretty recently there was a method added: `ConvertToString` which doesn't add any value for these cases. I'd rather saw an extensible implementation to overwrite the default behavior and also the option to set the name of the StringContent.

Issue #3669, opened 2021-10-19
DateTime in FromForm DTO isn't correctly convert


ebiton-eig:
If a DTO interface containing a TDateTime or TDateTimeOffset is passed as a parameter, the date is not converted to ISO 8601 format but to regional format (the default ConvertToString function is used).

For example, a DTO with DateTimeOffset and IFormFileCollection:

```csharp
    public class PlanningRequestInputDTO
    {
        [Required]
        public DateTimeOffset Start { get; set; }
        public DateTimeOffset End { get; set; }
        public IFormFileCollection Documents { get; set; }
    }
```

The controller function:

```csharp
        [ProducesResponseType(typeof(PlanningRequestOuputDTO), StatusCodes.Status201Created)]
        [HttpPost(Name = nameof(CreatePlanningRequest))]
        public async Task<IActionResult> CreatePlanningRequest([[Required][FromForm] PlanningRequestInputDTO planningRequest)
        {
               // [..]
        }
```

During the conversion, DateTime is processed by default regardless of the NSwag setting, It should use .ToString("s").

```csharp
        private string ConvertToString(object value, System.Globalization.CultureInfo cultureInfo)
        {
            if (value == null)
            {
                return "";
            }
        
            if (value is System.Enum)
            {
                var name = System.Enum.GetName(value.GetType(), value);
                if (name != null)
                {
                    var field = System.Reflection.IntrospectionExtensions.GetTypeInfo(value.GetType()).GetDeclaredField(name);
                    if (field != null)
                    {
                        var attribute = System.Reflection.CustomAttributeExtensions.GetCustomAttribute(field, typeof(System.Runtime.Serialization.EnumMemberAttribute)) 
                            as System.Runtime.Serialization.EnumMemberAttribute;
                        if (attribute != null)
                        {
                            return attribute.Value != null ? attribute.Value : name;
                        }
                    }
        
                    var converted = System.Convert.ToString(System.Convert.ChangeType(value, System.Enum.GetUnderlyingType(value.GetType()), cultureInfo));
                    return converted == null ? string.Empty : converted;
                }
            }
            else if (value is bool) 
            {
                return System.Convert.ToString((bool)value, cultureInfo).ToLowerInvariant();
            }
            else if (value is byte[])
            {
                return System.Convert.ToBase64String((byte[]) value);
            }
            else if (value.GetType().IsArray)
            {
                var array = System.Linq.Enumerable.OfType<object>((System.Array) value);
[...]

Issue #3665, opened 2021-10-16
Generated code puts enum named constant in URL instead int value


andrewnosworthy:
Hi,

I've come across an issue in C# code generation tool.

I'm using a remote swagger service which I'm a consumer of (and have no control over). The swagger.json file defines an enum as follows:
{"name":"req.stroke","in":"query","required":false,"type":"integer","format":"int32","enum":[0,1,2,3,4]}

swaggerClient.cs creates this enum:

`    [System.CodeDom.Compiler.GeneratedCode("NJsonSchema", "10.0.22.0 (Newtonsoft.Json v11.0.0.0)")]
    public enum Stroke
    {
        _0 = 0,
    
        _1 = 1,
    
        _2 = 2,
    
        _3 = 3,
    
        _4 = 4,
    
    }
`

(which is a bit ugly, but to be expected seeing as swagger.json file doesn't define names for the enum values)

swaggerClient.cs produces this code to produce the request (trimmed out irrelevant code):
`        public async System.Threading.Tasks.Task<Response6> Results2Async(... Stroke? req_stroke, ...)
        {
            var urlBuilder_ = new System.Text.StringBuilder();
            urlBuilder_.Append(...);
            ...
            if (req_stroke != null) 
            {
                urlBuilder_.Append(System.Uri.EscapeDataString("req.stroke") + "=").Append(System.Uri.EscapeDataString(ConvertToString(req_stroke, System.Globalization.CultureInfo.InvariantCulture))).Append("&");
            }
            ...
        }
`

If I set the req_stroke parameter to _0, The resulting URL is that it appends "req.stroke=_0", however the remote service obviously wants "req.stroke=0", seeing as the _0 label was invented by NSwag and not something the remote service knows about. 

I can work around by adding a cast in the generated code, but obviously it's pretty undesirable to be editing that:
`            if (req_stroke != null) 
            {
                urlBuilder_.Append(System.Uri.EscapeDataString("req.stroke") + "=").Append(System.Uri.EscapeDataString(ConvertToString((int)req_stroke.Value, System.Globalization.CultureInfo.InvariantCulture))).Append("&");
            }
`

Apologies if I have missed something really obvious here. This is my first crack at using this tool.

Issue #3625, opened 2021-09-13
[Question] C# DateTimeOffset ConvertToString using specific format "yyyy-MM-ddTHH:mm:sszzz"


lydonchandra:
Hi, is there an easy way to override ConvertToString so DateTimeOffset always uses specific format "yyyy-MM-ddTHH:mm:sszzz"
(instead of depending on System.Convert.ToString InvariantCulture ?

One way is to manually edit the generated code:

        private string ConvertToString(object value, System.Globalization.CultureInfo cultureInfo)
        {
            if (value == null)
            {
                return "";
            }
        
            if (value is System.Enum)
            {
                var name = System.Enum.GetName(value.GetType(), value);
                if (name != null)
                {
                    var field = System.Reflection.IntrospectionExtensions.GetTypeInfo(value.GetType()).GetDeclaredField(name);
                    if (field != null)
                    {
                        var attribute = System.Reflection.CustomAttributeExtensions.GetCustomAttribute(field, typeof(System.Runtime.Serialization.EnumMemberAttribute)) 
                            as System.Runtime.Serialization.EnumMemberAttribute;
                        if (attribute != null)
                        {
                            return attribute.Value != null ? attribute.Value : name;
                        }
                    }
        
                    var converted = System.Convert.ToString(System.Convert.ChangeType(value, System.Enum.GetUnderlyingType(value.GetType()), cultureInfo));
                    return converted == null ? string.Empty : converted;
                }
            }
            else if (value is bool) 
            {
                return System.Convert.ToString((bool)value, cultureInfo).ToLowerInvariant();
            }
            else if (value is byte[])
            {
                return System.Convert.ToBase64String((byte[]) value);
            }
            else if (value.GetType().IsArray)
            {
                var array = System.Linq.Enumerable.OfType<object>((System.Array) value);
                return string.Join(",", System.Linq.Enumerable.Select(array, o => ConvertToString(o, cultureInfo)));
            }
            else if (value is System.DateTimeOffset)
            {
                return ((System.DateTimeOffset)value).ToString("yyyy-MM-ddTHH:mm:sszzz");     // <--------- want this specific format
            }
            
            var result = System.Convert.ToString(value, cultureInfo);
            return result == null ? "" : result;
        }

lydonchandra:
Thanks, found out about the Template override Client.Class.ConvertToString.liquid

PR #4606, opened 2023-11-23
Allow customization of how values are converted to string

ahofman:
Users can specify any type they want for dates, however currently they are at the mercy of whatever `System.Convert.ToString()` does for that type. This is particularly problematic for `DateTime` / `DateTimeOffset`, as there's no way for users to customize how the date is serialized.

This PR allows the `ConvertToString` method to be overridden in a derived class, such that users can implement custom serialization for dates in whatever format suits their use case.

#3558 
#3669 
#3684 
#3665 
#3625


paulomorgado:
Give that this is in its own template, it's kind of easy to customize without requiring this change, which is incompatible with users that choose to make their client classes sealed.

PR #4610, opened 2023-11-25
Move argument validation exception to methods to allow JIT inlining.

paulomorgado:
With this PR the throw of `ArgumentNullException`s is moved to a separate method to allow for JIT inlining.


lahma:
JIT most likely won't inline such large methods.

paulomorgado:
> JIT most likely won't inline such large methods.

The methods aren't large, because they are asynchronous methods and will be converted into state machines.

However, the first part of the execution, up until the first `await` can still be optimized. Admittedly, not very much.

PR #4614, opened 2023-11-26
Added .NET 8 runtime target to nswag.cmd

CocaColaBear:
I noticed that for .NET 8 there is no target in nswag.smd


KarolZirex:
@RicoSuter can u check and fix issue? In one week I need nswag to work with dotnet8 :/

PR #4635, opened 2023-12-09
Only publish interesting artifacts in GH build

lahma:
* The new `artifacts`  directory layout produces a lot of noise to output and GitHub publish (`bin` and `obj` folders), picking the interesting ones now
* Removed the cache as Windows agents have lousy I/O which can basically destroy the benefit of caching
* Removed NET 8 setup step as it's already included in GH runners
* Removed dependencies setup from GH run as everything needed is already present


lahma:
Tested and expected artifacts were produced https://github.com/RicoSuter/NSwag/actions/runs/7152526035?pr=4635

PR #4656, opened 2023-12-21
Disable warning CS8625 "Cannot convert null literal to non-nullable reference type." in generated code

norberthuethmayr:
Similar to https://github.com/RicoSuter/NSwag/pull/4422/files but for warning CS8625.

image

### DIFF
--- a/src/NSwag.CodeGeneration.CSharp/Templates/File.Footer.liquid
+++ b/src/NSwag.CodeGeneration.CSharp/Templates/File.Footer.liquid
@@ -8,4 +8,4 @@
 #pragma warning restore 3016
 #pragma warning restore 8603
 #pragma warning restore 8604
-#pragma warning disable 8625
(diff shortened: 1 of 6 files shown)


src/NSwag.CodeGeneration.CSharp/Templates/File.Footer.liquid
```diff
@@ -7,4 +7,5 @@
 #pragma warning restore 8073
 #pragma warning restore 3016
 #pragma warning restore 8603
-#pragma warning restore 8604
\ No newline at end of file
+#pragma warning restore 8604
+#pragma warning disable 8625
```
Numpsy, line 11: Should that be ```restore 8625```?

norberthuethmayr: Yes... I'll address that tmr.

Issue #4205, opened 2022-10-31
C# generated code: exception classes are not created in subsequent references


null-d3v:
When using multiple `OpenApiReference`s with `NSwagCSharp` generation, exception classes are only generated for the first listed dependency.

### Example
```
<ItemGroup>
  <OpenApiReference Include="OpenApiReference/Service1_openapi.json" SourceUrl="http://localhost:5050/openapi.json">
    <CodeGenerator>NSwagCSharp</CodeGenerator>
    <ClassName>{controller}Service</ClassName>
    <Namespace>Service1</Namespace>
    <Options>/ExceptionClass:ServiceException /GenerateClientInterfaces:true /GenerateOptionalParameters:true</Options>
  </OpenApiReference>
  <OpenApiReference Include="OpenApiReference/Service2_openapi.json" SourceUrl="http://localhost:5051/openapi.json">
    <CodeGenerator>NSwagCSharp</CodeGenerator>
    <ClassName>{controller}Service</ClassName>
    <Namespace>Service2</Namespace>
    <Options>/ExceptionClass:ServiceException /GenerateClientInterfaces:true /GenerateOptionalParameters:true</Options>
  </OpenApiReference>
  <OpenApiReference Include="OpenApiReference/Service3_openapi.json" SourceUrl="http://localhost:5052/openapi.json">
    <CodeGenerator>NSwagCSharp</CodeGenerator>
    <ClassName>{controller}Service</ClassName>
    <Namespace>Service3</Namespace>
    <Options>/ExceptionClass:ServiceException /GenerateClientInterfaces:true /GenerateOptionalParameters:true</Options>
  </OpenApiReference>
  </ItemGroup>
<ItemGroup>
```

`ServiceException` definitions in the `Service2` and `Service3` generated files are absent, resulting in compiler failures for those two references. This cannot be rectified by explicitly using `GenerateExceptionClasses`; the exceptions will continue to fail to generate. Using `{controller}` placeholders for the exception class also has the same result.

janrhansen:
Bump - same issue here. Did you find a workaround?

hmoratopcs:
@paulomorgado

paulomorgado:
.NET's OpenAPI client generation has a concept of `FirstForGenerator` and NSwagCSharp uses that to generate exception classes only on the first execution of the generator.

That makes sense to me, as I don't want a different type of exception for all client classes. In fact, I use my own exception class and never generate.

You might try something like this:

```xml
[...]

PR #4668, opened 2023-12-29
Ignore switches overridden by Options

paulomorgado:
Addresses #4205


paulomorgado:
This seems to be impossible without a custom task,

PR #4677, opened 2024-01-04
Correct incorrect pragma usage for restoring 8618

psiservices-dbrink:
Fixes compilation warning "Unrecognized #pragma directive"


psiservices-dbrink:
Dupe of #4674, didn't see it.

PR #4691, opened 2024-01-07
Fix missing assignment of base URL

### LINKED ISSUE #4637: NSwag C# Client generation does not set BaseUrl properly
This is a generated client:

	public IdentityClient(HttpClient httpClient, string baseUrl)
	{
		_baseUrl = baseUrl;
		_httpClient = httpClient;
	}

The generated client should use the property setter to set the BaseUrl property, or the reader should add a trailing slash so the address is correct.

Obviously i can just fix this on my end myself but i thought i'd report it.

paulomorgado:
Fixes #4637

### DIFF
--- a/src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.liquid
+++ b/src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.liquid
@@ -5,7 +5,7 @@
 {% if UseBaseUrl and GenerateBaseUrlProperty -%}
 #pragma warning disable 8618 // Set by constructor via BaseUrl property
     private string _baseUrl;
-#pragma restore disable 8618 // Set by constructor via BaseUrl property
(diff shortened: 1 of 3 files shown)


olegd-superoffice:
This breaks code generation with `/GenerateBaseUrlProperty:false` option in v14.0.1
Here's small project which demonstrates regression: [GenerateBaseUrlPropertyFalse.zip](https://github.com/RicoSuter/NSwag/files/13919794/GenerateBaseUrlPropertyFalse.zip)
It works with `NSwag.ApiDescription.Client` 13.20.0 and 14.0.0 but fails with 14.0.1

paulomorgado:
@olegd-superoffice, please provide the swagger used.

soligen2010:
I think this PR broke BaseUrl.  Please see issue #4764

PR #4703, opened 2024-01-11
Support NSWAG_NOLOGO and NSWAG_NOVERSION environment variables

lahma:
After battling with correct regexes it dawned to me that maybe it's just easier to support specific environment variables like `NSWAG_NOLOGO` and `NSWAG_NOVERSION` - like we have `DOTNET_NOLOGO` and friends. As templates include the `v` prefix didn't touch that for backwards compatibility.


RicoSuter:
Thanks for the PR.

PR #4711, opened 2024-01-15
PropertyNameGenerator: add passing and failing tests demonstrating issue with CodeGeneration

johnnyreilly:
As mentioned here: https://github.com/RicoSuter/NSwag/issues/1874#issuecomment-1890400894

I've identified an issue with the TypeScript code generation of clients when using the `PropertyNameGenerator` property to control naming.

Essentially, when using `TypeScriptTypeStyle.Class` this is respected.  When using `TypeScriptTypeStyle.Interface` it is ignored.  This is illustrated by the enclosing tests, one of which fails; demonstrating the issue.

I'd love to help get this fixed and I figured this was a first step!

I've also added a C# test as well as I thought it'd be useful for coverage.

With the following `CustomPropertyNameGenerator` we can demostrate the issue.

```cs
        class CustomPropertyNameGenerator : NJsonSchema.CodeGeneration.IPropertyNameGenerator {
            public virtual string Generate(JsonSchemaProperty property) => $"XX_{property.Name}";
        }
```

If this is working as expected, all property names should be prefixed with `XX_`:

```ts
export interface IPerson {
    XX_FirstName: string;
    XX_LastName?: string | undefined;
    XX_Birthday?: Date | undefined;
    XX_Sex?: Sex | undefined;
    XX_Address?: Address | undefined;
}
```

But if using `TypeScriptTypeStyle.Interface` it is ignored and instead this is generated:

```ts
export interface Person {
    FirstName: string;
    LastName?: string | undefined;
    Birthday?: Date | undefined;
    Sex?: Sex | undefined;
    Address?: Address | undefined;
}
```

Ideally the `CustomPropertyNameGenerator` should always be in play.

----

Having had a dig through the code I wonder if the issue lies in NJsonSchema somewhere:

https://github.com/RicoSuter/NJsonSchema/blob/master/src/NJsonSchema.CodeGeneration.TypeScript/TypeScriptTypeResolver.cs
https://github.com/RicoSuter/NJsonSchema/blob/3585d60e949e43284601e0bea16c33de4c6c21f5/src/NJsonSchema.CodeGeneration.TypeScript/TypeScriptGeneratorSettings.cs#L36

### DIFF
--- a/src/NSwag.CodeGeneration.Tests/CodeGenerationTests.cs
+++ b/src/NSwag.CodeGeneration.Tests/CodeGenerationTests.cs
@@ -51,8 +51,6 @@ namespace NSwag.CodeGeneration.Tests
             var generator = new CSharpClientGenerator(document, settings);
             var code = generator.GenerateFile();
 
-            // System.IO.File.WriteAllText("/Users/jreilly/code/github.com/NSwag/src/NSwag.CodeGeneration.Tests/test.cs", code);
-
             // Assert
             Assert.Contains("public string XX_Street { get; set; }", code);
             Assert.Contains("public string XX_City { get; set; }", code);
--- a/src/NSwag.CodeGeneration.Tests/CodeGenerationTests.cs
+++ b/src/NSwag.CodeGeneration.Tests/CodeGenerationTests.cs
@@ -38,6 +38,34 @@ namespace NSwag.CodeGeneration.Tests
         }
 
         [Fact]
+        public void When_generating_CSharp_code_with_custom_PropertyNameGenerator_then_output_contains_expected_property_names()
+        {
+            // Arrange
+            var document = CreateDocument();
+
+            // Act
+            var settings = new CSharpClientGeneratorSettings { ClassName = "MyClass" };
+            settings.CSharpGeneratorSettings.Namespace = "MyNamespace";
+            settings.CSharpGeneratorSettings.PropertyNameGenerator = new CustomPropertyNameGenerator();
+
+            var generator = new CSharpClientGenerator(document, settings);
+            var code = generator.GenerateFile();
+
+            // System.IO.File.WriteAllText("/Users/jreilly/code/github.com/NSwag/src/NSwag.CodeGeneration.Tests/test.cs", code);
+
+            // Assert
+            Assert.Contains("public string XX_Street { get; set; }", code);
+            Assert.Contains("public string XX_City { get; set; }", code);
+            Assert.Contains("public string XX_FirstName { get; set; }", code);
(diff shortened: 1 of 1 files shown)


lahma:
I've created repro and fix on NJsonSchema side, but it's quite breaking change as basically all naming will change: https://github.com/RicoSuter/NJsonSchema/pull/1847/files

RicoSuter:
> Essentially, when using TypeScriptTypeStyle.Class this is respected. When using TypeScriptTypeStyle.Interface it is ignored. This is illustrated by the enclosing tests, one of which fails; demonstrating the issue.

As discussed with @lahma this is a feature not a bug. 

The main goal of the TS generator is to generate DTOs which match the specs, so if you generate a TS class we can use the property name generator and generate code which maps eg upper case (FooBar) to lower case (fooBar) within the class (init() method). In the case of TS interface output we cannot transform as it's just an interface describing the DTO directly and thus you cannot use property name generator in this context - i.e there is no transformation runtime code, we just "cast" the result.

The only option would be to implement this conversion in the actual client code (response processing) but this is a lot of work and I'm not sure whether this is a good idea performance wise etc. 

So if you need proper interface names (same applies for e.g. datetime properties which are only converted in class scenario) then you either need to fix it on server side or use TS class output.

RicoSuter:
Another option would be to just preprocess the OpenAPI document (fix property names there) and then generate TS: Though you'd have exactly the same problem: The runtime response would not match the generated interface.

johnnyreilly:
Thanks for the context @RicoSuter - it's possible that I'm coming at this from the wrong angle. Essentially my goal was simply to find a way to customise names.

To customise property names we use `NJsonSchema.CodeGeneration.IPropertyNameGenerator` - would it be appropriate to have an alternate generator for interfaces? Something like `NJsonSchema.CodeGeneration.IInterfaceNameGenerator` (better names are available)

What do you think?

lahma:
Maybe the settings could have just `public IPropertyNameGenerator InterfacePropertyNameGenerator { get; set; }` that would be used and by default could be null and old behavior of not modifying.

RicoSuter:
I'd say it's really against the idea to generate DTO/clients which do not match the spec. 

Instead of adding more complexity, i'd propose to preprocesss the spec/schema with the JsonSchemaVisitor/OpenApiVisitor, transform the property names and then run the regular generator.

RicoSuter:
The visitor can be found here: https://github.com/RicoSuter/NJsonSchema/blob/master/src/NJsonSchema/Visitors/JsonSchemaVisitorBase.cs

Just load the OpenAPI/JSON Schema document, implement an own visitor and modify the properties as you need them and then generate the code.

PR #4716, opened 2024-01-18
Update transitive Microsoft.AspNetCore dependencies in NSwag.Commands…

Numpsy:
… to fix reported security vulnerabilities

refs #2824 

Mend / Whitesource is complaining about libraries included in NSwag,MSBuild, but actually current Visual Studio versions will display warnings themselves in the package manager:
image

### DIFF
--- a/src/NSwag.Commands/NSwag.Commands.csproj
+++ b/src/NSwag.Commands/NSwag.Commands.csproj
@@ -31,6 +31,9 @@
 
   <ItemGroup Condition="'$(TargetFramework)' == 'net462'">
     <PackageReference Include="Microsoft.AspNetCore" Version="[2.1.7, 2.2)" />
+    <PackageReference Include="Microsoft.AspNetCore.Http" Version="[2.1.34, 2.2)" />
+    <PackageReference Include="Microsoft.AspNetCore.Server.Kestrel.Core" Version="[2.1.25, 2.2)" />
+    <PackageReference Include="Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets" Version="[2.1.40, 2.2)" />
     <PackageReference Include="Microsoft.AspNetCore.TestHost" Version="[2.1.1, 2.2)" />
     <PackageReference Include="Microsoft.AspNetCore.Mvc.ApiExplorer" Version="[2.1.3, 2.2)" />
     <PackageReference Include="Microsoft.AspNetCore.Mvc.Formatters.Json" Version="[2.1.18, 2.2)" />


olegd-superoffice:
@RicoSuter These are transitive references, so in order to update they need to be referenced directly. They are flagged as vulnerable by .Net 8 SDK as well. Here's output of `dotnet list package --vulnerable --include-transitive --framework net462`:
```
Project `NSwag.Commands` has the following vulnerable packages
[net462]:
> Transitive Package                                         Resolved   Severity   Advisory URL
> Microsoft.AspNetCore.Http                                  2.1.1      High       https://github.com/advisories/GHSA-hxrm-9w7p-39cc
> Microsoft.AspNetCore.Server.Kestrel.Core                   2.1.3      High       https://github.com/advisories/GHSA-6px8-22w5-w334
                                                                        High       https://github.com/advisories/GHSA-242j-2gm6-5rwx
> Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets      2.1.3      High       https://github.com/advisories/GHSA-vmch-3w2x-vhgq
> System.Net.Http                                            4.3.0      High       https://github.com/advisories/GHSA-7jgj-8wvc-jh57
> System.Text.RegularExpressions                             4.3.0      High       https://github.com/advisories/GHSA-cmhx-cq75-c4mj
```
@Numpsy `Microsoft.AspNetCore.*` packages are referenced only for .Net 4.6.2 framework, but `System.Net.Http` and `System.Text.RegularExpressions` are referenced for all targets and probably should be referenced directly as well? Does Mend / Whitesource complain about these two?

Numpsy:
I'm not seeing those listed at work currently in NSwag.MSBuild 14.0.2, and Mend does usually love to complain about the regex one.
Thing with that one is that it's often not a real problem as most .NET versions have inbox versions that are newer and which take precedence over the old nuget packages, and as Mend is only looking at the binaries in this case and not the source, there might not be any issue for end users.
We could still see about doing an update though.

Saying that though - 

Those old libraries seem to be geting pulled in via transitive references to NETStandard.Library version 1.6.1, and you can sometimes avoid the issue and also greatly reduce the number of referenced nuget packages by updating NETStandard.Library to version 2.

Also - 

- There is a transitive reference to NETStandard.Library version 1.6.1 via Microsoft.Extensions.PlatformAbstractions 1.1.0 (which is actually itself deprecated by Microsoft)
- The code at https://github.com/RicoSuter/NSwag/blob/0c9d87f1623e8a7f7d29ba9593108270e05f987a/src/NSwag.Commands/RuntimeUtilities.cs#L22 makes it look like Microsoft.Extensions.PlatformAbstractions isn't actually being used on .NET Framework
- If that's the case, maybe that dependency could be removed from the .NET 4.6.2 build?

PR #4729, opened 2024-01-22
Add support for sending and recieving large json arrays.

nonwombatant:
Supports large json arrays that may not fit in memory and cause out of memory exceptions.

Uses IAsyncEnumerable and a custom Http Content to send large json arrays.
Uses IAsyncEnumerator to recieve large json arrays.

By returning an IAsyncEnumerator it correctly handles memory safety and IDisposables.

Possible fix for
https://github.com/RicoSuter/NSwag/issues/4236 and https://github.com/RicoSuter/NSwag/issues/3909

Note: I'm like 95% confident about the implementation for reading large json arrays. I'm much less confident about the custom httpContent for sending large arrays being the best approach.


Chaosflo:
@RicoSuter  When will this be merged to support streaming?
I am desperately in need of this =/

nonwombatant:
> @RicoSuter When will this be merged to support streaming? I am desperately in need of this =/

This seems unlikely to be merged, kind of a niche use case I think? There's some workarounds that are posted on the issue pages.

Happy to follow up on this though.

leonandroid:
Does this ticket brings the ability to have streaming mode?

PR #4739, opened 2024-01-29
Annotate BaseUrl set with MemberNotNull to prevent CS8618.

### LINKED ISSUE #4704: NSwag 14.0.1 still issues warning CS8618
When compiling with 14.0.1, the generated C# sdk now has the following warning:

~~~
Warning	CS8618	Non-nullable field '_baseUrl' must contain a non-null value when exiting constructor. Consider declaring the field as nullable.
~~~

That's because even though the constructor assignes `BaseUrl`:

~~~csharp
            BaseUrl = "https://";
~~~

... the compiler is unable to detect that the `_baseUrl` field is in fact initialized through this property.

The way to get rid of the warning is to assign the field directly:

~~~csharp
            _baseUrl = "https://";
~~~

Also, the warnings reported in issue https://github.com/RicoSuter/NSwag/issues/4467 are still there.

powercode:
Fixes #4704
Generate BaseUrl like this

```csharp
 public string BaseUrl
 {
     get { return _baseUrl; }
     [System.Diagnostics.CodeAnalysis.MemberNotNull(nameof(_baseUrl))]
     set
     {
         _baseUrl = value;
         if (!string.IsNullOrEmpty(_baseUrl) && !_baseUrl.EndsWith("/"))
             _baseUrl += '/';
     }
 }
 ```
 This prevents the warning
 ```output
 warning CS8618: Non-nullable field '_baseUrl' must contain a non-null value when exiting constructor. Consider declaring the field as nullable.
 ```


srb-milestone:
This change seems to break .NET Standard 2.0 support.

PR #4756, opened 2024-02-03
Only exclude explicit registrations of IHostedService

### LINKED ISSUE #4722: Can't use NSwag.Build, because of cleaning of hosted services.
Hi!
I was looking into generating Typecript services and models with usage of nswag, but noticed that it will not work because of :
https://github.com/RicoSuter/NSwag/blob/116c636fb7491623eb6c70725b8fb20c450ec96c/src/NSwag.Commands/HostApplication.cs#L112C23-L119C26
I am trying understand why it was introduced when using GetServiceProviderWithHostFactoryResolver.
As right now I can't generate models when using umbraco cms, as it will throw me missing services which are host services!

desjoerd:
This fixes issues when a registered class or interface inherits from IHostedService because they where removed even when it would not run as an IHostedService.

Fixes #4722

Related pr #4523


RicoSuter:
Thanks for the fix.

PR #4766, opened 2024-02-06
Only generate MemberNotNull if NET_5_0_OR_GREATER

powercode:
With this change, .net standard will still produce CS8618, but will at least not fail.

The warning will need to be suppressed by other means.


lahma:
I believe this was resolved by removing the attribute already in master.

PR #4817, opened 2024-03-13
Add new settings Sync/AsyncMethodFormat

david-brink-talogy:
This PR adds two new settings which allow .net string format expressions to be used for generation of method names via the operation name.  

related to: Is it possible to make Adding Async at the end  of a method as Settings option ?

_Originally posted by @baskialdensys in https://github.com/RicoSuter/NSwag/issues/1730#issuecomment-486679134_


TivoliG:
I woul really love to get this feature. 

I have an api, where in the spec i have "CreateWidget" and "CreateWidgetAsync". Since asyn is always postfoxed, the build breaks, since CreateWidget becomes CreateWidgetAsync, and CreateWidgetAsync is also mapped to CreateWidgetAsync.

PR #4826, opened 2024-03-19
Fix: C# Code Generation generates method with return default(void)

### LINKED ISSUE #3912: C# Code Generation generates method with return default(void)
The C# generator generated client method with the following signature:

`public async System.Threading.Tasks.Task AttachmentsPatchAsync(int attachmentId, AttachmentRename body, System.Threading.CancellationToken cancellationToken)`

Problem is that the method tries to return default(void) which doesn't compile:

                    try
                    {
                        var headers_ = System.Linq.Enumerable.ToDictionary(response_.Headers, h_ => h_.Key, h_ => h_.Value);
                        if (response_.Content != null && response_.Content.Headers != null)
                        {
                            foreach (var item_ in response_.Content.Headers)
                        }
    
                        var status_ = (int)response_.StatusCode;
                        return default(void);
                    }

Instead of this I suppose plain return would work.

This is with OperationGenerationMode = OperationGenerationMode.MultipleClientsFromPathSegments.

The Open API spec is: https://dev.procountor.com/static/swagger.latest.json

### LINKED ISSUE #3996: Bug: Generated c# client methods return voids
NSwag v13.15.10.0 (NJsonSchema v10.6.10.0 (Newtonsoft.Json v11.0.0.0))
Windows 10
.Net6
Visual Studio 2022 Preview 6.0

I was trying to do that via VS (Add Service Reference -> OpenAPI) but it wasn't working. So I simply run the command bellow in powershell (`C:\Users\Admin\.nuget\packages\nswag.msbuild\13.15.10\tools\Win\NSwag.exe`).

The command:
```
NSwag.exe openapi2csclient /className:API /namespace:AccessPay /input:{path}swagger.json /output:{path}swaggerClient.cs
```

The swagger scheme: https://eazycollectservices.github.io/EazyCollectAPIv3/swagger.json

The above example generates not compilable client code:
``` c#
if (status_ == 200)
{
    return default(void); // Compile error
}
```

image

nzeemin:
This PR is made to avoid generation of C# code like this one (which leads to compilation error of course):
```
                        if (status_ == 204)
                        {
                            return default(void);
                        }
```

fixes #3912
fixes #3996

### DIFF
--- a/src/NSwag.CodeGeneration.CSharp/Models/CSharpOperationModel.cs
+++ b/src/NSwag.CodeGeneration.CSharp/Models/CSharpOperationModel.cs
@@ -115,7 +115,9 @@ namespace NSwag.CodeGeneration.CSharp.Models
         /// <summary>
         /// The default value of the result type, i.e. default(T) or default(T)! depending on whether NRT are enabled.
         /// </summary>
-        public string UnwrappedResultDefaultValue => $"default({UnwrappedResultType}){(_settings is CSharpClientGeneratorSettings { CSharpGeneratorSettings.GenerateNullableReferenceTypes: true } ? "!" : "")}";
+        public string UnwrappedResultDefaultValue => HasResult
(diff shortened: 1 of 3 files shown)


lahma (approved):
LGTM, thank you!

PR #4836, opened 2024-03-25
Add number type arguments to JsonSchemaToCSharpCommand

yolofy:
The commit introduces four new argument properties to the `JsonSchemaToCSharpCommand `class, specifically for handling number types in different formats. These include `NumberType`, `NumberFloatType`, `NumberDoubleType`, and `NumberDecimalType`. Each of these properties has a default .NET type assigned and can be easily switched if needed.

### DIFF
--- a/src/NSwag.Commands/Commands/CodeGeneration/JsonSchemaToCSharpCommand.cs
+++ b/src/NSwag.Commands/Commands/CodeGeneration/JsonSchemaToCSharpCommand.cs
@@ -85,6 +85,34 @@ namespace NSwag.Commands.CodeGeneration
             get { return Settings.TimeSpanType; }
             set { Settings.TimeSpanType = value; }
         }
+        
+        [Argument(Name = "NumberType", IsRequired = false, Description = "The number .NET type (default: 'double').")]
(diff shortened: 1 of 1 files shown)


JaDuyve:
Hi @RicoSuter,

First of all, thank you for this incredibly handy tooling — it’s been a great help!

Being able to configure this directly in the NSwag configuration file, alongside all other code generation settings, would make managing API contract updates much easier. Thanks @yolofy for raising and implementing the proposed changes.

Support for date and time-related data types is already in place, and extending this to number data types would be extremely useful.

In my case, I frequently work with APIs where numeric fields in the contract aren’t defined with the desired format. For example, monetary values are often not marked with the decimal format in the OpenAPI spec (OpenAPI decimal format). As a result, the generated code defaults to using double, which is not suitable for representing monetary values accurately.

Unfortunately, I often don’t have control over the API specification itself, so I’m forced to manually modify the generated code after each update. This adds extra overhead and requires careful documentation to ensure that these manual adjustments are consistently applied.

This change would really help streamline that process and reduce the chances of errors slipping in.

PR #4870, opened 2024-04-26
Fix parameter position ordering (not ignore Position property)

nesopo:
Fix parameter position ordering (not ignore Position property) at model. Fix moment import syntax error for typescript version 4.3 (rxJsVersion 7.0)


src/NSwag.CodeGeneration.TypeScript/Models/TypeScriptOperationModel.cs
```diff
@@ -36,13 +36,13 @@ public class TypeScriptOperationModel : OperationModelBase<TypeScriptParameterMo
             _settings = settings;
             _generator = generator;
 
-            var parameters = GetActualParameters();
+            var parameters = GetActualParameters().OrderBy(p => p.Position).ToList();
```
lahma, line 39: I think the `ToList()` calls should be removed from here and below, there's no need to materialize as list as it is done at the end anyway.

PR #4888, opened 2024-05-11
Restores support for instance-based JSON serializer settings in a non-breaking way

### LINKED ISSUE #4662: Please revert static UpdateJsonSerializerSettings
When trying to update to v14-preview, I found that our code no longer compiles because the generated `UpdateJsonSerializerSettings` partial method has been changed to become static. This doesn't work for us, because we provide a NuGet package that contains a base class to hook into the serializer settings, in order to support [JSON:API](https://jsonapi.org/) partial patch.

For example:
```c#
// In our NuGet package
public abstract class JsonApiClient : IJsonApiClient
{
   protected void SetSerializerSettingsForJsonApi(JsonSerializerSettings settings)
   {
        // implementation relies on instance-based state
   }
}

// Consumer of our NuGet package
public partial class ExampleApiClient : JsonApiClient
{
    partial void UpdateJsonSerializerSettings(JsonSerializerSettings settings)
    {
        SetSerializerSettingsForJsonApi(settings);
    }
}
```

Would you please consider reverting this change? I can't think of a way to work around it.

### LINKED ISSUE #4701: (STJ) Possibility to pass as parameter an instance of JsonSerializerOptions in constructor to follow Microsoft's best practices
According to the official documentation it says that it should not be instantiated continuously:

https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/configure-options?pivots=dotnet-8-0#reuse-jsonserializeroptions-instances

In the template you can see that it is instantiated without the possibility of passing an instance to it from outside.

Relevant sections of the template:

https://github.com/RicoSuter/NSwag/blob/v14.0.1/src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.liquid#L13-L17

https://github.com/RicoSuter/NSwag/blob/v14.0.1/src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.liquid#L68-L81

Code in charge of generating the instantiation in NJsonSchema:

https://github.com/RicoSuter/NJsonSchema/blob/v11.0.0/src/NJsonSchema.CodeGeneration.CSharp/CSharpJsonSerializerGenerator.cs#L78-L89

**I would like it to be in the constructor and not in a static way by following dependency injection patterns which is how I instantiate the clients.**

bkoelman:
This PR adds an instance-based field for serializer settings in generated clients, which is `null` by default. Consumers can implement the `Initialize()` partial method and assign the `_instanceSettings` field from there, which takes precedence over static settings.

Example usage to change serializer settings at runtime, based on configuration:
```c#
public sealed class AppOptions
{
    public bool PrettyPrintJson { get; set; }
}

public partial class ExampleApiClient
{
    private readonly IOptionsMonitor<AppOptions> _optionsMonitor;

    private static readonly JsonSerializerSettings JsonIndented = new() { Formatting = Formatting.Indented };
    private static readonly JsonSerializerSettings JsonNone = new() { Formatting = Formatting.None };

    public ExampleApiClient(IOptionsMonitor<AppOptions> optionsMonitor, HttpClient httpClient)
        : this(httpClient)
    {
        _optionsMonitor = optionsMonitor;
    }

    partial void Initialize()
    {
        _instanceSettings = _optionsMonitor.CurrentValue.PrettyPrintJson ? JsonIndented : JsonNone;
    }
}
```

The real use case is adding stateful JSON converters, such as recommended at https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/converters-how-to?pivots=dotnet-8-0#preserve-references.

@RicoSuter Please consider this PR seriously. I've explained in #4662 why adding this is _critical_ for us to continue supporting NSwag usage.

Fixes #4662, fixes #4701.


TWhidden:
Appreciate you putting together a PR for this. I'm on my mobile phone trying to dig through it. Does it also support System.Text.Json? We switched away from Newtonsoft, I imagine the PR would have to support both options.  But most definitely something that needs to be done as we have to support changing serialization options at runtime, based on the server version our clients are connected to.  Thanks again, and the right idea I feel with a clean option to support both ways.

bkoelman:
@TWhidden Yes, I've tested that it also supports System.Text.Json.

ErikPilsits-RJW:
@bkoelman this looks good to me! It's a good combination of safe and flexible.

@RicoSuter

bkoelman:
@RicoSuter I'm not sure what you're referring to, can you clarify? This change is fully backwards compatible. It just adds a field and a partial method to enable overruling the static serializer. The existing behavior is unaffected unless user code activates the added bits.

RicoSuter:
Lgtm, thx.

quentin-lalitte:
Hi @bkoelman,
I still have an issue when using NSwag v14.6.0
My JsonConverter is not found through UpdateJsonSerializerSettings.
The workaround is to remove this line in the generated client class : 
`static partial void UpdateJsonSerializerSettings(Newtonsoft.Json.JsonSerializerSettings settings);`

Did I miss something ?

bkoelman:
@quentin-lalitte Did you assign the private field `_instanceSettings` from the partial `Initialize` method in your partial class? It works for me.

quentin-lalitte:
> @quentin-lalitte Did you assign the private field `_instanceSettings` from the partial `Initialize` method in your partial class? It works for me.

I see but the thing is i'm not using a partial class, I have an abstract class which implement UpdateJsonSerializerSettings.
I found the parameter GenerateUpdateJsonSerializerSettingsMethod, setted at false that resolve my problem.
Thanks for the answer btw.

PR #4891, opened 2024-05-12
Fixed: impossible to generate exception classes for non-first client

### LINKED ISSUE #4890: Bug: impossible to generate exception classes for non-first client
When a project contains multiple `<OpenApiReference>` entries, the `ApiException` class is only generated for the first entry. While this is a nice default, since v14 it has become _impossible_ to generate exception classes for all entries (they live in separate namespaces).

The problematic line is in `NSwag.ApiDescription.Client.targets`:
```xml
<Command Condition="!%(FirstForGenerator) OR (%(NSwagGenerateExceptionClasses) != '' AND !%(NSwagGenerateExceptionClasses))">%(Command) /GenerateExceptionClasses:false</Command>
```
which _always_ adds `/GenerateExceptionClasses:false` for the non-first entry, ignoring the configured value of `NSwagGenerateExceptionClasses` in `<OpenApiReference>`.

When trying to work around it by adding the argument explicitly using:
```xml
<Options>/GenerateExceptionClasses:true</Options>
```
the build fails because the `/GenerateExceptionClasses` parameter is now specified multiple times (which makes NConsole fail).

[GenerateExceptionClassesDemo.zip](https://github.com/RicoSuter/NSwag/files/15285883/GenerateExceptionClassesDemo.zip) contains a minimal repro. It fails to build the generated `QueryStringsClient` file (second entry in csproj), because it doesn't declare the `ApiException` class while referencing it.

To fix this, the problematic line should be replaced with:
```xml
        <Command Condition="!%(FirstForGenerator) AND ('%(NSwagGenerateExceptionClasses)' == '')">%(Command) /generateExceptionClasses:false</Command>
        <Command Condition="'%(NSwagGenerateExceptionClasses)' != ''">%(Command) /generateExceptionClasses:%(NSwagGenerateExceptionClasses)</Command>
```

The fix _only_ adds `/generateExceptionClasses:false` for non-first entries that aren't configured. In all other cases, it takes the configured value, similar to all the other `NSwag*` properties. The same fix should be applied for `NSwagGenerateResponseClasses`, which works similarly.

Additionally, the `JsonInheritanceConverter`, `JsonInheritanceConverterAttribute`, and `JsonInheritanceAttribute` could benefit from the same fix, but there's no `NSwag*` property to control whether they are generated.

bkoelman:
Fixes #4890.

### DIFF
--- a/src/NSwag.ApiDescription.Client/NSwag.ApiDescription.Client.targets
+++ b/src/NSwag.ApiDescription.Client/NSwag.ApiDescription.Client.targets
@@ -19,7 +19,8 @@
         <Command>$(_NSwagCommand) openapi2csclient /className:%(ClassName) /namespace:%(Namespace)</Command>
       </CurrentOpenApiReference>
       <CurrentOpenApiReference>
-        <Command Condition="!%(FirstForGenerator) OR (%(NSwagGenerateExceptionClasses) != '' AND !%(NSwagGenerateExceptionClasses))">%(Command) /GenerateExceptionClasses:false</Command>
+        <Command Condition="!%(FirstForGenerator) AND ('%(NSwagGenerateExceptionClasses)' == '')">%(Command) /generateExceptionClasses:false</Command>
+        <Command Condition="'%(NSwagGenerateExceptionClasses)' != ''">%(Command) /generateExceptionClasses:%(NSwagGenerateExceptionClasses)</Command>
       </CurrentOpenApiReference>
       <CurrentOpenApiReference>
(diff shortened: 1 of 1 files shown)


RicoSuter:
Personally I do not use this functionality. Are you sure that this does not introduce regressions?

bkoelman:
I'm pretty confident this is safe to take, I've tested various scenarios. I understand your concerns and am happy to help out in case regressions surface. I've been a professional  .NET developer for 20 years. This change and #4891 are blocking us from adopting v13. Please let me know what I can do get this merged.

PR #4948, opened 2024-07-30
♻️ CodeGen - CSharp - generate additional properties as a `deepObject`

MattParkerDev:
Currently, additionalProperties are generated like a dictionary, regardless of the `"style" = "deepObject"`

#### example parameter from openapi spec
```json
{
  "name": "attributes",
  "in": "query",
  "required": false,
  "explode": true,
  "style": "deepObject",
  "schema": {
    "minProperties": 1,
    "type": "object",
    "properties": {
      "name": {
        "type": "string"
      }
    },
    "additionalProperties": {
      "type": "string"
    }
  }
}
```

Currently, named properties such as `name` in the example json, generate correctly in the C# code, however un-named additional properties do not, and should.

as per https://swagger.io/docs/specification/serialization/
image
image


MattParkerDev:
@RicoSuter is this something that can be merged? :)

PR #4964, opened 2024-08-16
removes warnings when using system.text.json in combination with nullable reference types

### LINKED ISSUE #4963: Using System.Text.Json for C# client generation causes warnings in combination with Nullable Reference Types
The generated JsonInheritanceConverter<TBase> has two NRT warnings that could be surpressed with #pragma to prevent build warnings:

#pragma warning disable 8602 // Disable "CS8602 Dereference of a possibly null reference"
#pragma warning disable 8600 // Disable "CS8600 Converting null literal or possible null value to non-nullable type"

kev-andrews:
fixes #4963


lahma:
This is already part of master, thanks for the PR though.

PR #4994, opened 2024-10-07
Fix required handling for formdata properties

KirkMunroSagent:
This is my first pull request for this project.

The goal here is simple: to update nswag such that it respects the required attribute on formdata parameters (currently it does not).

Please let me know if the proposed changes are in an inappropriate location in this project. After stepping through the code multiple times, this seems like the right place for the changes that I included, but if not, I'll update the PR according to the feedback.

Thank you!

### DIFF
--- a/src/NSwag.Generation.AspNetCore.Tests.Web/Controllers/Parameters/FileUploadController.cs
+++ b/src/NSwag.Generation.AspNetCore.Tests.Web/Controllers/Parameters/FileUploadController.cs
@@ -36,5 +36,21 @@ namespace NSwag.Generation.AspNetCore.Tests.Web.Controllers.Parameters
 
             public IFormFile Contents { get; set; }
         }
+
+        [HttpPost("UploadAttachment2")]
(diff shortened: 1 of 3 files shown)


KirkMunroSagent:
@lahma I opened this PR a year ago today. What does it take to get a PR workflow approved so that the PR can be merged-in?

KirkMunroSagent:
@RicoSuter: Maybe you can share some insight? I have a PR open for a year with no feedback of any kind on what needs to be done to get it approved. It has been complete and ready to go for a year. I have issue #5261 where I identified a problem, and the solution, and @lahma suggested I open a PR, but I'm hesitant to do any work here when PRs are not properly managed by the repo owners. And I have been running into breaking changes over the past several releases, such as the one identified in #5242, which I'm pretty sure still isn't fixed in 14.6.2. I want to keep using nswag, we're quite dependent on it at the moment, but I've been finding it very unreliable lately, and the maintainers haven't been responsive so I'm not sure what to do.

KirkMunroSagent:
@RicoSuter Bump.

PR #5016, opened 2024-11-11
Fix for unwanted parameter explosion in C# client 

simonnieder:
In response to #4310, since it seems like there will be no follow-up.

In the current implementation the parameter name is specified every time in the request:
`/users?id=3&id=4&id=5`

However, as per specification, if we set `explode` to false, we would expect:
`/users?id=3,4,5`

This pull request removes the unnecessary explosion, when `explode` is set to false, for arrays with collectionFormat "csv" and in the case when no collectionFormat is set.


RicoSuter:
thx for the pr

Issue #4518, opened 2023-09-14
Help with C# Client IFormFile


kingua:
I am using NSwag to generate a C# client from a dotnet 7 Web API that is consumed by a Blazor WASM application. I am having trouble getting a Post with an `IFormFile` to upload an attachment with the client.

Here is my API Controller method. Note that it takes an `IFormFile` parameter.

```
[MapToApiVersion(Constants.ApiVersions.V1)]
[HttpPost]
[ProducesResponseType(typeof(BlobContentInfo), StatusCodes.Status201Created)]
[ProducesResponseType(typeof(IEnumerable<ValidationError>), StatusCodes.Status400BadRequest)]
[ProducesResponseType(StatusCodes.Status404NotFound)]
public async Task<ActionResult<BlobContentInfo>> Post(
    [FromRoute] int projectId,
    [FromRoute] int equipmentId,
    [FromForm] string? description,
    [FromForm] IFormFile file,
    [FromForm] AttachmentType? attachmentType = null)
{
    return await TryGetAsync<ActionResult<BlobContentInfo>>(async _ =>
    {
        var user = await GetUser();

        var equipment = await Mediator.Send(new GetEquipment { EquipmentId = equipmentId, ProjectId = projectId, User = user });
        if (equipment == null)
            return NotFound();

        var attachment = new AttachmentUpload
        {
            EquipmentId = equipmentId,
            UploadedOn = DateTimeOffset.UtcNow,
            AttachmentType = attachmentType,
            Description = description,
            FormFile = file
        };

        var result = await _validator.ValidateAsync(attachment);

        if (!result.IsValid)
            return BadRequest(result.ToValidationErrors());

        var uploaded = await Mediator.Send(new UploadAttachment
        {
            ProjectId = projectId,
            AttachmentUpload = attachment,
            User = user
        });

        return CreatedAtAction(nameof(Get), new { projectId, equipmentId }, uploaded);
    }, (ex, _) => Error<BlobContentInfo>(ex));
}
```

Here is my Service Method which calls the generated API client method.

1. Note that is splits the `IFormFile` into parameters which make up its constituent parts. Is there a way to avoid this?
[...]

Issue #4174, opened 2022-10-10
An operation with a required IFormFile is not treated as non-nullable parameter


frabe1579:
An operation with a required IFormFile is not treated as non-nullable parameter

### Requirements:
Asp.Net Core 6.0.9
NSwagStudio v13.16.1.0
*Important:* Nullable Refeence Types enabled on client.

### How to reproduce:
On server side, (asp.net core 6), add a controller action like this:
```
    [HttpPost]
    public ActionResult CreateFile([BindRequired] IFormFile file) {
      return Ok();
    }
```
When generating client code via NSwagStudio, a method like this is generated:
```
    Task Post(FileParameter file = null);
```
I have enabled Nullable Reference Types, so this warning is generated:
warning CS8625: Cannot convert null literal to non-nullable reference type.

### Expected result
With Nullable Reference Types enabled, no default value for file parameter is expected, because it is required.
This is not a problem with other normal actions, where Required/BindRequired parameters or models are treated as non-nullable non-default correctly.

Daylily-Zeleen:
I meet this issue, too.

In my case, the required parameter `List<IFormFile> fiels` will be generated  as `System.Collections.Generic.IEnumerable<FileParameter> files = null` .

Of course, all Defatult Refrence Type Null Handling are setted as `NotNull` like below:
```
services.AddOpenApiDocument(config =>{  
    config.PostProcess = document =>
    config.DefaultReferenceTypeNullHandling = ReferenceTypeNullHandling.NotNull;
    config.DefaultDictionaryValueReferenceTypeNullHandling = ReferenceTypeNullHandling.NotNull;
    config.DefaultResponseReferenceTypeNullHandling = ReferenceTypeNullHandling.NotNull;
});

```

michaelhahn-doka:
I have the same issue and haven't found a proper solution (or even a workaround) besides disabling nullability entirely for now.

ShemJM:
I am having a related issue where the `FileParameter[]` array is not being added to the FormData in the method so not being posted to the api

`if (proofImages !== undefined && proofImages !== null)
            proofImages && proofImages.forEach(item => { url_ += "ProofImages=" + encodeURIComponent("" + item) + "&"; });
        url_ = url_.replace(/[?&]$/, "");`
        
        
        
        `        const content_ = new FormData();
        if (deliveryIds !== null && deliveryIds !== undefined)
[...]

PR #5023, opened 2024-11-14
Adds check to consider the required flag on multipart forms in the C# Code generator

simonnieder:
The problem is that for multipart forms the `required` fields are ignored. This causes the client to throw an `ArgumentNullException` even if the parameter is NOT required.

For
```json
            "multipart/form-data": {
              "schema": {
                "required": [
                  "file"
                ],
                "type": "object",
                "properties": {
                  "file": {
                    "type": "string",
                    "description": "",
                    "format": "binary"
                  },
                  "fileDescription": {
                    "type": "string",
                    "description": ""
                  }
                }
              }, ...
```
we would expect
```csharp
                    if (file == null)
                        throw new System.ArgumentNullException("file");
                    else
                    {
                     ...
                    }

                    if (fileDescription != null)
                    {
                     ...
                    }
```
However, we get
```csharp
                    if (file == null)
                        throw new System.ArgumentNullException("file");
                    else
                    {
                     ...
                    }

                    if (fileDescription == null)
                        throw new System.ArgumentNullException("fileDescription");
                    else
                    {
                     ...
                    }
```

### DIFF
--- a/src/NSwag.CodeGeneration.CSharp.Tests/BinaryTests.cs
+++ b/src/NSwag.CodeGeneration.CSharp.Tests/BinaryTests.cs
@@ -442,7 +442,7 @@ components:
             var code = codeGenerator.GenerateFile();
 
             //// Assert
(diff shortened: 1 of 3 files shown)


AroglDarthu:
I think this would also solve issue #4174 and #4518 right?
@RicoSuter please accept the PR!

Issue #3414, opened 2021-04-13
C# generated client using FormUrlEncodedContent instead of StringContent


leo-mck:
I am working with an API that generate a specification like this:

```
{
    "swagger": "2.0",
    "info": {
        "version": "v1",
        "title": "SwaggerTest"
    },
    "host": "localhost:44386",
    "schemes": [
        "https"
    ],
    "paths": {
        "/api/Values": {
            "post": {
                "tags": [
                    "Values"
                ],
                "operationId": "Values_Post",
                "consumes": [
                    "application/json",
                    "text/json",
                    "application/xml",
                    "text/xml",
                    "application/x-www-form-urlencoded"
                ],
                "produces": [],
                "parameters": [
                    {
                        "name": "value",
                        "in": "body",
                        "required": true,
                        "schema": {
                            "type": "string"
                        }
                    }
                ],
                "responses": {
                    "204": {
                        "description": "No Content"
                    }
                }
            }
        }
    },
    "definitions": {}
}
```

This used to generate code like this:

```
var content_ = new System.Net.Http.StringContent(Newtonsoft.Json.JsonConvert.SerializeObject(value, _settings.Value));
content_.Headers.ContentType = System.Net.Http.Headers.MediaTypeHeaderValue.Parse("application/json");
```

Today I updated NSwagStudio and now it generates code like this:

```
var json_ = Newtonsoft.Json.JsonConvert.SerializeObject(value, _settings.Value);
var dictionary_ = Newtonsoft.Json.JsonConvert.DeserializeObject<System.Collections.Generic.Dictionary<string, string>>(json_, _settings.Value);
var content_ = new System.Net.Http.FormUrlEncodedContent(dictionary_);
content_.Headers.ContentType = System.Net.Http.Headers.MediaTypeHeaderValue.Parse("application/json");
```

This looks like an error from the API since it should only accept `application/json` but I do not have access to the API source code, if I manually delete the `application/x-www-form-urlencoded` from `consumes`, it works like before but I would like to understand if this change is a bug or if there is something I can configure on client-side to be able to generate the code as it was.

AGlezB:
Same here. This happened from 13.10.8 to 13.10.9 with this change in src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.liquid:

https://github.com/RicoSuter/NSwag/commit/df18a5d7a2d4a8000e8d172640aa183f95f481d8

[...]

PR #5029, opened 2024-11-19
Fix for #3414 - C# generated client using FormUrlEncodedContent instead of StringContent

ErikApption:
This is a fix for #3414. When "x-www-form-urlencoded" is accepted, NSwag will generate an invalid client that sends url encoded form as "application/json". This change checks if "application/json" is accepted and will prioritize it.

- renamed ConsumesFormUrlEncoded to ConsumesOnlyFormUrlEncoded in liquid templates
- ConsumesOnlyFormUrlEncoded will be set to false if json is accepted


RicoSuter:
Thanks for the PR. 
Hopefully this does not break anyone..

PR #5032, opened 2024-11-20
Fix: When generated TS class name is "Error", it is conflicting with TS Error class

doruk-info:
Where one of the generated class name is "Error"  it cause confusion to `throw new Error(...)` operations and causing the build to fail.

Following is example build failure message:

```
ERROR in ./src/api/index.ts:3008:29
TS2559: Type '"The parameter 'foo' must be defined."' has no properties in common with type 'IError'.
    3006 |         let url_ = this.baseUrl + "/foo";
    3007 |         if (foo === undefined || foo === null)
  > 3008 |             throw new Error("The parameter 'foo' must be defined.");
         |                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

This is because generated class "Error" accepts IError generated interface in ctor.

To resolve this issue we can use globalThis.Error when generating request bodies.

image


lahma (approved):
LGTM, please create a separate branch when submitting PRs, now I had to push against your master.

PR #5036, opened 2024-11-21
Add analysis rules and relaxations

lahma:
Minimalist approach before #5033 to be safe to merge


lahma:
Contains only trivial changes and most pointing to tests, so merging when green.

PR #5052, opened 2024-11-23
Upgrade to NUKE 9

lahma:


Saibamen:
Can you please fix GitHub Actions for forks? They are crashing, because publish job requires some keys.

This could be skipped in yaml files by:
1. `if: github.repository_owner == 'RicoSuter'`
2. `if: github.event.repository.fork == false`

I'm not familiar with NUKE to do this

Issue #3620, opened 2021-09-06
Typescript client - generated parameter not typed as array for dictionary query params 


RogueJay:
I am raising this as it follow on from https://github.com/RicoSuter/NSwag/issues/3529, which I have determined in slightly different.

I'm trying to define a swagger file that will allow for URLs for in the following format
`?filter[aField]=value&filter[otherField]=otherValue`

I have worked out that the following swagger definition allows Swagger.UI to process the inputs properly, and create the correct URL
```
            "name": "filter",
            "in": "query",
            "style": "deepObject",
            "schema": {
              "type": "object",
              "additionalProperties": {
				  "type": "string"
				},
            }
```

The deepObject property is required for Swagger.UI to work correctly.

This nicely produces the typescript method definition of
`filter: { [key: string]: string; } | undefined`

However, the code to parse this query parameter is just defined as 
url_ += "filter=" + encodeURIComponent("" + filter) + "&"

This is not how a dictionary for a query parameter. At least not a deepObject type. I think the following code would be required in Client.RequestUrl.liquid to replicate the desired approach

```
{%     elseif parameter.IsDictionary && parameter.IsDeepObject -%}
	for (let key in {{ parameter.VariableName }}) {
		if (Object.prototype.hasOwnProperty.call({{ parameter.VariableName }}, key)) {
			url_ += "{{ parameter.Name }}[" + key + "]=" + encodeURIComponent("" + {{ parameter.VariableName }}[key]) + "&";
		}
	};
```

joskoanicic:
Was there any work on this? 
The C# generator has the same problem it recognises the parameter as Dictionary and ends using this: 
```
{% elsif parameter.IsDictionary -%}
foreach (var item_ in {{ parameter.VariableName }}) { urlBuilder_.Append(System.Uri.EscapeDataString(item_.Key) + "=").Append(System.Uri.EscapeDataString(ConvertToString(item_.Value, System.Globalization.CultureInfo.InvariantCulture))).Append("&"); }
```

What would work is the modified lined used from the IsDeepObject segment:


```
{% elseif parameter.IsDictionary && parameter.IsDeepObject -%}
foreach (var item_ in {{ parameter.VariableName }}) { 
[...]

PR #5066, opened 2024-12-11
Fix for #3620 - .NET Client generation - Query string dictionary is missing property name

claudiunicola:
This is a fix for https://github.com/RicoSuter/NSwag/issues/3620: Query string dictionary is missing property name

**Proposed solution:**
Prepended property name for dictionary keys, in `src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.QueryParameter.liquid`

**Before:**
`?key1=value1&key2=value2`

**After:**
`?propertyDictName.key1=value1&propertyDictName.key2=value2`


Rasmus715:
@RicoSuter bump! 
Please take a look :)

PR #5068, opened 2024-12-12
fix media type on empty request body

jamesmanningrapidscale:
the request body should match the media type of what the call consumes (what the server expects), not what the call produces.

This caused runtime break in my use case, as the Produces value includes parameters that don't parse as a simple media type

image


RicoSuter:
Thanks for the PR.

PR #5076, opened 2024-12-30
Nswag clients with some endpoints excluded

knoepdan:
I added 2 new settings that allow some endpoints to be NOT generated. 

Main benefits: 
- reduced size (especially for typescript)
- reduced complexity (especially API is big but only needs a fraction of the endpoints)
- helps with some workflows (depends on the workflows and companies, see comment below). 

The first 2 benefits are relatively obvious but what this Pull-Request would help us the most are actually that we could facilitate some workflows. 
a) When have a breaking change to an api we usually create a new version of the same endpoint and deprecate the old one. A few releases later we then remove the old api. However, when we remove the endpoint, NSWAG often changes some names of the new endpoint so the calling code has to be adapted as well. By always ignoring the deprecated endpoints, one could adjust the naming immediately and it not when we remove the endpoint (when it's less obvious). 
b) sometimes we deliberately want an application to only use (or exclude) certain endpoints because of reasons like permissions / future plans for the api's etc. and it would help us to exclude some endpoints from the start to prevent new developers from unwanted endpoints by accident. 

I hope this PR would also benefit a lot of other NSWAG users. 
However, since I'm not too familiar with the project (from a contributor perspective), I will gladly make changes or accept feedback. 

-     general idea + approach??
-     chosen naming ok?
-     I added the change specifically to OpenApiToCSharpClientCommand + OpenApiToTypeScriptClientCommand. I could move it up to the base class. Would that be ok or would you rather have it limited to Typescript (or for all TS, C# Client + C# Controller)
-     I was not sure where to place tests.... (for Typescript I added a new class)

Anyway, would be glad about feedback and even more so if I could contribute to NSWAG.

PS: I tried a pull request before but for some reason it was rejected due to an ubuntu issue. (which I'm pretty but sure.. about 99.9% this PR should not affect this)


knoepdan:
The ubuntu run seems to fail, but I believe it has more to do with the setup than with my changes. 
How can I fix it? (even though I doubt it's in my power)

knoepdan:
Thanks for the tip. I did just that but now the windows check is failing.  Again something that I didn't touch :-(

**Error message**
19:16:09 [DBG] C:\Program Files\dotnet\sdk\9.0.200\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.Sdk.targets(262,5): error NETSDK1216: Package Microsoft.Net.Sdk.Compilers.Toolset is not downloaded but it is needed because your MSBuild and SDK versions are mismatched. Ensure version 9.0.200 of the package is available in your NuGet source feeds and then run NuGet package restore from Visual Studio or MSBuild. [D:\a\NSwag\NSwag\src\NSwag.AspNetCore.Launcher\NSwag.AspNetCore.Launcher.csproj::TargetFramework=net8.0]

knoepdan:
Closed it due to some (non-related) build issues, merge conflicts and limited feedback. Might try improve this and create another PR at some point

PR #5090, opened 2025-01-15
204 response enhancements

Banner-Keith:
# Purpose of change

Partially addresses issues from #1259 

204 is a valid response type and it is valid to potentially return a 200 or a 204 from the same route. Currently nswag generates a server error for 204 results. In addition, this covers the situation where one of the success response types may return a null, but not all.

# Change Summary

- New document generator setting `ResponseStatusCodesToTreatAsNullable` to allow for certain response codes to be treated as a nullable response, even if the response type is void. This means that we don't have to decorate everything that can return a 204 with the `CanBeNull` attribute.

- On the client generation side, we treat all 2xx responses with the same response type as the 200 response as success responses. Expanded that to include responses with the same type but nullable. e.g. a 200 with `MyClass` and a 204 with `MyClass?` would both return true when `IsSuccessStatusCode` is checked. A void type on the 204 would still not return true. That could be another enhancement to make later.

- During client generation, if there are multiple success responses (as defined by the result of `IsSuccessStatusCode`) where some are nullable, we set the whole method return type as nullable.

# Usage Example

In Program.cs we tell the document generator that we want all 204 methods that do not have a `NotNull` attribute to be treated as nullable response types. We also give the response type the same class as the 200 result to pass checks in the generator library that only treat results as successful if they have the same return type as the 200. That part is existing behavior.
``` csharp
services.AddOpenApiDocument(document => 
{ 
    document.ResponseStatusCodesToTreatAsNullable = [HttpStatusCode.NoContent];
});
```

We have an example method in a .net web api.
``` csharp
/// <summary>Example Summary</summary>
[HttpGet]
[ProducesResponseType(typeof(MyClass), StatusCodes.Status200OK)]
[ProducesResponseType(typeof(MyClass), StatusCodes.Status204NoContent)]
public async Task<IActionResult> Get()
{
    MyClass? result = await _myservice.GetResult();
    
    if (result == null)
    {
	    return NoContent();
    }
    
    return Ok(result);
}
```

Note that the 200 is not a nullable response and the 204 is nullable due to the setting in Program.cs. They both use the same return type.
Using this example open api spec and `generateNullableReferenceTypes` set to true. 

``` json
{
  "/api/Example": {
    "get": {
      "summary": "Example Summary",
      "operationId": "Example_Get",
      "responses": {
        "200": {
          "description": "",
          "content": {
            "application/json": {
              "schema": {
                "$ref": "#/components/schemas/MyClass"
              }
            }
          }
        },
        "204": {
          "description": "",
          "content": {
            "application/json": {
              "schema": {
                "nullable": true,
                "oneOf": [
                  {
                    "$ref": "#/components/schemas/MyClass"
                  }
                ]
              }
            }
          }
        }
      }
    }
  }
}
```

We would previously have generated something like this:

``` csharp
public virtual async System.Threading.Tasks.Task<MyClass> Get(System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) 
{
    if (status_ == 200)
    {
        var objectResponse_ = await ReadObjectResponseAsync<InlPerson>(response_, headers_, cancellationToken).ConfigureAwait(false);
        if (objectResponse_.Object == null)
        {
            throw new ApiException("Response was null which was not expected.", status_, objectResponse_.Text, headers_, null);
        }
    }
    else
    if (status_ == 204)
    {
        // Throws error on 204 which makes this client unusable
        var objectResponse_ = await ReadObjectResponseAsync<InlPerson?>(response_, headers_, cancellationToken).ConfigureAwait(false);
        throw new ApiException<InlPerson?>("A server side error occurred.", status_, objectResponse_.Text, headers_, objectResponse_.Object, null); 
    }
}
```

Now we get this:

``` csharp
// Correctly marks whole method return type as nullable because one of the success status codes returns a nullable type.
public virtual async System.Threading.Tasks.Task<MyClass?> Get(System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) 
{
    if (status_ == 200)
    {
        var objectResponse_ = await ReadObjectResponseAsync<InlPerson>(response_, headers_, cancellationToken).ConfigureAwait(false);
        // Note that there is still a null check for a 200 result, which is correct because I did not mark my action result as nullable.
        if (objectResponse_.Object == null)
        {
            throw new ApiException("Response was null which was not expected.", status_, objectResponse_.Text, headers_, null);
        }
    }
    else
    if (status_ == 204)
    {
        var objectResponse_ = await ReadObjectResponseAsync<InlPerson?>(response_, headers_, cancellationToken).ConfigureAwait(false);
        // Correctly returns the value, which will be null for 204.
    }
}
```

### DIFF
--- a/src/NSwag.CodeGeneration/Models/ResponseModelBase.cs
+++ b/src/NSwag.CodeGeneration/Models/ResponseModelBase.cs
@@ -15,7 +15,6 @@ namespace NSwag.CodeGeneration.Models
     public abstract class ResponseModelBase
     {
         private readonly IOperationModel _operationModel;
-        private readonly OpenApiResponse _response;
         private readonly OpenApiOperation _operation;
         private readonly JsonSchema _exceptionSchema;
         private readonly IClientGenerator _generator;
@@ -37,7 +36,7 @@ namespace NSwag.CodeGeneration.Models
             string statusCode, OpenApiResponse response, bool isPrimarySuccessResponse,
             JsonSchema exceptionSchema, TypeResolverBase resolver, CodeGeneratorSettingsBase settings, IClientGenerator generator)
         {
-            _response = response;
+            Response = response;
             _operation = operation;
             _exceptionSchema = exceptionSchema;
             _generator = generator;
@@ -50,6 +49,9 @@ namespace NSwag.CodeGeneration.Models
             ActualResponseSchema = response.Schema?.ActualSchema;
         }
 
+        /// <summary>The underlying response</summary>
+        internal OpenApiResponse Response { get; }
+
         /// <summary>Gets the HTTP status code.</summary>
         public string StatusCode { get; }
 
@@ -61,14 +63,14 @@ namespace NSwag.CodeGeneration.Models
 
         /// <summary>Gets the type of the response.</summary>
         public string Type =>
-            _response.IsBinary(_operation) ? _generator.GetBinaryResponseTypeName() :
+            Response.IsBinary(_operation) ? _generator.GetBinaryResponseTypeName() :
             _generator.GetTypeName(ActualResponseSchema, IsNullable, "Response");
 
         /// <summary>Gets a value indicating whether the response has a type (i.e. not void).</summary>
         public bool HasType => ActualResponseSchema != null;
 
         /// <summary>Gets or sets the expected child schemas of the base schema (can be used for generating enhanced typings/documentation).</summary>
-        public ICollection<JsonExpectedSchema> ExpectedSchemas => _response.ExpectedSchemas;
+        public ICollection<JsonExpectedSchema> ExpectedSchemas => Response.ExpectedSchemas;
 
         /// <summary>Gets a value indicating whether the response is of type date.</summary>
         public bool IsDate => ActualResponseSchema != null &&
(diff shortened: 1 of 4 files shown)


Banner-Keith:
@RicoSuter & @lahma I took a stab at addressing part of the problem in the linked issue. Please review and let me know your thoughts. Happy to make any changes needed.

Cpcrook:
This change would solve the issues I've been running into and I'd love to see this integrated sooner rather than later.

src/NSwag.CodeGeneration/Models/ResponseModelBase.cs
```diff
@@ -110,9 +112,11 @@ public bool IsSuccess
                 }
 
                 var primarySuccessResponse = _operationModel.Responses.FirstOrDefault(r => r.IsPrimarySuccessResponse);
+
+                // We should ignore nullability when evaluating if both responses have the same return type.
                 return HttpUtilities.IsSuccessStatusCode(StatusCode) && (
```
markusahrweileramcon, line 117: IMHO, this check is factually incorrect.
A response is a success, when the status code is between 200 and 300 (400), no?
A 204 has by design no content, which means `Data?` is wrong as response type.
It should be void.

Also, if someone would want to send a string on 201, but an object on 200, this check would still mark the response as failure

markusahrweileramcon: I guess, we could introduce an option which says to treat only the same response types as success which is true by default. Not sure, if this class can have easy access to the options

Banner-Keith:
Abandoning this PR. We have moved away from NSwag and are now using Microsoft.Extensions.ApiDescription.Server to generate the open api spec at build time and then a Roslyn Source Generator to generate our client.

I will leave the branch in existence if anyone wants to take this over.

PR #5099, opened 2025-01-31
Suppress compiler warnings for possible nulls in generated code

velvolue:
Silences some warnings that show up when enabling `generateNullableReferenceTypes`.


RicoSuter:
Thanks for the Pr.

Issue #4083, opened 2022-07-11
Feature: System.Text.Json


tamusjroyce:
Is there work in progress to move to System.Text.Json.Serialization? Not asking create this serialization option feature, as it would be a major change. But curious if it has been considered or is in the works. Was thinking about taking a try at it.

https://docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-migrate-from-newtonsoft-how-to?pivots=dotnet-6-0#table-of-differences-between-newtonsoftjson-and-systemtextjson

As a featurish question, feel free to close at any time (no response necessary. especially if you don't have the time).

Thanks for the great library! Visual Studio built-in support works great, even for Azure Functions calling other Azure Functions.

SebastianStehle:
I think many parts are already independent from the serializer. For you as a developer it does not really matter how OpenAPI documents are serialized.

I have used nswag with JSON.NET and STS and the only thing missing for me is support for inheritance with STS.

pekspro:
There already is some experimental support for System.Text.Json:

![image](https://user-images.githubusercontent.com/3163293/180636522-966b7a5b-94ed-4cc5-ae66-d516bad96b33.png)

oduis:
Careful though, the new SystemTextJson ist still pretty buggy (relly experiemental):
- Sometimes it generates a mix between System.Text.Json and NewtonsoftJson (e.g. having a Serializier and SerializerOptions in Newtonson, while the rest of the file is already correctly using System.Text.Json).
- It also generates properties with JsonIgnoreCondition.WhenWritingNull instead of WhenWritingDefault (which create problems down the road, but I have a bugfix script to simply replace it in the text file)

SkinnySackboy:
Hi,

It's been a while now, is System.Text.Json still experimental in 2024?

oliverw:
> Hi,
> 
> It's been a while now, is System.Text.Json still experimental in 2024?

Would like to know that as well?

oduis:
@oliverw We use it in production for several years now. It works fine from my point of view.

oliverw:
> > It's been a while now, is System.Text.Json still experimental in 2024?
> 
[...]

PR #5102, opened 2025-02-09
Use JsonContent when creating request bodies

AustinWise:
This PR applies to the generated code for clients when using System.Text.Json serialization library.

This PR uses the [`JsonContent`](https://learn.microsoft.com/en-us/dotnet/api/system.net.http.json.jsoncontent?view=net-9.0) class to send objects as the request body. This is the [documented way](https://learn.microsoft.com/en-us/dotnet/fundamentals/networking/http/httpclient) to serialize objects when making HTTP requests. It can be more efficient than serializing the entire request body into a byte array, particularly if the request body is large. This is because the `JsonContent` can stream the body while serializing and reuse buffers between requests.

In terms of the backwards compatibility impact of this change, the documentation says `JsonContent` was added in .NET 5. When running on .NET Framework, a reference to the `System.Net.Http.Json` Nuget package would now be required.

Contributes to #4083

### DIFF
--- a/src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.liquid
+++ b/src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.liquid
@@ -220,9 +220,11 @@
 {%         elsif operation.HasPlainTextBodyParameter -%}
                 var content_ = new System.Net.Http.StringContent({{ operation.ContentParameter.VariableName }});
                 content_.Headers.ContentType = System.Net.Http.Headers.MediaTypeHeaderValue.Parse("{{ operation.Consumes }}");
+{%         elsif UseSystemTextJson -%}
+                var content_ = System.Net.Http.Json.JsonContent.Create({{ operation.ContentParameter.VariableName }}, System.Net.Http.Headers.MediaTypeHeaderValue.Parse("{{ operation.Consumes }}"), {% if UseRequestAndResponseSerializationSettings %}Request{% endif %}JsonSerializerSettings);
 {%         else -%}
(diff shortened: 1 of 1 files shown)


AustinWise:
I tested this PR with OpenAPI specification for Ollama I'm using. I confirmed that:

* Code generation is unchanged when using NewtonSoft.
* The code compiles and sends a request successfully with the new System.Text.Json code generation.

Here is what a diff of the generated code looks like:


```diff
                 using (var request_ = new System.Net.Http.HttpRequestMessage())
                 {
-                    var json_ = System.Text.Json.JsonSerializer.SerializeToUtf8Bytes(body, JsonSerializerSettings);
-                    var content_ = new System.Net.Http.ByteArrayContent(json_);
-                    content_.Headers.ContentType = System.Net.Http.Headers.MediaTypeHeaderValue.Parse("application/json");
+                    var content_ = System.Net.Http.Json.JsonContent.Create(body, System.Net.Http.Headers.MediaTypeHeaderValue.Parse("application/json"), JsonSerializerSettings);
                     request_.Content = content_;
```

RicoSuter:
Thanks for the PR. 
Is this behaving exactly the same? Is this only available with newer .NET versions?

AustinWise:
I added some snapshot tests to ensure the generated code compiles.

> Is this behaving exactly the same?

I believe so. I ensured that that Content-Type header is the same. The implementation of `JsonContent` is pretty small, ultimately it also calls into `JsonSerializer`, [see here](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Net.Http.Json/src/System/Net/Http/Json/JsonContent.cs#L119).

One change is users targeting .NET Standard or .NET Framework will have to add a reference to the [System.Net.Http.Json Nuget package](https://www.nuget.org/packages/System.Net.Http.Json).

> Is this only available with newer .NET versions?

The [`JsonConten` type](https://learn.microsoft.com/en-us/dotnet/api/system.net.http.json.jsoncontent?view=net-9.0) has been include with .NET since version 5. When targeting .NET Standard or .NET Framework, the [package that provides the type](https://www.nuget.org/packages/System.Net.Http.Json/9.0.7#supportedframeworks-body-tab) has the same supported frameworks as [the package that provides the main System.Text.Json serializer](https://www.nuget.org/packages/System.Text.Json/9.0.7#supportedframeworks-body-tab).

Issue #3929, opened 2022-03-09
Generated Client NRT Warnings


csehatt741:
When the request body is required and nullable reference types are allowed in the project containing the generated code, warnings are raised in the generated client.

Command: openapi2csclient
JsonLibrary: SystemTextJson
NSwag version: 13.15.10.0

OpenAPI:
```yaml
requestBody:
  required: true
  content:
    application/json:
      schema:
```

```csharp
if (status_ == 200)
{
    var responseData_ = response_.Content == null ? null : await response_.Content.ReadAsStringAsync().ConfigureAwait(false);
    var result_ = (ResponseType)System.Convert.ChangeType(responseData_, typeof(ResponseType));
    return new SwaggerResponse<ResponseType>(status_, headers_, result_);
}
```

**Required fixes:**
`var responseData_ = response_.Content == null ? null : await response_.Content.ReadAsStringAsync().ConfigureAwait(false);` ->

```csharp
if (response_.Content == null)
{
    string responseText_ = (response_.Content == null) ? string.Empty : await response_.Content.ReadAsStringAsync().ConfigureAwait(false);
    throw new ApiException("A non - empty request body is required.", status_, responseText_, headers_, null);
}

var responseData_ = await response_.Content.ReadAsStringAsync().ConfigureAwait(false);
```

FlorianArnould:
I have the same kind of warning using an equivalent configuration: 

Command: openapi2csclient
JsonLibrary: Newtonsoft.Json
NSwag version: 13.15.10.0

I have a route responding a `text/plain` content-type, I have this code generated :
```csharp
if (status_ == 200)
{
    var responseData_ = response_.Content == null ? null : await response_.Content.ReadAsStringAsync().ConfigureAwait(false);
    var result_ = (string)System.Convert.ChangeType(responseData_, typeof(string)); // [CS8600] Converting null literal or possible null value to non-nullable type.
    return result_;
}
```
and this is giving me this warning :
[CS8600] Converting null literal or possible null value to non-nullable type.

FlorianArnould:
Still the same with NSwag.MSBuild 13.16.1

ariatpi:
Still seeing this in 13.18.0

ogazioglu:
In according to the definition of `ChangeType` method:
```csharp
[return: NotNullIfNotNull(nameof(value))]
public static object? ChangeType(object? value, Type conversionType);
```
marking `responseData_` with `!` like below can resolve this case :

```csharp
[...]

PR #5108, opened 2025-02-17
Cast result to nullable type for response data #3929

ogazioglu:


RicoSuter:
Is there a fix without using ? NRT as the generator should also work with pre-NRT C# versions?

PR #5121, opened 2025-03-17
Upgraded Microsoft.Extensions.ApiDescription.* to 9.0.3

ntyrrell:
Upgraded Microsoft.Extensions.ApiDescription.Server & Microsoft.Extensions.ApiDescription.Client to 9.0.3

This would update the runtime dependencies (which are still dependent on AspNetCore 6.0.3) as well as the following bugfixes for Microsoft.Extensions.ApiDescription.Server:
[Fix up packaging for Extensions.ApiDescription.Server](https://github.com/dotnet/aspnetcore/pull/44136)
[Fixes for OpenAPI document generation with M.E.ApiDescription.Server](https://github.com/dotnet/aspnetcore/pull/57096)
[Upgrade to Microsoft.OpenApi 2.x and support OpenAPI v3.1](https://github.com/dotnet/aspnetcore/pull/59480)
[#59749 move TrimEnd after path normalization](https://github.com/dotnet/aspnetcore/pull/59753)
And the following for Microsoft.Extensions.ApiDescription.Client:
[#59749 move TrimEnd after path normalization](https://github.com/dotnet/aspnetcore/pull/59753)
[Fix duplicate add of items to Compile on Microsoft.Extensions.ApiDescription.Client.targets](https://github.com/dotnet/aspnetcore/pull/51846)

### DIFF
--- a/src/NSwag.ApiDescription.Client/NSwag.ApiDescription.Client.nuspec
+++ b/src/NSwag.ApiDescription.Client/NSwag.ApiDescription.Client.nuspec
@@ -15,7 +15,7 @@
     <repository type="git" url="https://github.com/RicoSuter/NSwag.git"/>
     <developmentDependency>true</developmentDependency>
     <dependencies>
-      <dependency id="Microsoft.Extensions.ApiDescription.Client" version="6.0.3" />
+      <dependency id="Microsoft.Extensions.ApiDescription.Client" version="9.0.3" />
       <dependency id="NSwag.MSBuild" version="1.0.0" />
     </dependencies>
     <references />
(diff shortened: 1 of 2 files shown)


lahma:
I think you can get these also by just referencing the latest versions in the consuming project?

RicoSuter:
Not sure whether this is a good idea, i think this might break e.g. .NET 8 users...
Isnt this package's version defined by the host application?

lahma:
My intuition would say that lowest non-vulnerable LTS would be a good approach, like 8.0.0. Some corporates might get confused if they are tied to say NET 8 LTS and 9.x dependencies start to show up.

ntyrrell:
Thanks for the feedback. I'll adjust it to 8.0.* (not sure what the latest patch number is off hand)

RicoSuter:
Thanks for the PR.

Issue #4396, opened 2023-04-12
Generate client that does not compile


nicomartinezsage:
We are using NSwag version 13.18.2 to generate a client for the CloudFlare api.
The command used is

```cmd
openapi2csclient /input:"https://raw.githubusercontent.com/cloudflare/api-schemas/main/openapi.json" /namespace:SampleNameSpace.Client.DecoratedProxy /output:"C:/.automation/nswag/Proxy.generated.cs" /clientClassAccessModifier:internal /typeAccessModifier:public  /generateOptionalPropertiesAsNullable:true /generateNullableReferenceTypes:true
```

but the generated code cannot be compiled, 

it generates methods like;

``` csharp
        /// <summary>
        /// Create Secondary Zone Configuration
        /// </summary>
        /// <remarks>
        /// Create secondary zone configuration for incoming zone transfers.
        /// </remarks>
        /// <returns>Create Secondary Zone Configuration response</returns>
        /// <exception cref="ApiException">A server side error occurred.</exception>
        public virtual System.Threading.Tasks.Task<Single_response_incoming> SecondaryDns(SecondaryZone)CreateSecondaryZoneConfigurationAsync(IdentifierKjRn97d3 zone_identifier, DnsSecondarySecondaryZone body)
        {
            return SecondaryDns(SecondaryZone)CreateSecondaryZoneConfigurationAsync(zone_identifier, body, System.Threading.CancellationToken.None);
        }
```

methodname with `(`

pbolduc:
The reason this occurs is because CloudFlare has those characters in the operationId,

```json
"operationId":"secondary-dns-(-secondary-zone)-create-secondary-zone-configuration"
```

The OpenAPI spec just says the operationId must be unique but does not constrain the valid characters. NSwag must not of expected these extra characters. I guess the function that converts from operationId to a method should take this into account.

joriswit:
See also #2447

Issue #2447, opened 2019-10-05
Remove invalid characters in path segments when generating operation names


cogwire:
Some OpenApi paths include characters which are not valid in a C# identifier and should not be included in the generated operation name.

Here is an example:

```javascript
"/path1/path2/>pathWithInvalidPrefixChar": {
      "get": {
        "description": "",
        "operationId": "pathWithInvalidPrefixChar",
...
```

In this example, the ">" character should not be included in the generated operation name.

josundt:
When using route path based API versioning with version numbers with dot(s), generation fails for C#. I believe this will be a showstopper for a lot of people:

Example:
```js
{
// ...
  "/api/v2.0/vaules": {
      "get": {
        "description": "",
        "operationId": "pathWithInvalidPrefixChar",
        // ...
    },
    // ...
}
```

This is a very unfortunate bug that blocks us from using Connected Services -> OpenApi in Visual Studio with our APIs.

ArXen42:
Encountered similar problem with typescript client generator. Some of my API endpoints have routes like `.../preview.jpg` which results in invalid code being generated.

RicoSuter:
We'd need to fix that in the path OperationNameGenerator

jeanluc162:
I'm trying to generate Code for an API where one of the Operations Starts with '$', meaning the YAML
```
/$batch:
    post:
```
generates
```
public virtual System.Threading.Tasks.Task<string> $batchAsync(string body)
```
which can't be compiled ('$' is an illegal token). I think this character should be replaced as well.

PR #5123, opened 2025-03-19
#2447 Replace operation name characters which are invalid in code identifiers

joriswit:
A fix for the issue that some api's use characters in operation names that are invalid identifiers, resulting in code that does not compile.
Fix for #2447, #4396


src/NSwag.CodeGeneration/OperationNameGenerators/OperationNameGeneratorHelpers.cs
```diff
@@ -0,0 +1,62 @@
...
+        /// <remarks>
+        /// Because this function is used for both C# as TypeScript, some valid identifiers will be changed because they are invalid in the other. E.g. $test is valid TypeScript, but not C#.
+        /// </remarks>
+        /// <param name="identifier">The identifier.</param>
+        /// <returns>The fixed identifier</returns>
+        public static string FixIdentifier(string identifier)
+        {
+            if (identifier.Length > 0)
+            {
+                var validCategoriesFirstChar = new[] { UnicodeCategory.UppercaseLetter, UnicodeCategory.LowercaseLetter, UnicodeCategory.TitlecaseLetter, UnicodeCategory.ModifierLetter, UnicodeCategory.OtherLetter, UnicodeCategory.LetterNumber };
```
lahma, line 28: This does not seem ideal. You are allocating arrays on every invocation even though they could be static class fields. Also using `Contains` will use LINQ version instead of more performant array checks.

joriswit: Thanks for your review! I have applied your suggestions.

joriswit:
The tests made me realize I have added the invalid character checks in the wrong place. I added it where the current check for the `.` was replaced with the `_`. But this is too early in the process. I have changed it so the check is done much later.

joriswit:
Sorry, I had not read the Contributor License Agreement. I don't know if my employer approves this. I'm closing this. Do not merge.

PR #5158, opened 2025-04-26
Fix array parameter explosion check

### LINKED ISSUE #5133: Explode property for query parameter in C# code generator
I'm using NSwag version 14.3.

In my OpenAPI 3.0 specification, the following query parameter is defined:

`
{
  "name": "state",
  "in": "query",
  "description": "Appointment state",
  "required": false,
  "style": "form",
  "explode": false,
  "schema": {
    "type": "array",
    "items": {
      "$ref": "#/components/schemas/AppointmentState"
    }
  }
}
`

However, the generated code always serializes the array like this:

/appointments/?state=FREE&state=ACCEPTED

But I want it to be serialized like this:

/appointments/?state=FREE,ACCEPTED

I found a ticket (https://github.com/RicoSuter/NSwag/pull/5016) that was apparently merged in version 14.3, which specifically addresses this serialization issue. What am I doing wrong here, or is this still a bug?

lahma:
fixes #5133


KirkMunroSagent:
@RicoSuter This is a breaking change, and it is implemented incorrectly as well.

With this change, the nswag generator changes the way it generated a C# client from a swagger doc. I have endpoints with query parameters that are arrays, and before this change they would correctly generate as ?myArray=val1&myArray=val2&myArray=val3, but now they generate as ?myArray=val1,val2,val3. Per [the documentation](https://learn.microsoft.com/en-us/dotnet/api/microsoft.openapi.models.openapiparameter.explode#microsoft-openapi-models-openapiparameter-explode), "When style is form, the default value is true. For all other styles, the default value is false." If you then read on the [Style property](https://learn.microsoft.com/en-us/dotnet/api/microsoft.openapi.models.openapiparameter.style), the docs say "Default values (based on value of in): for query - form; for path - simple; for header - simple; for cookie - form.", so, for query parameters the default style is form, which means the default value of explode is true, as it was before. Please have someone fix this breaking change so that query parameters work as they should per the documentation.

lahma:
> Please have someone fix this breaking change so that query parameters work as they should per the documentation.

@KirkMunroSagent can you be the one? A PR with new test coverage showing the desired way of working.

KirkMunroSagent:
> > Please have someone fix this breaking change so that query parameters work as they should per the documentation.
> 
> @KirkMunroSagent can you be the one? A PR with new test coverage showing the desired way of working.

I'm on vacation, so there's no likelihood that I'll have time for this anytime soon.

This PR is a breaking change and should be reverted. It goes against the documentation, so the simplest way to fix this issue that didn't exist before is to create another PR that reverts this set of changes and then leave it until someone can re-do them properly while respecting the documentation.

NenadSteric:
we also ran into this issue. 
current version of nswag (14.4.0.0) is not generating correct c# code for array parameters.

KollarM:
Same here.  The current version 14.4.0.0 is not generating correct c# code for array parameters.

AFAde:
Hello, 
I am experiencing the same error. Is a fix supposed to be released any time soon?

Thanks

schauers:
Hello,
I am experiencing the same error. Is a fix supposed to be released any time soon?

Thanks

PR #5170, opened 2025-05-13
Better support for reading yaml from a stream

simonthum:
This simply exposes one more overload of to be able to pass in a TextReader, a subtype of which would be created from the data string anyway.

Using this overload properly will make the parser run while the stream is being read, so completion will happen earlier.


lahma (approved):
Looks good, thank you!

lahma:
You can get a preview from myget feed, no release plans yet.

simonthum:
I can only find

NSwag.Core.Yaml                 | 14.4.0-preview-20250428-2155

Could you please trigger the prerelease?

lahma:
I've incremented the version number and fixed some build issues, hopefully there will be a new pre-release soon(ish).

simonthum:
Thanks a lot!

PR #5171, opened 2025-05-14
Better support for reading yaml from a stream (part two)

simonthum:
I figured I would expect the TextReader overloads to be first-class, so I could just as well add them.

SFTN.


lahma (approved):
Looks good, thank you!

PR #5173, opened 2025-05-15
Fixed template to generate non-deprecated code for _observableThrow method

MohitGhodke:
Currently generated code for Framework.RxJs.ObservableThrowMethod does not comply with RxJs 7+ versions.
_observableThrow is generated as deprecated for the same reason.
Fixed Angular client template to generate non--deprecated code.


RicoSuter:
can you fix the build actions, probably the snapshots need to be updated?

PR #5176, opened 2025-05-21
Fix to explode array parameters implicitly when SchemaType is OpenAPI 3.0

### LINKED ISSUE #5165: Code generated by NSwag.ApiDescription.Client 14.4.0 generates incorrect URLs
**Describe the bug**

Code generated with *NSwag.ApiDescription.Client* 14.4.0 generates urls with enumerable **int** parameters a way that the standard WebApi model binder cannot handle.

```
using System.Globalization;

var builder144 = new System.Text.StringBuilder();

int[]? fundingSchemeIds = { 1, 2, 3 };

if (fundingSchemeIds != null)
{
    builder144.Append(System.Uri.EscapeDataString("FundingSchemeIds") + "=");
    foreach (var item_ in fundingSchemeIds)
    {
        builder144.Append(System.Uri.EscapeDataString(ConvertToString(item_, System.Globalization.CultureInfo.InvariantCulture))).Append(",");
    }
    builder144.Length--;
    builder144.Append("&");
}

Console.WriteLine("{0} // 14.4.0", builder144);

var builder143 = new System.Text.StringBuilder();

if (fundingSchemeIds != null)
{
    foreach (var item_ in fundingSchemeIds) { builder143.Append(System.Uri.EscapeDataString("FundingSchemeIds")).Append('=').Append(System.Uri.EscapeDataString(ConvertToString(item_, System.Globalization.CultureInfo.InvariantCulture))).Append('&'); }
}


Console.WriteLine("{0} // 14.3.0", builder143);
Console.Read();

string ConvertToString(int i, CultureInfo cultureInfo)
{
    return $"{i}";
}

```

Output
```
FundingSchemeIds=1,2,3& // 14.4.0 --> nok
FundingSchemeIds=1&FundingSchemeIds=2&FundingSchemeIds=3& // 14.3.0 --> ok
```

While the endpoint looks like
```
    [HttpGet(Name = nameof(SearchGrantsAsync))]
    [AllowAnonymous]
    [ProducesResponseType(StatusCodes.Status400BadRequest)]
    [ProducesResponseType(typeof(SearchResult), StatusCodes.Status200OK)]
    public async Task<ActionResult<SearchResult>> SearchGrantsAsync([FromQuery] IEnumerable<int> fundingSchemeIds, CancellationToken cancellationToken)
```

Somehow similar to: https://github.com/RicoSuter/NSwag/issues/5096 (but the version 14.2.0 seems off).

Nuget package NSwag.ApiDescription.Client --version 14.4.0

**To Reproduce**

Visual Studio 2022 (Version 17.13.6) right click on project using the service > Add > Connected Service > Service References > (Existing) Refresh

**Expected behavior**

The url generated in the client still works as it did before in Version 14.3.0

Phoosha:
Fixes #5165

## Problem

There's a regression, where for certain OpenAPI 3.0 documents that used to work, the C# code generated by NSwag for array query parameters is not anymore compatible with ASP.NET Core.
 - That's because ASP.NET Core always expects exploded array parameters (i.e. `?ids=a&ids=b&ids=c`).
 - However, NSwag generates un-exploded array parameters (i.e. `?ids=a,b,c`), _**when**_ an OpenAPI 3.0 document does not contain the properties `style` and `explode`.
  
These properties are for example missing when Swashbuckle.AspNetCore.SwaggerGen was used to generate the OpenAPI document.

## Expected behavior

The expected behavior is different for OpenAPI 3.0 vs Swagger 2, particularly regarding the implicit behavior in the absence of parameters like `style` and `explode` resp. `collectionFormat`.

Therefore _**it is necessary to respect the document schema**_ to implement the expected behavior according to the OpenAPI 3.0 vs. Swagger 2 specifications.

**OpenAPI 3.0:**
- `style` and `explode` are used (instead of `collectionFormat)
- both are allowed to be absent
- default values (when absent) depend on the kind of parameter (i.e. `in` or `ParameterModelBase.Kind`)
  - `style: form` and `explode: true` (i.e. exploded array) are the defaults for
    - Query Parameters (i.e. `in: query`)
    - Cookie Parameters (i.e. `in: cookie`)
  - `style: simple` and `explode: false` is the default in every other case 

[Source](https://swagger.io/docs/specification/v3_0/serialization/#query-parameters)

**Sagger 2:**
- `collectionFormat` is used (instead of `style` and `explode`)
- it is allowed to be absent
- `csv` (i.e. un-exploded array) is the default value (when absent) and does not depend

[Source](https://swagger.io/docs/specification/v2_0/describing-parameters/#array-and-multi-value-parameters)

## Current behavior

The document schema is not respected when deciding whether to explode array parameters. Instead when all parameters (i.e. `style`, `explode` and `collectionFormat`) are absent, then the array is not exploded. Therefore, the implicit assumption is that the document schema is Swagger 2, even if it is not.

## Context

- The current behavior was introduced by the following PR:
  https://github.com/RicoSuter/NSwag/pull/5158
- A comment in that PR also describes the issue addressed by my PR:
  https://github.com/RicoSuter/NSwag/pull/5158#issuecomment-2839717451
- The above PR improved the behavior introduced by the following less recent PR:
  https://github.com/RicoSuter/NSwag/pull/5016

### DIFF
--- a/src/NSwag.CodeGeneration.CSharp.Tests/ArrayParameterTests.cs
+++ b/src/NSwag.CodeGeneration.CSharp.Tests/ArrayParameterTests.cs
@@ -71,6 +71,296 @@ namespace NSwag.CodeGeneration.CSharp.Tests
             CodeCompiler.AssertCompile(code);
         }
 
+        [Theory]
+        [InlineData(@"""explode"": false,")]
+        [InlineData(@"""style"": ""form"", ""explode"": false,")]
+        public async Task When_explode_is_false_then_do_not_explode(string explode)
+        {
+            // Arrange
+            var json = @"{
+              ""x-generator"": ""NSwag v13.5.0.0 (NJsonSchema v10.1.15.0 (Newtonsoft.Json v11.0.0.0))"",
+              ""openapi"": ""3.0.4"",
+              ""info"": {
+                ""title"": ""My Title"",
+                ""version"": ""1.0.0""
+              },
+              ""paths"": {
+                ""/removeElement"": {
+                  ""delete"": {
+                    ""tags"": [
+                      ""api""
+                    ],
+                    ""operationId"": ""removeElement"",
+                    ""parameters"": [
+                      {
+                        ""name"": ""X-User"",
+                        ""in"": ""header"",
+                        ""description"": ""User identifier"",
+                        ""required"": true,
+                        ""schema"": {
+                          ""type"": ""string""
+                        }
+                      },
+                      {
+                        ""name"": ""elementId"",
+                        ""in"": ""query"",
+                        ""description"": ""The ids of existing elements that should be removed"",
+                        ""required"": false,
+                        " + explode + @"
+                        ""schema"": {
+                          ""type"": ""array"",
+                          ""items"": {
+                            ""type"": ""integer"",
+                            ""format"": ""int64""
+                          }
+                        }
+                      }
+                    ],
+                    ""responses"": {
+                      ""200"": {
+                        ""description"": """",
+                        ""content"": {
+                          ""application/octet-stream"": {
+                            ""schema"": {
+                              ""type"": ""string"",
+                              ""format"": ""binary""
+                            }
+                          }
+                        }
+                      }
+                    }
+                  }
+                },
+    
+              },
+              ""components"": {}
+            }";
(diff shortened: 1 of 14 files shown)


Valkozaur:
Will this be released soon?

lahma:
You should be able to get a preview from MyGet feed.

Issue #5186, opened 2025-06-09
CancellationToken should be passed to ReadAsStringAsync/ReadAsStreamAsync


arshrp:
The code generated by NSwag allows the consumer to pass a CancellationToken, which it uses on the call to HttpClient.SendAsync, but it doesn't use on the downstream call to response.Content.ReadAsStringAsync (or ReadAsStreamAsync).

This is a potential issue, because the NSwag-generated call to SendAsync uses HttpCompletionOption.ResponseHeadersRead, meaning the SendAsync call will return as soon as the headers are read (it will not wait for the full response body to be read, that will happen on the subsequent call to response.Content.ReadAsStringAsync).

The [Microsoft Documentation](https://learn.microsoft.com/en-us/dotnet/api/system.net.http.httpcompletionoption?view=net-9.0) suggests the use of the HttpCompletionOption.ResponseHeadersRead option means the following call to ReadAsStringAsync could block, because the CancellationToken (as well as the default timeout) only applies to the SendAsync call.

I am seeing this behavior at my end. Http log statements before and after the SendAsync call are appearing, but nothing afterward. My best guess is the host service becomes unresponsive during the ReadAsStringAsync call, causing an infinite block.

PR #5187, opened 2025-06-09
Adding cancellationToken to all calls to ReadAsStringAsync() and ReadAsStreamAsync()

arshrp:
Adding cancellationToken to all calls to ReadAsStringAsync() and ReadAsStreamAsync()

As per #5186

### DIFF
--- a/src/NSwag.CodeGeneration.CSharp.Tests/FileDownloadTests.cs
+++ b/src/NSwag.CodeGeneration.CSharp.Tests/FileDownloadTests.cs
@@ -36,7 +36,7 @@ namespace NSwag.CodeGeneration.CSharp.Tests
 
             // Assert
             Assert.Contains("System.Threading.Tasks.Task<FileResponse> DownloadFileAsync();", code);
-            Assert.Contains("ReadAsStreamAsync()", code);
+            Assert.Contains("ReadAsStreamAsync(cancellationToken)", code);
         }
 
         [Fact]
--- a/src/NSwag.CodeGeneration.CSharp.Tests/Snapshots/ArrayParameterTests.When_CollectionFormat_is_csv_then_do_not_explode.verified.txt
+++ b/src/NSwag.CodeGeneration.CSharp.Tests/Snapshots/ArrayParameterTests.When_CollectionFormat_is_csv_then_do_not_explode.verified.txt
@@ -149,7 +149,7 @@ namespace MyNamespace
 
             if (ReadResponseAsString)
             {
-                var responseText = await response.Content.ReadAsStringAsync().ConfigureAwait(false);
+                var responseText = await response.Content.ReadAsStringAsync(cancellationToken).ConfigureAwait(false);
                 try
                 {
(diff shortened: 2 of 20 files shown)


arshrp:
@lahma can you review this PR, it is pretty simple.  Not sure what the issue is with the windows check failing, I don't think it has anything to do with my changes, and I'm reluctant to apply the fix suggested by CoPilot:

_Update MSBuild to Match the SDK Version Ensure that the MSBuild version used in the workflow supports .NET 9.0.204. You may need to install or update the Visual Studio Build Tools to align with this SDK._

lahma:
Do you see any possible SDK targets that would not have an overload that takes cancellation token?

arshrp:
@lahma I don't think that's the issue with the check.  Both Dotnet 8 and 9 support the cancellation token parameter on both the ReadAsStringAsync and ReadAsStreamAsync methods.  

And if SDK incompatibility were the issue, all 3 checks would have failed.  Only the windows-latest check failed, while ubuntu-latest and macos-latest both passed.

Here's the error from the windows-latest check failure:

C:\Program Files\dotnet\sdk\9.0.300\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.Sdk.targets(262,5): error NETSDK1216: Package Microsoft.Net.Sdk.Compilers.Toolset is not downloaded but it is needed because your MSBuild and SDK versions are mismatched. Ensure version 9.0.300 of the package is available in your NuGet source feeds and then run NuGet package restore from Visual Studio or MSBuild. [D:\a\NSwag\NSwag\src\NSwag.Annotations\NSwag.Annotations.csproj::TargetFramework=netstandard2.0]

lahma:
I was more worried about possible full framework usage, like net472.

arshrp:
@lahma Ah, yes, you are right.  The overload with a CancellationToken parameter is only applicable to .NET, and not .NET Framework or .NET Standard.  Any suggestions?  See #5186 for why this fix is needed.

Side note, any idea what's causing the check failure I mentioned?

arshrp:
@lahma What if I have the templates emit preprocessor directives that will support all .NET versions? Ex:

#if NETCOREAPP3_0_OR_GREATER || NET5_0_OR_GREATER
    : await response_.Content.ReadAsStringAsync(cancellationToken).ConfigureAwait(false);
#else
    : await response_.Content.ReadAsStringAsync().ConfigureAwait(false);    
#endif

lahma (approved):
Thanks for checking and improving. Looks good to me.

arshrp:
@lahma Wait, I hadn't added the preprocessor directives, I was only asking if that was a valid solution to the problem you raised...

The branch as is will not work on .NET Framework or .NET Standard, can we "unmerge" until I add those changes?

lahma:
Can you send a second PR to add them? We can do incrementally.

lahma:
> Side note, any idea what's causing the check failure I mentioned?

It's the windows build image GitHub Actions is using, needs some tweaking, unrelated to this.

PR #5189, opened 2025-06-11
Feature/add cancellation token to content read

arshrp:
@lahma Added preprocessor directive as discussed.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -40,11 +40,6 @@ jobs:
         run: |
           echo "Adding GNU tar to PATH"
           echo C:\Program Files\Git\usr\bin>>"%GITHUB_PATH%"
-      - uses: actions/setup-dotnet@v4
-        with:
-          dotnet-version: |
-            8.0
-            9.0.204
       - uses: actions/checkout@v4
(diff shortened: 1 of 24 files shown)


lahma:
You probably need to rebase against master, you have extra commit here.

lahma:
I think it would be cleaner to inject this via liquid model, having just in template something like: `ReadAsync({{ AsyncReadCancellationToken }})` and it would be empty string for NET Framework and "cancellationToken" for others. Now there's a lot of duplication and client code grows.

lahma:
And while thinking a bit more, might be quite hard to know the target moniker for generated client which would be required for Liquid template variables..

arshrp:
> You probably need to rebase against master, you have extra commit here.

@lahma Can you help with this?  I tried rebasing, but the conflicts remain.  I'm no Git expert...

arshrp:
> And while thinking a bit more, might be quite hard to know the target moniker for generated client which would be required for Liquid template variables..

Yes, I did consider this, but as you say, at build time the framework may not be known, so I thought best to leave it dynamic in the generated code.  Your thoughts?

PR #5190, opened 2025-06-13
Adjusted cancellationToken to work for all NSwag supported frameworks

arshrp:
Added preprocessor directive that ensures cancellationToken is used only for frameworks that support it

### DIFF
--- a/src/NSwag.CodeGeneration.CSharp.Tests/FileDownloadTests.cs
+++ b/src/NSwag.CodeGeneration.CSharp.Tests/FileDownloadTests.cs
@@ -36,7 +36,7 @@ namespace NSwag.CodeGeneration.CSharp.Tests
 
             // Assert
             Assert.Contains("System.Threading.Tasks.Task<FileResponse> DownloadFileAsync();", code);
-            Assert.Contains("ReadAsStreamAsync(cancellationToken)", code);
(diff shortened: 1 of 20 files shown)


arshrp:
@lahma I reapplied the changes on a new branch of master and that seems to have avoided the conflicts.  Are you able to approve this PR?  I deleted the older PR and branch.

arshrp:
This is a part of #5186

lahma:
@arshrp I simplified the code noise a bit with helpers, no generated file will also be slightly smaller. We don't need to account for legacy NETCOREAPP3. What are your thoughts?

arshrp:
@lahma looks good!  After double checking, I think the version of ReadAsStringAsync with the cancellationToken overload was introduced in .NET 5.0, so I think your change is necessary.  Thanks!

lahma:
Thanks, generally using anything below NET 6 is very very obsolete, while NET 6 is also unsupported. If one jumps the modern NET train, lowest LTS should be the supported target.

arshrp:
@lahma thanks so much for your help on this.  Any sense of when the next release that contains this PR will be?

lahma:
> @lahma thanks so much for your help on this. Any sense of when the next release that contains this PR will be?

No release date, but you can always get a version from MyGet feed.

Issue #1363, opened 2018-06-02
Generated c# interface for Client with internal modifier on DTO classes should be internal too


jonnybi:
A public interface ist generated for a client that has access modifier set to `internal ` and DTO classes have access modifier `internal ` too. 
The generated interface should have `internal ` modifier too.

RSuter:
The problem is that internal (or any fine grained modifieres) cannot be expressed by Swagger/JSON Schema...

bangonkali:
Interesting. We could generate multiple results instead, one carrying the Swagger/Json Schema and the other is an optional file containing information regarding access modifiers.

RSuter:
I think it would be better to implement a custom json schema processor class (NJsonSchema) which adds a new property, eg ‘x-modifier’ to properties and extend the template engine so that you can easily modify the modifieres with a custom extension...

RSuter:
Ping me if you have time to help here... thx

PolaEdward:
If the DTOs are being generated as internal, then the interface should follow the same, otherwise you will get   
[CS0050] Inconsistent accessibility: return type....
So the simple fix is to use whatever access modifier used for DTOs for the interface also.

pbolduc:
I think this file [Client.Interface.liquid](https://github.com/RicoSuter/NSwag/blob/c0e16f9f6e4d692d3ab8dc6c7b2ce5e53498841c/src/NSwag.CodeGeneration.CSharp/Templates/Client.Interface.liquid#L3) would need to use the `TypeAccessModifier` (The DTO class/enum access modifier). However, I was unable find the code/template that generates the DTO model types to verify the correct variable. Where are the models generated?

pbolduc:
@jmevel I am trying to come up with the possible rules. I am not sure if there are other valid combinations.  I want to ensure any change covers all the use cases.  But it does seem the generated interface must use the same assess modifier as the model.

[Access Modifiers](https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/access-modifiers)

## Valid  Modifiers:
- public
- internal

## Invalid Modifiers

All of these modifieds will error out with _CS1527 Elements defined in a namespace cannot be explicitly declared as private, protected, protected internal, or private protected_

- private
- protected
- protected internal
- private protected

## Valid Combinations

| class    | model    | interface  |    |
| -------- | -------- | ---------- | -- |
| public   | public   | public     |    |
[...]

PR #4820, opened 2024-03-17
Allow setting of the client interface access modifier based on the model access modifier

pbolduc:
This PR allows changing the client interface access modifier.  This PR is to address #1363.

 Changes:

- When run from the command line, the model type `TypeAccessModifier` is used. When `TypeAccessModifier` is set in the `OpenApiToCSharpClientCommand`, it sets the settings `ClientInterfaceAccessModifier` to the same value. This is required because `TypeAccessModifier` is found in the Newtonsoft library and not in `CSharpClientGeneratorSettings`.
- `CSharpClientGeneratorSettings` defaults `ClientInterfaceAccessModifier` to `public`
- Updates the C# `Client.Interface.liquid` to use the new  `ClientInterfaceAccessModifier` setting
- Add unit tests to ensure changing the access modifier is correctly changed in the generated C#.

This PR does not address valid combination of setting the generated C# class and model to public, but the interface internal. This would require additional work to expose additional arguments to the command line tool and changes to the UI tools. Someone would need to justify the use case that make this additional combination valid and the effort to implement.

Valid in the table means it is valid C# and will compile.  This PR really addresses the last item in the table.

| valid | class    | model    | interface  |  Handled?  |
|:---:|:-------- |:-------- |:---------- | -- |
| Yes | public   | public   | public     | Yes - worked this way before |
| Yes | public   | public   | internal | No - use case? |
| No | public   | internal | public    | No - not valid C# |
| No | public   | internal | internal | No - not valid C# |
| Yes | internal | public    | public   | Yes - worked this way before |
| Yes | internal | public    | internal | No - edge case? |
| No | internal |  internal  | public | No - not valid C# |
| Yes | internal |  internal  | internal  | **Yes** - This PR fixes this specific combination |

PR #5195, opened 2025-06-27
Make the access modifier for the client interface match the client

norwd:
This makes the `Client.Interface.liquid` template match the [`Client.Class.liquid`](https://github.com/RicoSuter/NSwag/blob/3cc1c36c7f9d632174bd6b478311879cb343251a/src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.liquid#L3) template. This will prevent CS0050 from occurring when the `clientClassAccessModifier` and `typeAccessModifier` settings are set to `internal`.

Currently CS0050 is thrown when these settings are set to `internal` because the interface is `public` while the types in the method signatures are `internal`.


norwd:
Resolves #1363 and #5105

lahma:
Thanks for the suggestion, #4820 won here as it was first and also contains tests for the change.

norwd:
@lahma, no worries, glad I could revitalise #1363 and get it resolved 🙂

Any idea when this might be released?

lahma:
My stock answer.. don't know about next release but you can always get a preview from MyGet...

Issue #2747, opened 2020-03-25
Reusable request bodies are not handled correctly


i-sinister:
According to [OpenApi 3.0 spec](https://swagger.io/docs/specification/describing-request-body/) "reusable request bodies" are supported. However with latest NSwag (version 13.3.0.0) controllers are not generated correctly - action argument is always of type 'object':
```yaml
openapi: '3.0.0'
paths:
  /bugs:
    post:
      summary: Returns list on NSwagStudio bugs
      requestBody:
        $ref: "#/components/requestBodies/NSwagStudioBug"
      responses:
        '200':
          $ref: '#/components/responses/NoContent'
components:
  requestBodies:
    description: Describes NSwagStudio bug
    required: true
    content:
      application/json:
        schema:
          type: string
  responses:
    NoContent:
      description: No content
```
and here is the generated code generated for action:
```c#
[Microsoft.AspNetCore.Mvc.HttpPost, Microsoft.AspNetCore.Mvc.Route("bugs")]
public System.Threading.Tasks.Task Bugs([Microsoft.AspNetCore.Mvc.FromBody] object body)
{
    return _implementation.BugsAsync(body);
}
```

Expected: body type should be 'string'.

JordyThien:
Seems like a duplicate of https://github.com/RicoSuter/NSwag/issues/2122 
However, the problem is not fixed

Doomic:
your example is not correct.. it state `$ref: "#/components/requestBodies/NSwagStudioBug"`. But "NSwagStudioBug" key is missing in the requestBodies definition.
However; the issue remains (i have the same issue) when you wrap the requestbody in a NSwagStudioBug.
I will make a pullrequest to fix this.. hopefully it will be approved soon

Doomic:
mm i just found #4747 .. I think it fixes this isse already... not sure why it is still not approved.

@JordyThien ; i think that #2122 adds support for $ref inside the body (like params).. not for a $ref for the body itself.

PR #4747, opened 2024-01-31
Enhance support for $ref in requestBody of endpoint

Eruzo:
This pull request enhances support for the `$ref` keyword in  request bodies. Given the following oas endpoint:
```
/end-point-uri:
    post:
      tags:
        - SomeTag
      summary: Search for something with a request body
      description: |
       Search something with a request body that is defined as a reference
      operationId: PostSearchSomething
      parameters:
        - $ref: '#/components/parameters/CorrelationId'
        - $ref: '#/components/parameters/ReferenceDate'
      requestBody:
        $ref: '#/components/requestBodies/PostSearchSomethingRequest'
```

The generator would generate the following function `PostSearchSomethingRequestAsync(object body)` with the supplied change it will generate `PostSearchSomethingRequestAsync(PostSearchSomethingRequest body)`.

Fixes #4093 as well I think

PR #5196, opened 2025-07-01
Fix for Reusable request bodies are not handled correctly #2747

Doomic:
Generating a (csharp) client did not type the body parameter if the body contains a $ref to a reusable request body.
This pullrequest fixes the bug #2747 "Reusable request bodies are not handled correctly".

sidenote; the class name of the typed body will saidly be called "Body" (or with a number if there are multiple bodyies).. can't find a way to fix that without changing the NJsonSchema project. (out of scope for now).


Doomic:
i just saw that there is already a pull request #4747 .. it looks almost the same as mine.. not sure why that one is not approved.. but maybe my pullrequest is not needed.

PR #5198, opened 2025-07-02
Prioritize wildcard over json over first entry for accept content-type

mdi1984:
We have an issue with a external API definition that defines the following response:
```
      responses:
        '200':
          description: Status Code 200
          content:
            application/xml:
              schema:
                $ref: '#/components/schemas/QueryResult'
            application/json:
              schema:
                $ref: '#/components/schemas/QueryResult'
```

When generating a C# Client for this, the accept header will be set to "application/xml" while the client tries to deserialize the response as json. Since json is prioritized in many other spots this should be done here as well.


lahma:
Should there be some test case(s) that prove this is what is wanted and prevent from breaking in the future?

mdi1984:
@lahma sure thing. Thanks for noticing this. I have added some tests.

Issue #4045, opened 2022-06-10
[Typescript client] AbortSignal signal is not added as parameter to generated client interface


VolodymyrSenchak:
When I generate a typescript client with an interface and use the Fetch template with AbortSignal, the client class has a signal parameter, but the interface does not have such a parameter. As a result we cannot use correctly interface.
![image](https://user-images.githubusercontent.com/22198447/173053662-cf30e4b7-a44e-44c1-90fd-0d5f7082e450.png)

Issue #3634, opened 2021-09-16
openapi2tsclient does not add AbortSignal to interfaces


Strepto:
Hi!

When combining `/GenerateClientInterfaces:true`  and `/UseAbortSignal:true` the interface does not not contain the AbortSignal.

Example output snippet: 

```ts
export interface IApiStatusClient {
    getApiStatusIsOnline(): Promise<void>;
}

export class ApiStatusClient implements IApiStatusClient {
...
    getApiStatusIsOnline(signal?: AbortSignal | undefined): Promise<void> {
...
}
```

Note that Signal is defined in the class, but ignored in the interface.

Full command below:
```bash
nswag openapi2tsclient 
/output:src/services/generated/StatusApiClient.ts 
/input:https://URL/swagger/v1/swagger.json 
/ClassName:"{controller}Client"
 /NullValue:undefined 
/GenerateClientInterfaces:true 
/GenerateOptionalParameters:true 
/SupportsStrictNullChecks:true
/EnumStyle:StringLiteral 
/ServiceSchemes:https 
/OperationGenerationMode:MultipleClientsFromFirstTagAndOperationId 
/UseAbortSignal:true
/runtime:Net50
```

Version Info:
```text
NSwag NPM CLI
NSwag command line tool for .NET Core Net50, toolchain v13.13.2.0 (NJsonSchema v10.5.2.0 (Newtonsoft.Json v12.0.0.0))
```

Is this behaviour intentional? I might consider fixing this if its not by design.

PR #5202, opened 2025-07-09
Fix to expose an optional param for AbortSignal (for axios) / CancelToken (for fetch) when GenerateClientInterfaces = true

tomvdk:
Fix to expose an optional param for AbortSignal (for axios) / CancelToken (for fetch) on the client's interface similar to the concrete implementation so you can actually use it.
Fixes https://github.com/RicoSuter/NSwag/issues/3634 (#3634)
but also mentioned in this issue: https://github.com/RicoSuter/NSwag/issues/4045 (#4045) which implemented this only on the concrete implementation.


tomvdk:
Hi @lahma,

Can you please review and approve? Many thanks.

Kind regards,

Tom

PR #5205, opened 2025-07-10
Expose the UseRequiredKeyword on CSharpClientGeneratorSettings for CSharp client generator

r-englund:
A while ago NJsonSchema  (PR RicoSuter/NJsonSchema#1711) added the option `UseRequiredKeyword ` for the C# client generator. 
As far as I can tell there is currently no way to enable this keyword when generating clients using NSwag cli/msbuild tool. 

This PR exposes the option to the `OpenApiToCSharpClientCommand`, to allow that. 

When looking through the open issues, my interpretation is that there is two separate issues related to required properties on DTO's. 
 * C# 11 Keyword not begin added in generated code (addressed  in this PR)
 * The document generator only detecting properties as required if they have the Required attribute, not if they (only) have the C# 11 required keyword.  (not addressed in this PR) 

I added some unit-test for the `UseRequiredKeyword ` setting that adds tests for the combination of having required keyword or attribute (either, both or none) when the setting is on and when the setting is off. Testing using both `Newtonsoft` and the `SystemText` `JsonSchemaGenerator`. However, due to the second bullet point  above the test for required keyword is currently failing.  I feel like addressing that issue is out of scope for the PR. I did some debugging on it and guessing the code change is in either the `NJsonSchema` codebase or somewhere else.

### DIFF
--- a/src/NSwag.CodeGeneration.CSharp.Tests/RequiredParameterTests.cs
+++ b/src/NSwag.CodeGeneration.CSharp.Tests/RequiredParameterTests.cs
@@ -35,7 +35,7 @@ namespace NSwag.CodeGeneration.CSharp.Tests
             public string NotRequired { get; set; }
         }
 
-        private static async Task<string> GenereateCode(bool UseRequiredKeyword)
+        private static async Task<string> GenerateCode(bool UseRequiredKeyword)
         {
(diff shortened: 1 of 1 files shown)


lahma (approved):
LGTM, thank you!

Issue #3895, opened 2022-02-19
Client generated code does not compile


cabadam:
I am attempting to generate a client for the OpenAPI specification you can download from the top of this page: https://api.mattermost.com/

I have tried multiple different ways of generating:
1. Visual Studio 2022 adding a "Connected Service"
2. NSwagStudio
3. Using the NSwag.Msbuild

All more or less produce the same code and the same problem.  There is one easy to work-around issue where there is a property called in the json file called "<CHANNEL_ID>" and Nswag tries to generate a C# property of the same name.  Naturally that fails.  Easy enough to change the json to remove the <> characters.

However, after that I still get a variety of issues.  Some methods of generating will put this line just inside the namespace:
`using System = global::System;`

When compiling, that gives many of:

> Namespace 'MattermostTest' contains a definition conflicting with alias 'System'

If I comment out that line in the generated code, then the rest of the file throws compilation errors as if it has no idea what exists under "System." For example, the first class gets a GeneratedCode attribute (typename fully qualified):
`[System.CodeDom.Compiler.GeneratedCode("NSwag", "13.15.9.0 (NJsonSchema v10.6.8.0 (Newtonsoft.Json v13.0.0.0))")]`

That gives compilation error:

> error CS0426: The type name 'CodeDom' does not exist in the type 'System'

Dumping the json file into editor.swagger.io does give a few errors.  Stripping out the relevant pieces of the spec however yields no change in the errors.

I've tried it with multiple different versions of .NET (from .NET Framework 4.8, .NET Core 3.1, and .NET 6 which was my intended target). I also had a different person try generating the client code, no luck.

What are we missing?

Thanks

ovska:
I looked into it and there seem to be multiple things going wrong:

- As you noticed, the `<CHANNEL_ID>` parameter isn't sanitized. Not sure if this could be fixed in codegen side (not sure if it validates that identifiers are valid C# at all)
- The definition has a type called `System`, which causes the compilation problems you're encountering (note the "does not exist in the **type** 'System'"). There isn't a way to rename a generated type that I know of, so hard to say what to do about this one.
- The `UploadFile`-endpoint generates its' request body (multipart formdata?) parameters as normal request parameters, causing there to be two `string channel_id`s.

Manually fixing the client takes a bit of work, and isn't very robust.

cabadam:
[...]

PR #5208, opened 2025-07-16
Allow providing an alias for global System namespace

nycdotnet:
Hi,

Thanks for such a great library.

This PR somewhat addresses this issue: #3895 by adding a new parameter `GlobalSystemNamespaceAlias`.

An example of why you'd want to support this is to generate a client for the SpaceTraders.io game which has `System` as an entity (as in star system): https://github.com/SpaceTradersAPI/api-docs/blob/main/models/System.json

In this case, you could provide `/GlobalSystemNamespaceAlias:DotNetSystem` at the command line, for example, and it would bypass the problem.

This is not completely implemented - unfortunately it will require a change in NJsonSchema as well (for things like Enums which reference "System.Runtime.Serialization.EnumMember".).  I am curious of your thoughts on if you would take a change of this size or if you like this solution?  If so I can implement it there and update this PR.  If not, no worries.  We can probably also achieve this with some sort of rename logic.

Note: All tests pass with the current implementation.

Thank you!

### DIFF
--- a/src/NSwag.CodeGeneration.CSharp/CSharpClientGeneratorSettings.cs
+++ b/src/NSwag.CodeGeneration.CSharp/CSharpClientGeneratorSettings.cs
@@ -116,5 +116,16 @@ namespace NSwag.CodeGeneration.CSharp
 
         /// <summary>Gets or sets a value indicating whether to expose the JsonSerializerSettings property (default: false).</summary>
         public bool ExposeJsonSerializerSettings { get; set; }
+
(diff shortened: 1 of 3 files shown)


RicoSuter:
Wouldnt it be better to just preprocess the schema and rename the DTO name there to avoid conflicts (then you do not need to change NSwag/NJS at all)?

nycdotnet:
Yes you're right that would probably be better.

PR #5218, opened 2025-07-25
Convert NSwagStudio setup to use WiX 6

lahma:


lahma:
Tested installer and update works, generating code works, start menu has shortcut and icon that works. Uninstall and install works.

PR #5221, opened 2025-07-26
Replace C# string contain asserts with Verify tests

lahma:


r-englund:
@lahma 
Doesn't this PR remove a lot of actual tests logic? 
Many of the tests updated seems to have had checks that checked if certain code was actually genereated. 
For example: Take the test in `src/NSwag.CodeGeneration.CSharp.Tests/RequiredParameterTests.cs`
They verified that the properties in the genereated code correctly used the `required` keyword. Now they seem to only verify that the code compiles.

lahma:
> @lahma Doesn't this PR remove a lot of actual tests logic? Many of the tests updated seems to have had checks that checked if certain code was actually genereated. For example: Take the test in `src/NSwag.CodeGeneration.CSharp.Tests/RequiredParameterTests.cs` They verified that the properties in the genereated code correctly used the `required` keyword. Now they seem to only verify that the code compiles.

There are asserts like `await VerifyHelper.Verify(code);`, this checks that the full output is exactly as before - the snapshot files are stored in VCS. So the older asserts were weaker as they checked that "there's something somewhere in code" - now the full output is compared against known good state stored in snapshot files.

r-englund:
Ah, neat. Make sense now :) 
Sounds like a more robust solution

PR #5232, opened 2025-07-31
Update nswag.js

KaterSchnurz:
prevent passing through the /runtime arguments to the executable of nswag, because that causes errors in the actual tool.

completly restructured.

should now do what was expected from this script


lahma (approved):
Looks good, I think we can try this out and then revert or patch of complaints would rise.

PR #5235, opened 2025-08-01
Add credentials and mode attributes to fetch template generation

etriebe:
Haven't tested yet but just attempting to get this started.


lahma:
You can mark a PR as a draft so it won't be accidentally merged.

lahma:
Please add at least one test case that enables this option and does both `VerifyHelper.Verify()` and `TypeScriptCompiler.AssertCompiles()` there are examples around the TS test suite.

src/NSwag.CodeGeneration.TypeScript/Templates/FetchClient.liquid
```diff
@@ -51,6 +51,12 @@
 {%-    if UseAbortSignal -%}
             signal,
 {%-    endif -%}
+{%-       if WithCredentials -%}
+            credentials: 'include',
+{%-        endif -%}
+{%-       if UseCorsMode -%}
+            mode: 'cors',
```
lahma, line 58: because there are three possible values: https://developer.mozilla.org/en-US/docs/Web/API/RequestInit#mode should this be an enum on settings side? I don't know about other clients if they support same options, but then it would be `public enum CorsMode` or `FetchCorsMode` with default of `None`, client template model could have `Mode` property which would map enum to string version

etriebe: Yeah, I can change to be an enum that I assume would be nullable so people don't have to have it set?

etriebe: Any objection to me using the following enum?

https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.components.webassembly.http.browserrequestmode?view=aspnetcore-9.0

etriebe: Short of me just creating a boolean for each different enum option in NSwag.CodeGeneration.TypeScript\Models\TypeScriptClientTemplateModel.cs. Which feels hacky but would be relatively easy.

lahma: Just add a string property that returns correct value, null should evaluate to false in bif check

lahma:
Checked the PR and spec, some notes:

* `RequestCredentialsType` probably doesn't need NotSet, as spec says that: Defaults to `same-origin` 
* `RequestModeType` probably doesn't need NotSet, as spec says that: Defaults to `cors`

Now it remains a question whether it needs to be rendered in the JS file if it is the default - your current logic incorrectly returns empty string, like I said, null needs to be returned.

etriebe:
I was thinking about having no NotSet but then it'd change a bunch of people's client files unnecessarily. I would generally say I'd want it to not appear in people's file and therefore I'd need to do one of two things

1. Keep NotSet as an option for both and don't put anything in the file if it is there
2. Regardless of if they set it themselves or it stays the default, if either of these two options are present we don't add anything to the ts file. This doesn't feel great because if people have that set they probably want to see it show up in the file.

lahma:
> I was thinking about having no NotSet but then it'd change a bunch of people's client files unnecessarily. I would generally say I'd want it to not appear in people's file and therefore I'd need to do one of two things
> 
> 1. Keep NotSet as an option for both and don't put anything in the file if it is there
> 2. Regardless of if they set it themselves or it stays the default, if either of these two options are present we don't add anything to the ts file. This doesn't feel great because if people have that set they probably want to see it show up in the file.

I think the NotSet will be the value if you save the nswag.json using NSwagStudio for example, so the property will pesisted there, if you don't want it to appear then it probably needs JsonProperty attribute to ignore the default value.

lahma:
@etriebe do you have time/plans to fix the current test problems? 🙏🏻

etriebe:
Let me take a look. My development desktop died the other week so it's been a minute to get setup and configured again.

etriebe:
@lahma I am running locally and even just running for one of my tests I see the following and it spins for a very long time. Is there some sort of setup I need to do to make this work?

========== Starting test run ==========
Connecting to client host '127.0.0.1' port '55825'

lahma:
> @lahma I am running locally and even just running for one of my tests I see the following and it spins for a very long time. Is there some sort of setup I need to do to make this work?
> 
> ========== Starting test run ========== Connecting to client host '127.0.0.1' port '55825'

Haven't seen this before, might be related to your reset setup?

etriebe:
I think tsc is failing to install for me and/or it isn't on my path variable. How do you install tsc?

lahma:
Try running build.cmd in repository root.

etriebe:
When I run build.cmd it fails and says it can't find msbuild. And I have VS 2022 installed. 

```
PowerShell Desktop version 5.1.26100.4768
Microsoft (R) .NET SDK version 9.0.304
Unhandled exception. System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation.
 ---> System.InvalidOperationException: No instances of MSBuild could be detected.
Try calling RegisterInstance or RegisterMSBuildPath to manually register one.
   at Microsoft.Build.Locator.MSBuildLocator.RegisterDefaults()
   at Build..ctor() in D:\repos\NSwag\build\Build.cs:line 28
   at System.RuntimeType.CreateInstanceOfT()
   --- End of inner exception stack trace ---
   at System.RuntimeType.CreateInstanceOfT()
   at System.Activator.CreateInstance[T]()
   at Nuke.Common.Execution.BuildManager.Execute[T](Expression`1[] defaultTargetExpressions) in /_/source/Nuke.Build/Execution/BuildManager.cs:line 48
   at Nuke.Common.NukeBuild.Execute[T](Expression`1[] defaultTargetExpressions) in /_/source/Nuke.Build/NukeBuild.cs:line 78
   at Build.Main() in D:\repos\NSwag\build\Build.cs:line 44
```

etriebe:
I had previously just installed .NET 9 SDK but installing .NET 8 SDK allowed this to run, too.

etriebe:
OK tests are passing locally now. @lahma Can you approve to run them here?

lahma (approved):
Great addition, thanks for iterating! As a later improvement, we probably want these also in NSwagStudio and nswag.json.

PR #5245, opened 2025-08-10
Fix NPM CLI argument handling

### LINKED ISSUE #5238: not reading variables
**Describe the bug**

Nswag run command line not reading variables correctly anymore with version 14.5

**To Reproduce**

nswag run /runtime:Net80 /variables:Configuration=Release

now producing this error:
 NConsole.UnusedArgumentException: Unrecognised arguments are present: []
   at NConsole.CommandLineProcessor.ProcessSingleAsync(String[] args, Object input)
   at NConsole.CommandLineProcessor.ProcessAsync(String[] args, Object input)
   at NSwag.Commands.NSwagCommandProcessor.ProcessAsync(String[] args) in /_/src/NSwag.Commands/NSwagCommandProcessor.cs:line 61

lahma:
NConsole seems to have a bug which makes it a requirement that there needs to be one unused argument. This restores the old behavior and runtime parameter is passed to actual NET CLI.

fixes #5238

### DIFF
--- a/src/NSwag.Npm/bin/nswag.js
+++ b/src/NSwag.Npm/bin/nswag.js
@@ -36,8 +36,6 @@ args.forEach((arg, idx) => {
 if (runtimeIndices.length > 1) {
     console.error("Error: Multiple /runtime:* arguments detected. Please specify only one. Maybe remove the legacy --core argument?");
     process.exit(1);
-} else if (runtimeIndices.length === 1) {
-    args.splice(runtimeIndices[0], 1);
 }
 
 if (runtimeValue) {


RicoSuter:
Yeah getting rid of NConsole and using something better would be nice. 
But that would also be quite a breaking change.

Demivan:
@RicoSuter This causes https://github.com/RicoSuter/NSwag/issues/4678. It is preventing from selecting proper runtime for the run command
`nswag /runtime:Net100 run nswag-file.json` or `nswag run /runtime:Net100 nswag-file.json` does not work now.

PR #5248, opened 2025-08-15
Add option to generate code only for selected operation IDs

amayouf:
## 🌟 Why this feature is needed

This feature addresses a key pain point for users dealing with large OpenAPI/Swagger specifications. Previously, when generating C# or TypeScript clients, all operations within the specification were included by default. This led to unnecessarily large generated files and increased compile times, particularly when only a subset of the API was needed. This new feature provides a way to filter which operations are included in the generated code, resulting in leaner, more focused clients.

***

## 🛠️ How it works

This feature introduces two new configuration properties for code generation: **`includedOperationIds`** and **`excludedOperationIds`**.

* **`includedOperationIds`**: You can provide a list of specific operation IDs. When this list is provided, **only** the operations with matching IDs will be included in the generated client code.
* **`excludedOperationIds`**: You can provide a list of operation IDs to exclude. When this list is provided, all operations **except** those with matching IDs will be included.

It is important to note that you cannot use both `includedOperationIds` and `excludedOperationIds` simultaneously. The tool will throw an `InvalidOperationException` if both are configured, as this would create an ambiguous selection of operations.

These options are available in the following places:

* **Command-line interface (CLI):** You can pass these as arguments (e.g., `/includedOperationIds:GetPerson,CreatePerson`).
* **`.nswag` configuration file:** You can define a list of operation IDs in your `nswag.json` file.
* **MS Build properties:** The properties `NSwagIncludedOperationIds` and `NSwagExcludedOperationIds` have been added to the MSBuild targets to allow for configuration directly in project files.

***

## 📝 Tests and Documentation

* **Tests:** This patch includes new unit tests to ensure the feature works as expected. The tests cover both the `includedOperationIds` and `excludedOperationIds` scenarios and verify that only the correct operations are generated. There's also a test to ensure an exception is thrown if both properties are used at the same time.
* **Documentation:** The official documentation has been updated to include a section on how to use the new `includedOperationIds` and `excludedOperationIds` options in the `generate-proxy-client.md` file. A sample `nswag` file has also been updated to demonstrate its usage.


lahma (approved):
I think this is a neat feature, thank you for your contribution!

amayouf:
Thank you for the review and kind words!

RicoSuter:
Question: Does this also correctly scan all DTOs and exclude them as well based on usage?

The reason why this feature was not implemented yet is because you cannot just exclude operations but you'd also want to exclude DTOs which are not used anymore...

To avoid all this complexity I always said that you should just preprocess the swagger with another custom tool and then feed into the generator so we dont add more and more complexity to NSwag.

amayouf:
Hello Rico,

Unfortunately, it doesn't currently handle DTO exclusion. In the sandbox project I was using, I had `excludedTypeNames` configured, so most DTOs were being filtered out and I failed to notice this gap in my implementation.

I can implement a filter/processor to handle unused DTO exclusion when I have time this weekend.

amayouf:
Hello @RicoSuter, @lahma,

I’ve been exploring how to implement smart DTO filtering to complete the feature, but I ran into some difficulties.

Since `CodeArtifact.TypeName` from `NJsonSchema` only provides the short type name (without the namespace), it’s impossible to distinguish between cases like:
```
MyCompany.ProjectA.TypeName
MyCompany.ProjectB.TypeName
```

When digging deeper into `NJsonSchema`, I started wondering: should this kind of logic live in `NJsonSchema.CodeGeneration`, or is there another recommended place/approach for handling it?

Apologies if this is a naive question, I’m still new to these projects and trying to get the full picture.

Thanks a lot for your insights!

RicoSuter:
Yeah this is why I never added this feature, it's not simple and you should do it with own OpenAPI preprocessing (preprocess the openapi.json file before passing to NSwag)... this just adds unnecessary complexity to nswag and will not work correctly (very hard to correctly exclude only used DTOs). 

On the openapi.json generator side (aspnetcore openapi generator) you can exclude operations and there only used DTOs are included (works as expected).

lahma:
Maybe we should just unfortunately revert this PR due to these conceptual problems? My bad.

RicoSuter:
@lahma  is it even released yet?

lahma:
Nope, GitHub shows the release the PR was introduced in so this has not yet reached general public.

RicoSuter:
@lahma  what do you think? should we revert or keep this feature?

lahma:
Well I'm not sure. The main use case I saw here was easy configuration when you want minimal clients, the methods themselves create a lot of code - though the contract types remain.

RicoSuter:
Yeah let's keep it... but we should really not take more stuff in which is only half complete (DTOs are not correctly excluded) and adds more configs (NSwag has already too many configs, and it's already a testing hell) and can be solved externally... let's focus on features we have because in the end we are the ones who have to maintain and fix problems with all the features... the initial implementation is the smallest cost.

PR #5251, opened 2025-08-15
#5250 - K6 template for Typescript has errors when void is return type

### LINKED ISSUE #5250: K6 template for Typescript has error when void is return type
**Describe the bug**

K6 Template for typescript adds an extra `,` when a return type of void in the client.  

latest main (unreleased)

**To Reproduce**

npm run nswag with k6 as template type in nswag.json

**Expected behavior**

The generated typescript client won't have an extra comma in the properties of the function.

**Additional context**

The k6 template for typescript was introduced with #5239

jsclifford:
Fixes #5250 - Fixing errors by handling for void by having k6 http.Response as a return type so k6 checks can check status codes to identify success or failure.


lahma (approved):
Thank you for your contribution!

PR #5252, opened 2025-08-19
Restore Microsoft.Extensions.ApiDescription.Server reference

sindrekroknes:
The reference to Microsoft.Extensions.ApiDescription.Server was removed from NSwag.AspNetCore in #5217
Was that intentional, or should we add it back?


lahma:
What is not working?

RicoSuter:
In general, this feature will be removed in the next .NET version anyway.. so you should switch to nswag.json config file instead

Issue #5247, opened 2025-08-14
openApiToCSharpController: incorrect return type for files when useActionResultType=true


gaelj:
**Describe the bug**

When generating a controller stub that returns a file, and the generator is configured with `useActionResultType: true`, the return type should be `Task<FileResult>`, but instead we get a `Task<Actionresult<FileResult>>` which fails at runtime because FileResult is an implementation of ActionResult.
The runtime error is `Invalid type parameter 'Microsoft.AspNetCore.Mvc.FileResult' specified for 'ActionResult<T>'.`

Before NSwag 14.5.0, it was generated as an `IActionResult`, so there was no problem.

**Version of NSwag toolchain, computer and .NET runtime used**

`14.5.0`, dotnet 9

**To Reproduce**

Config:

```json
  "codeGenerators": {
    "openApiToCSharpController": {
      "controllerBaseClass": "Microsoft.AspNetCore.Mvc.Controller",
      "controllerStyle": "Abstract",
      "controllerTarget": "AspNetCore",
      "useCancellationToken": true,
      "useActionResultType": true,
      "generateModelValidationAttributes": false,
      "routeNamingStrategy": "OperationId",
      "basePath": "/",
      "className": "{controller}",
      "operationGenerationMode": "MultipleClientsFromFirstTagAndOperationId",
      "additionalNamespaceUsages": [],
      "additionalContractNamespaceUsages": [],
      "generateOptionalParameters": true,
      "generateJsonMethods": true,
      "enforceFlagEnums": false,
      "parameterArrayType": "System.Collections.Generic.List",
      "parameterDictionaryType": "System.Collections.Generic.IDictionary",
      "responseArrayType": "System.Collections.Generic.List",
      "responseDictionaryType": "System.Collections.Generic.IDictionary",
      "wrapResponses": false,
      "wrapResponseMethods": [],
      "generateResponseClasses": true,
      "responseClass": "SwaggerResponse",
      "namespace": "xxxx.Server.API",
      "requiredPropertiesMustBeDefined": true,
      "dateType": "System.DateOnly",
      "jsonConverters": null,
      "jsonLibrary": "SystemTextJson",
      "anyType": "string",
      "dateTimeType": "System.DateTime",
      "timeType": "System.TimeSpan",
      "timeSpanType": "System.TimeSpan",
      "arrayType": "System.Collections.Generic.List",
      "arrayInstanceType": "System.Collections.Generic.List",
[...]

PR #5255, opened 2025-08-25
Fix binary operation result type generation for AspNetCore CSharp

xamele0n:
Fix generation ActionResult<FileResult> insead of IActionResult
#5247


lahma (approved):
Awesome, thank you!

Issue #5216, opened 2025-07-24
Support for .NET 10?


bkoelman:
Many OpenAPI breaking changes are being announced for .NET 10, most notably the [removal of OpenApiReference](https://github.com/aspnet/Announcements/issues/518).

Is the NSwag team aware of these and preparing to update NSwag accordingly? Is there a branch or dev feed with packages/binaries that are compatible with the latest preview of .NET 10, along with guidance? We'd like to proactively test and plan support for .NET 10 using the NSwag C# client generator.

lahma:
See https://github.com/RicoSuter/NSwag/discussions/5169 , pull requests are welcome.

bkoelman:
I don't think this is a matter of patching up the sources, which worked for earlier .NET updates. See also https://github.com/dotnet/aspnetcore/issues/59208.

Fundamental building blocks in ASP.NET that exist to support NSwag are being deprecated. It's currently unclear when the related parts in Visual Studio (Add Service Reference) will be removed.

How can we use NSwag if `ServiceReference` no longer exists? Please provide guidance on alternatives. I don't have an answer. This may require a major redesign of NSwag, I don't know.

Swashbuckle is making heavy changes for .NET 10 support.
The complete silence here is deeply concerning.

lahma:
I think you are talking about the tooling support for specific use case instead of general NET 10 support. The CLI tooling should probably work as before and so the common approach described here: https://github.com/RicoSuter/NSwag/wiki/NSwag.MSBuild

bkoelman:
I'm not familiar with all the details, but I suspect that https://stevetalkscode.co.uk/openapireference-commands (which describes how to use https://github.com/RicoSuter/NSwag/wiki/ServiceProjectReference) won't be possible anymore. But there's more: certain endpoint metadata, intended for Swashbuckle and NSwag usage, will no longer be populated.

My point is that things are changing, more than usual, and a proactive investigation by the NSwag maintainers could influence key decisions and ensure NSwag can still be used after .NET 10 has shipped. Please assess .NET 10 compatibility now, while there's still an opportunity to influence ASP.NET 10.

Releasing preview versions of the NSwag packages/binaries that can be used with .NET 10 Preview is only the first step. Community members can then try them on their codebases and report any issues that arise. This is what's currently happening with Swashbuckle.

lahma:
[...]

PR #5271, opened 2025-09-16
Add .NET 10 support

RicoSuter:
Ref #5216


RicoSuter:
@lahma it seems that we need to install .NET8/9 as well, right? But this script only installs/ensures the one from global.json. What's the right way to fix this?

lahma:
@RicoSuter I think you need to adjust here: 

https://github.com/RicoSuter/NSwag/blob/6db16d724bc8bac77d18437f5755923eb25b7030/build/Build.CI.GitHubActions.cs#L49-L53

Maybe shouldn't have OS check then. Then run build.cmd to regenerate.

metoule:
It looks like the https://github.com/RicoSuter/NSwag/blob/master/src/NSwag.ApiDescription.Client/NSwag.ApiDescription.Client.targets haven't been updated for net10.0. Will that cause issues?

PR #5283, opened 2025-10-07
fix axios template: [object Object] is not valid JSON.

### LINKED ISSUE #4973: BUG: "[object Object]" is not valid JSON - the processor try to run json.parse on axios object
On version v14.1.0. When choose generate typescript DTO interfaces with axios. The response processor will throw exception. I'm using the debugger, found that the generated client try to use JSON.parse on a Object type.
image
image
image

adnanalbeda:
Fixes #4973

Added condition to check if template is `Framework.IsAxios` then return data without parse, cause axios convert json string to data object by default.


RicoSuter:
can you maybe update with latest master, hopefully CI goes green

adnanalbeda:
@RicoSuter  Done. Now CI is waiting for approval.

pgadmin7:
`Added condition to check if template is Framework.IsAxios then return data without parse, cause axios convert json string to data object by default`

I bumped from v14.2 to v14.7.2. Also use Axios but it doesn't magically parse string to object.
Am I missing a setting in the config?

PR #5299, opened 2025-11-10
Updated packages for CVE-2025-55315

erjdev:
See https://github.com/dotnet/announcements/issues/371 for more information

I believe this will handle the security vulnerability being reported transitively for this package; however, if the reported vulnerability is not resolved, I could see us also needing to potentially add a direct reference to https://www.nuget.org/packages/Microsoft.AspNetCore.Server.Kestrel.Core version 2.3.6 to explicitly override its version.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -1,7 +1,7 @@
 <Project>
   <PropertyGroup>
     <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
-    <CentralPackageTransitivePinningEnabled>false</CentralPackageTransitivePinningEnabled>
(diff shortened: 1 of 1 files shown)


erjdev:
According to the build logs in CI, it appears that this issue is not yet resolved and will need an explicit override for net462, which means transitive dependency overrides will need to be enabled.

erjdev:
This should now be ready for review with the last commit.  My local no longer complains with the transitive dependency override present.

bpouillaude-genetec:
@RicoSuter, any chance to get your input for this? 
Would be appreciated :)

erjdev:
Any chance we can get movement on this?  This .NET security issue is affecting all versions of nswag that I attempt to utilize at work.  It's preventing local and CI usage of the tool for generating client/server classes because of our vuln scanner.

Issue #4010, opened 2022-05-17
Query Params' Default Values missing in C# Method


Germandf:
Hi! My team and I are using NSwag.ApiDescription.Client to generate OpenAPI specification in our Api and autogenerate DTOs and methods necessary to use the Api from another project.

We have two optional query params in every endpoint, Env and ApiKey. They both have a default value:

```csharp
services.AddSwaggerGen(c =>
{
    c.DocumentFilter<AddDefaultsQueriesParameters>();
});
```

```csharp
internal class AddDefaultsQueriesParameters : IDocumentFilter
{
    public void Apply(OpenApiDocument swaggerDoc, DocumentFilterContext context)
    {
        foreach(var p in swaggerDoc.Paths)
        {
            foreach(var (t,o) in p.Value.Operations)
            {
                o.Parameters.Add(new OpenApiParameter()
                {
                    Name = "Env",
                    In = ParameterLocation.Query,
                    Description = "Environment",
                    Required = false,
                    Schema = new OpenApiSchema()
                    {
                        Type = "string",
                        Default = new OpenApiString("Test")
                    }
                });
                o.Parameters.Add(new OpenApiParameter()
                {
                    Name = "ApiKey",
                    In = ParameterLocation.Query,
                    Description = "ApiKey",
                    Required = false,
                    Schema = new OpenApiSchema()
                    {
                        Type = "string",
                        Default = new OpenApiString("")
                    }
                });
            }
        }
    }
}
```

For this example, let's take the NewJobObservation method:

``` csharp
[HttpPost("NewJobObservation", Name = "NewJobObservation")]
public async Task<NewJobObservationResponse> NewJobObservation([FromBody] NewJobObservationRequest request)
{
    var result = await _mediator.Send(request);
    return result;
}
```

![capture](https://user-images.githubusercontent.com/69018178/168876689-facff4ce-7f69-43f2-a981-0473a87c8a13.png)

Everything is fine so far. You can use the endpoint without writing Env neither ApiKey. The problem is in another project that connects to it via a service reference. When we refresh it by downloading the latest api definition, the autogenerated methods don't have the default value. They look like this:

```csharp
/// <param name="env">Environment</param>
/// <param name="apiKey">ApiKey</param>
/// <returns>Success</returns>
/// <exception cref="ApiException">A server side error occurred.</exception>
[...]

Issue #3697, opened 2021-11-08
С# client generation is missing default values for parameters


IhnatKlimchuk:
Api description in bi-swagger.json:
```
{
  "openapi": "3.0.1",
  "paths": {
    "/api/integration/bi/reports/profiles": {
      "get": {
        "tags": [
          "Reports"
        ],
        "operationId": "GetProfileReportAsync",
        "parameters": [
          {
            "name": "from",
            "in": "query",
            "schema": {
              "type": "string",
              "format": "date-time"
            }
          },
          {
            "name": "to",
            "in": "query",
            "schema": {
              "type": "string",
              "format": "date-time"
            }
          },
          {
            "name": "skip",
            "in": "query",
            "schema": {
              "type": "integer",
              "format": "int32"
            }
          },
          {
            "name": "api-version",
            "in": "header",
            "description": "The requested API version",
            "required": true,
            "schema": {
              "type": "string",
              "default": "2.0"
            }
          }
        ],
        "responses": {
          "200": {
            "description": "Success",
            "content": {
              "text/plain": {
                "schema": {
                  "type": "array",
                  "items": {
                    "$ref": "#/components/schemas/ProfileReportItem"
                  }
                }
              },
              "application/json": {
                "schema": {
                  "type": "array",
                  "items": {
                    "$ref": "#/components/schemas/ProfileReportItem"
                  }
                }
              },
              "text/json": {
                "schema": {
                  "type": "array",
                  "items": {
                    "$ref": "#/components/schemas/ProfileReportItem"
                  }
                }
              }
            }
          }
        }
      }
    }
  },
  "components": {
    "schemas": {
      "ProfileReportItem": {
        "type": "object",
        "properties": {
          "id": {
            "type": "string",
            "format": "uuid"
          }
        },
        "additionalProperties": false
      }
    }
  }
}
```
Nswag file:
```
{
  "runtime": "Net50",
  "defaultVariables": null,
  "documentGenerator": {
    "fromDocument": {
      "url": "./bi-swagger.json",
      "output": null,
      "newLineBehavior": "Auto"
    }
  },
  "codeGenerators": {
    "openApiToCSharpClient": {
      "clientBaseClass": null,
      "configurationClass": null,
      "generateClientClasses": true,
      "generateClientInterfaces": true,
      "clientBaseInterface": null,
      "injectHttpClient": true,
      "disposeHttpClient": true,
      "protectedMethods": [],
      "generateExceptionClasses": true,
[...]

PR #5311, opened 2025-11-29
fix: С# client generation is missing default values for parameters

renatosc:
This is a fix for #3697 and #4010

image

### DIFF
--- a/src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.liquid
+++ b/src/NSwag.CodeGeneration.CSharp/Templates/Client.Class.liquid
@@ -131,7 +131,7 @@
 {%     if GenerateOptionalParameters == false -%}
     {% template Client.Method.Documentation %}
     {% template Client.Method.Annotations %}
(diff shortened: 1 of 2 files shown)


renatosc:
@RicoSuter @lahma when you have time. And Thank you for the work on this repo.

renatosc:
I am checking the tests

renatosc:
I think I may need so guidance with those tests. 

For example, I look at that windows-latest fail. It says that "  NSwag.CodeGeneration.CSharp.Tests.ClientGenerationTests.**CanGenerateFromJiraOpenApiSpecification** [FAIL]" because the **JIRA_OpenAPI.received.txt** is not the same as JIRA_**OpenAPI.verified.txt**

My understanding is that the test **CanGenerateFromJiraOpenApiSpecification**  is using the **jira-open-api.json** as the openapi schema source.

I got the contents of that file and I generated the csharp client using NSwagStudio compiled from main branch and also from my PR branch. Both generated clients have the same contents which were equal to the **JIRA_OpenAPI.verified.txt** file.

I also ran the **build Compile Test Pack** locally and all tests are passing.

Any idea on where I should look at?

PR #5312, opened 2025-12-01
Fix "Controller" suffix removal

marius-hofmann:
* Only remove last occurrence of "Controller" when processing controller name in `WebApiOpenApiDocumentGenerator`
* To match implementation of aspnetcore: https://github.com/dotnet/aspnetcore/blob/main/src/Mvc/Mvc.Core/src/ApplicationModels/DefaultApplicationModelProvider.cs#L164-L167

### DIFF
--- a/src/NSwag.Generation.WebApi/WebApiOpenApiDocumentGenerator.cs
+++ b/src/NSwag.Generation.WebApi/WebApiOpenApiDocumentGenerator.cs
@@ -331,10 +331,7 @@ namespace NSwag.Generation.WebApi
             }
             else
             {
-                if (controllerName.EndsWith("Controller", StringComparison.Ordinal))
-                {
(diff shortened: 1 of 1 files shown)


marius-hofmann:
@lahma thanks for having a look at my PR. I'm sorry, that the tests failed, but I think, the failing is not caused by my changes. It seems like http://petstore.swagger.io/v2/swagger.json returned a 502 during the test execution. I don't know if this is a permanent issue or just a short service outage. Could you re-run the checks in order to verify if there is a more permanent problem?

RicoSuter:
@marius-hofmann should we do the same for the asp.net core implementation as well?

marius-hofmann:
> @marius-hofmann should we do the same for the asp.net core implementation as well?

@RicoSuter What exactly do you mean? I could not find another position which uses this logic.

PR #5316, opened 2025-12-25
Option to exclude deprecated endpoints

knoepdan:
I added a new option "excludeDeprecated" to exclude deprecated endpoints from client generation. 

My motivation for this change comes from the experience I have within my own company where we have a huge api that is used internally by many different applications but also exposed to the outside world. We have a policy that we never consume deprecated endpoints ourselves. However, sometimes this policy is hard to follow (sometimes depending on the repo and it's settings). Excluding every deprecated endpoint would help us a lot as we would get compile time errors, once "excludeDeprecated" is set to true. Furthermore, there it size of the clients would be removed, especially for Typescript and some complexity for the developer is removed. 

PS: A lot of the changes were similar as a checkin from 4 months ago that added the settings "includedOperationIds" and "excludedOperationIds". Helped me a lot and I think that functionality will also help us. Thanks Amine. But I still think this would be helpful


knoepdan:
Is there a chance this PR gets reviewed and merged?

knoepdan:
Thanks CoPilot.. great job ;-)   will look into it

knoepdan:
Updated the PR. Addressed all the co-pilot feedback. 

Hope it is ok as is :-)

PS: if this one is merged, I might try to add regular expression support for IncludedOperationIds/ExcludedOperationIds  (unless I find a good reason not to do it)

knoepdan:
Anyone? Does this PR have a chance getting merged?

PR #5345, opened 2026-03-01
Fix compile error in generated C# client when media type contains quotes

### LINKED ISSUE #5344: Compile error if double quotes in media type
**Describe the bug**

Using the C# client generator on an OpenAPI document that contains a media type whose value has double quotes fails to compile.

The [JSON:API specification](https://jsonapi.org/format/) requires extensions and profiles to be surrounded by double quotes, for example, [here](https://jsonapi.org/ext/atomic/#specific-operations):
```
POST /operations HTTP/1.1
Host: example.org
Content-Type: application/vnd.api+json;ext="https://jsonapi.org/ext/atomic"
Accept: application/vnd.api+json;ext="https://jsonapi.org/ext/atomic"
```

When multiple extensions are used, the spec [requires](https://github.com/json-api/json-api/issues/1768) them to be present in a single "ext" parameter, separated by whitespace. Thus, double quotes are needed to preserve the value. For example:
```
Content-Type: application/vnd.api+json; ext="my-ext1 my-ext2 my-ext3"
```

v14.6.3 on .NET 10

**To Reproduce**

Generate a C# client from the following openapi.json document:
```json
{
  "openapi": "3.0.1",
  "paths": {
    "/Test": {
      "post": {
        "requestBody": {
          "content": {
            "application/vnd.api+json; ext=\"https://jsonapi.org/ext/atomic https://www.jsonapi.net/ext/openapi\"": {
              "schema": {
                "type": "object"
              }
            }
          }
        },
        "responses": {
          "200": {
            "description": "OK",
            "content": {
              "application/vnd.api+json; ext=\"https://jsonapi.org/ext/atomic https://www.jsonapi.net/ext/openapi\"": {
                "schema": {
                  "type": "object"
                }
              }
            }
          }
        }
      }
    }
  }
}
```

**Expected behavior**

Generated C# client compiles and works.

**Actual behavior**

The generated client fails to compile because double quotes in the media type are not escaped.

Image

bkoelman:
Fixes #5344.

### DIFF
--- a/src/NSwag.CodeGeneration.CSharp.Tests/ControllerGenerationFormatTests.cs
+++ b/src/NSwag.CodeGeneration.CSharp.Tests/ControllerGenerationFormatTests.cs
@@ -427,10 +427,10 @@ namespace NSwag.CodeGeneration.CSharp.Tests
             var code = codeGen.GenerateFile();
 
             // Assert
-            var fromHeaderCustomAttributeCount = Regex.Matches(code, "public class FromHeaderAttribute :").Count;
+            var fromHeaderCustomAttributeCount = Regex.Count(code, "public class FromHeaderAttribute :");
             Assert.Equal(1, fromHeaderCustomAttributeCount);
 
-            var fromHeaderCustomBindingCount = Regex.Matches(code, "public class FromHeaderBinding :").Count;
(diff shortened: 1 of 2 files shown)


bkoelman:
Thanks for the review.

PR #5352, opened 2026-03-15
fix(binary): Add IFormFile support for binary in multipart requests

basizov:
- Added IFormFile generation for using binary in multipart/form-data
- Added test for this case


basizov:
@RicoSuter could't you see please?

Issue #1692, opened 2018-10-28
Make blobToText replaceable in Typescript template


ruskom:
I found difficult to replace the functionality implemented in blobToText function.
Because File API (FileReader) is not present in server side code (Angular SSR -> Universal), the calls are always failing with error message "ReferenceError: FileReader is not defined".
Not easy to avoid this issue, because the calls and the function itself hard coded.

It would be great if it can be more flexible somehow or avoid using DOM related functions like File API.

RSuter:
You can replace this particular template (File.Utilities.liquid): https://github.com/RSuter/NSwag/wiki/Templates

What wouldbe a good always working replacement?

ruskom:
Thank you. I'll give it a try.

Chreekar:
We solved this by manually editing the generated code to replace `responseType: "blob"` with `responseType: "json"` in the options object. Then we could just cast the response body to our dto model and delete the blobToText function. 

It would have been great to have a select in the NSwag UI to use json instead of blob in HttpClient like this, would that be possible in the future?

RicoSuter:
So ideally it would use "json" for JSON responses and "blob" for file/binary responses, correct?

Chreekar:
Exactly :-)

bluee:
@Chreekar thanks for suggesting the responseType change. 

For now I did Creekar's change with responseType: json but to prevent changing everywhere in the swagger file I kept blobToText with a minor change. It's not optimal but it works with Angular SSR.

Would be nice with a permanent fix though.


```
function blobToText(blob: any): Observable<string> {
    return new Observable<string>((observer: any) => {
        observer.next(JSON.stringify(blob));
        observer.complete();
    });
}
```

ashrafsa:
@Chreekar 
I think that its possible to work with providers for customization instead of manually change the auto generated output.
for instance:

- blobToTextProvider which is injectable and the developer can assign the ts class in the UI.

- httpRequestOptionsProvider

Chreekar:
I haven't understood the concept of providers, is this something I just change in the UI or do I need to create them in code somehow?

Chreekar:
[...]

PR #5363, opened 2026-04-13
fix: replace FileReader with Blob.text() in Angular blobToText for SSR compatibility

### LINKED ISSUE #3177: blobToText in NativeScript Angular -> FileReader event is undefined
When using the generated file in NativeScript Angular, in this part of code, in the reader.onload event the passed event seems to be undefined.

`function blobToText(blob: any): Observable<string> {
    return new Observable<string>((observer: any) => {
        if (!blob) {
            observer.next("");
            observer.complete();
        } else {
            let reader = new FileReader();
            reader.onload = event => {
                observer.next((<any>event.target).result);
                observer.complete();
            };
            reader.readAsText(blob);
        }
    });
}
`

changing this line 
`                observer.next((<any>event.target).result);
`
to this
`
 observer.next(reader.result);`
every time after the file is TypeScript client is generated seems to fix this.

PeterTheOne:
## Problem

The generated Angular `blobToText` function uses `FileReader`, which is a browser-only DOM API. This causes crashes in any non-browser Angular environment:

- **Angular SSR / Universal** — `ReferenceError: FileReader is not defined` on every API call, since NSwag sets `responseType: 'blob'` for all requests and every response (including error bodies) is routed through `blobToText`
- **NativeScript Angular** — `FileReader.onload` event is undefined in NativeScript's runtime (see #3177)
- **Jest / jsdom test environments** — depending on jsdom version and configuration, `FileReader` may not be available or may behave differently
- **Electron main process** — Node.js context, no DOM APIs

Because `blobToText` sits in the hot path of every generated `processXxx` method (both success and error branches), there is no easy way to work around this without patching the generated output or adding layered interceptors that fight each other.

## Fix

Replace `FileReader` with `Blob.text()`, which is:
- Available in **all modern browsers** (Chrome 76+, Firefox 69+, Safari 14.1+, Edge 79+)
- Available in **Node.js 16+** as a stable global

The change is in `File.Utilities.liquid`, Angular block only — the AngularJS `blobToText` is unchanged:

```diff
-            let reader = new FileReader();
-            reader.onload = event => {
-                observer.next((event.target as any).result);
-                observer.complete();
-            };
-            reader.readAsText(blob);
+            (blob as Blob).text().then(
+                text => { observer.next(text); observer.complete(); },
+                err => observer.error(err)
+            );
```

All 8 affected Angular snapshot files have been updated accordingly.

## Verification

I have applied this fix locally in an Angular SSR project (Node 18) and the crash is resolved. However I want to be upfront: **real-world testing has been limited** — I haven't verified all edge cases like large blobs, binary responses, or error body handling across different Angular versions. Additional review and testing from maintainers or other users would be very welcome before merging.

## Related issues

Fixes #3177
Related: #1692

---

*Drafted with the help of [Claude Code](https://claude.ai/code). The root cause analysis, template change, and snapshot updates were done with AI assistance — flagging this for transparency.*

### DIFF
--- a/src/NSwag.CodeGeneration.TypeScript.Tests/Snapshots/AngularTests.When_consumes_is_url_encoded_then_construct_url_encoded_request.verified.txt
+++ b/src/NSwag.CodeGeneration.TypeScript.Tests/Snapshots/AngularTests.When_consumes_is_url_encoded_then_construct_url_encoded_request.verified.txt
@@ -146,12 +146,10 @@ function blobToText(blob: any): Observable<string> {
             observer.next("");
             observer.complete();
         } else {
-            let reader = new FileReader();
-            reader.onload = event => {
-                observer.next((event.target as any).result);
-                observer.complete();
-            };
-            reader.readAsText(blob);
+            blob.text().then(
+                text => { observer.next(text); observer.complete(); },
+                err => observer.error(err)
+            );
         }
     });
 }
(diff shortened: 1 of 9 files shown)


mcq8:
This also fixes the error `Can't find variable: FileReader` for a browser on iOS with lockdown mode enabled as it does not have FileReader.

Issue #5368, opened 2026-04-21
[Axios] [Blob] Vue Typescript compiler show errors relevant to TS(2322) "Type AxiosHeaderValue | undefined is not assignable to type string | undefined"


sangnt-developer:
**Describe the bug**

I have an API endpoint which returns **Microsoft.AspNetCore.Mvc.FileContentResult response** to client on ASP.NET 8+ back end side.
<img width="1130" height="212" alt="Image" src="https://github.com/user-attachments/assets/5dd64e55-17c0-4332-b358-66b4233bb17f" />

I installed and run nswag command to generate an API service on Vue 3+ project with typescript and axios (version ^1.9.0). But after I had updated API service class to map with latest version on API by `nswag run` CLI command, then I tried to run build  and got errors on API service as following screenshot.
<img width="1515" height="485" alt="Image" src="https://github.com/user-attachments/assets/8fffcb97-eab9-4db4-8fd7-3d2b07c564e4" />

<img width="1538" height="478" alt="Image" src="https://github.com/user-attachments/assets/f3fb8fcc-52d8-4992-abfa-5cb8b8d39acd" />

**Version of NSwag toolchain, computer and .NET runtime used**
<img width="1052" height="321" alt="Image" src="https://github.com/user-attachments/assets/b539778e-a57f-4397-b87b-64237b084537" />

- Nswag NPM CLI
- Runtime: Net80
- Toolchain: v13.20.0 | v14.7.1

**To Reproduce**

<img width="328" height="361" alt="Image" src="https://github.com/user-attachments/assets/836625c7-a6ec-4ee7-bee1-80e92c366485" />

  1. Init **Vue 3+** project
  2. Add typescript support and compiler packages.
    - **Typescript**: 5.8.3
    - **Vue Typescript Compiler (vue-tsc)**: 3.2.7 
  3. Add Nswag package
    - **Nswag**: 14.7.1
  4. Add Nswag configuration file (_api-public.nswag_ file)
```json
{
"codeGenerators": {
    "openApiToTypeScriptClient": {
      "className": "{controller}ApiClientService",
      "typeScriptVersion": 3.9,
      "template": "Axios",
      "promiseType": "Promise",
      "httpClass": "HttpClient",
      "useSingletonProvider": true,
      "rxJsVersion": 6.5,
      "dateTimeType": "Date",
      "nullValue": "null",
      "generateClientClasses": true,
      "generateClientInterfaces": false,
      "generateOptionalParameters": true,
[...]

PR #5373, opened 2026-04-28
Fix Axios typing issue after axios upgrade to 1.15.2 by changing the liquid template to include ?.toString() to convert it to a string as the type

tomvdk:
chore(axios): Fix Axios typing after axios upgrade to 1.15.2 by changing the liquid template to include ?.toString() to convert it to a string as the type . See #5368 - [Axios] [Blob] Vue Typescript compiler show errors relevant to TS(2322) "Type AxiosHeaderValue | undefined is not assignable to type string | undefined"

This PR fixed the issue for me. Any remarks welcome.


tomvdk:
Can someone approve this PR please? @RicoSuter @lahma 
Thank you.

PR #5388, opened 2026-06-15
Update File.liquid

sasjaq:
In c# code generation, AdditionalNamespaceUsages option is not propagated

### DIFF
--- a/src/NSwag.CodeGeneration.CSharp/Templates/File.liquid
+++ b/src/NSwag.CodeGeneration.CSharp/Templates/File.liquid
@@ -11,7 +11,9 @@
 {% for usage in NamespaceUsages -%}
 using {{ usage }};
 {% endfor -%}
-
+{% for usage in AdditionalNamespaceUsages -%}
+using {{ usage }};
+{% endfor -%}
 {% template File.Header %}
 
 namespace {{ Namespace }}


RicoSuter:
`AdditionalNamespaceUsages` exists on `CSharpGeneratorBaseSettings`, but it is not a property of `CSharpFileTemplateModel`, which is passed to this template. The model already exposes that setting through [`NamespaceUsages`](https://github.com/RicoSuter/NSwag/blob/c953f29fadbf7d1d33d66d40b545859203702074/src/NSwag.CodeGeneration.CSharp/Models/CSharpFileTemplateModel.cs#L54-L57): full/implementation output uses `AdditionalNamespaceUsages`, while contracts output uses `AdditionalContractNamespaceUsages`.

I verified the behavior on this PR's commit (`c953f29`): with `AdditionalNamespaceUsages = ["System.Linq"]`, the existing `NamespaceUsages` loop emits the import, while the added loop resolves to an empty collection and emits nothing. Comparing the original and proposed templates across 12 client/controller generation scenarios produced equivalent output apart from whitespace. The C# test suite also passed (129 passed, 2 skipped).

I found no breaking change, but this patch does not fix the reported issue. Could you share a minimal failing configuration or reproduction, ideally with a regression test? If the missing imports are in a separate contracts file, please check whether `AdditionalContractNamespaceUsages` addresses that case.

PR #5396, opened 2026-08-10
Use undefined as return when it's configured.

atmoos:
todo


atmoos:
Sorry, this ended up here prematurely. Closing the PR.

PR #5398, opened 2026-08-13
Enable typescript undefined response

atmoos:
## Rationale

Returning no value as a valid response to an endpoint (Controller action) is a viable pattern that is apparently currently only supported via `null` by the NSwag typescript code generator.

The apparent configuration value `nullValue = Undefined` is unfortunately not propagated to return types, meaning they are only ever `TReturn | null` instead of `TReturn | undefined`.

## Proposal Outline

This PR bridges that gap in the [TypeScriptOperationModel](https://github.com/RicoSuter/NSwag/pull/5398/changes/f3690c23840ab59d4a185f7bf439835039a68459), by propagating the _already existing_ setting to return types, enabling the following:

| C# return type 1) | settings `nullValue` | TS return type | 
|--|--|--|
| `TReturn` | "doesn't apply" | `TReturn` |
| `TReturn?` | "default" | `TReturn \| undefined` 2) |
| `TReturn?` | `Undefined` |  `TReturn \| undefined` |
| `TReturn?` | `Null` |  `TReturn \| null` |

This is also mirrored in the [tests](https://github.com/atmoos/NSwag/blob/enableTypescriptUndefinedResponse/src/NSwag.CodeGeneration.TypeScript.Tests/TypeScriptOperationReturnTests.cs) I added.

Also the above table applies not just to C#'s nullability feature but also when using the nullability attributes. Naturally, this actually relies on the OpenApi definition `"x-nullable": "true/false"`. 

### Notes

1. The return type irrespective of how it's defined in C# (`TReturn`, `Task<TReturn>`, `{I}ActionResult<TReturn>`). The OpenApi definition is relevant.

2. The fact that the default value for `nullValue` is `Undefined` is arguably the right choice from a TS point of view. In the context of this change it is possibly somewhat unfortunate for backward compatibility. But see the discussion about alternatives further below.

## Alternatives

This PR represents the (arguably) simplest possible way to enable undefined returns. It may not be the optimal solution in terms of backward compatibility (point 2)). Hence, here some Ideas to change the implementation, if so desired by the maintainer.

### A) Additional Setting

Add another independent setting: say a `nullReturnValue` which would take exactly the same values as `nullValue`, where the default would be set to `Null` for backward compatibility.

### B) Changing the default value of `nullValue` to `Null`

This would keep return types the same, but might wreak havoc on input types and Dto parameters. If this PR were merged with a changed default value, any project that does not set that value would suddendly be seeing `TInputParameter | null` instead of `TInputParameter | undefined`. (Although by setting the `nullValue` setting this would be easily fixable).

## Tests

It may seem that there are too many changed tests. Here's why they changed

- As far as I can tell the projects don't yet make use of C# built in nullability feature meaning a return type of `TReturn` (some class) would be interpreted as nullable (consistent with C#'s default behaviour prior to the nullability feature)
- The settings in the OpenAPI base generator used in the tests interprets a return type without attribute annotation as "CanBeNull"
  - If `_settings.DefaultResponseReferenceTypeNullHandling` is added to the configuration of the `typeDescriptor` on line 283 in [OperationResponseProcessorBase.cs](https://github.com/atmoos/NSwag/blob/1d4820d8c95b328bb22a60cbeaf01af53064ea43/src/NSwag.Generation/Processors/OperationResponseProcessorBase.cs#L283), the interpretation would change to "NotNull". This would lead to changes in the test verifications form `TReturn | null` to `TReturn` only


src/update-ts-snapshots.sh
atmoos, line 1: Oh, I meant to delete this. I'll definitely do this before anything is merged.

src/NSwag.CodeGeneration.TypeScript/Models/TypeScriptOperationModel.cs
```diff
@@ -84,9 +85,10 @@ public override string ResultType
             {
                 var response = GetSuccessResponse();
                 var isNullable = response.Value?.IsNullable(_settings.CodeGeneratorSettings.SchemaType) == true;
+                var optionalReturn = _settings.TypeScriptGeneratorSettings.NullValue == TypeScriptNullValue.Undefined ? "undefined" : "null";
 
-                var resultType = isNullable && UnwrappedResultType != "void" && UnwrappedResultType != "null" ?
-                    UnwrappedResultType + " | null" :
+                var resultType = isNullable && UnwrappedResultType != "void" ?
```
RicoSuter, line 90: shouldnt it be

```suggestion
                var resultType = isNullable && UnwrappedResultType != "void" && UnwrappedResultType != optionalReturn 
```

?

atmoos: Quite right. I'll add a test for that on Monday and fix it.

atmoos: Ok,so I added two test that assert the returned Promise as being `Promise<any>` when the return type is a nullable any. So we don't get a redundant `Promise<any | null>`, as `any` already "includes" `null` (as well as `undefined`).

Was that the intent of your original comment? If so, I had to change your suggestion to not check against `UnwrappedResultType != optionalReturn` but to verify that `UnwrappedResultType is not any`.

Does that capture your intention?

---

Update: The original `UnwrappedResultType != "null"` guarded a generated result type of `null`, which likely can't happen. The same holds for `undefined` as a bare result type. But the `any` case is precisely what is generated by a controller action that returns `object`. So I'd argue this is what we should be guarding against. Do you agree?

RicoSuter:
The generated return type and runtime value currently disagree. For example, the new nullable-string snapshot declares `Promise<string | undefined>`, but a JSON `null` response still produces JavaScript `null`. A caller checking only for `undefined` can therefore compile successfully and crash at runtime.

Could we use your alternative A, with a separate `ResponseNullValue` setting defaulting to `Null`? That would preserve existing behavior while allowing users to explicitly opt into `Undefined`. When enabled, it should change both the return type and the actual response conversion consistently across supported templates.

We should add runtime tests alongside the snapshots to verify this. Empty bodies and HTTP 204 also need explicit consideration, since no response content is distinct from JSON `null` and isn't implied by a nullable schema.

I've merged latest `master` into this branch, including the development-tooling fixes; all 67 TypeScript tests now pass locally.

atmoos:
> The generated return type and runtime value currently disagree. For example, the new nullable-string snapshot declares `Promise<string | undefined>`, but a JSON `null` response still produces JavaScript `null`. A caller checking only for `undefined` can therefore compile successfully and crash at runtime.
> 
> Could we use your alternative A, with a separate `ResponseNullValue` setting defaulting to `Null`? That would preserve existing behavior while allowing users to explicitly opt into `Undefined`. When enabled, it should change both the return type and the actual response conversion consistently across supported templates.
> 
> We should add runtime tests alongside the snapshots to verify this. Empty bodies and HTTP 204 also need explicit consideration, since no response content is distinct from JSON `null` and isn't implied by a nullable schema.
> 
> I've merged latest `master` into this branch, including the development-tooling fixes; all 67 TypeScript tests now pass locally.

Thanks for your feedback! I'll gladly,
- realise option A), the `ResponseNullValue`
- align runtime with compile time types & values
- align the response conversion
- add runtime tests (& others if needed)

This might take some time to get done properly. Let's say on the order of a week.

